PR #3, opened 2015-03-15
C-Christenson Bowling Challenge

c-christenson:


README.md
```diff
@@ -1,54 +1,24 @@
...
-
-In the image below you can find some score examples.
-
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
-
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+- Fix certain style features:
+    - remove "pinsHit" from input field
+    - set restrictions on input field (currently any value can be inputted)
+    - resize input field
```
spike01: :+1 on writing your README

spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,167 @@
...
+      scorecard.hitPins(10);
+      scorecard.tallyFrameScore();
+      scorecard.tallyTotalScore();
+      expect(scorecard.score).toEqual(300);
+      expect(scorecard.isGameOver).toEqual(true);
+    });
+
+  });
+
+});
```
spike01: These tests are great. However, there's always room for improvement, so:
1. Try to avoid the word "should" - it either does, or it doesn't. Prefer a more concise, descriptive style with less repeated phrases
2. One expectation per test, which you managed right up to the last test!
3. You can use helper methods to dry your tests up - rolling a perfect game could be DRYed up with a for loop, for example

src/ScoreCard.js
```diff
@@ -0,0 +1,81 @@
...
+    bowl1: 0, 
+    bowl2: 0, 
+    score: 0,
+    strike: false,
+    spare: false,
+  }
+};
+
+ScoreCard.prototype.hitPins = function(numberOfPins){
+  if (this.currentBowl === 1) {
```
spike01: Try to extract conditional checks into their own, descriptive method (ok, this one is not the best example as it's quite clear, but it's a good habit to get into)

src/ScoreCard.js
```diff
@@ -0,0 +1,81 @@
...
+    this.frames[this.currentFrame].bowl1 = numberOfPins;
+    this.currentBowl += 1
+    if ((this.currentFrame === this.totalFrames + 1 & this.frames[this.totalFrames].spare === true) || (this.currentFrame === this.maxFrames)) {
```
spike01: This would be a good candidate for a method extraction

src/ScoreCard.js
```diff
@@ -0,0 +1,81 @@
...
+      this.frames[this.currentFrame].bowl2 = numberOfPins;
+      if (this.frames[this.currentFrame].bowl1 + numberOfPins === 10) {
+        this.frames[this.currentFrame].spare = true;
+      }
+      if ((this.currentFrame === this.totalFrames && this.frames[this.currentFrame].spare === false) || (this.currentFrame === this.totalFrames + 1)) {
+        this.isGameOver = true
+      }
+      this.changeFrame();
+    }
+};
```
spike01: Avoid:

- nested conditionals (these can be decomposed into helper methods)
- unclear control flow: every time you have these stacked if statements, it's not clear if one, none, or all branches fire. A lot of these feel like they could be their own methods

src/ScoreCard.js
```diff
@@ -0,0 +1,81 @@
...
+      this.frames[frameIndex].score += (this.frames[frameIndex + 1].bowl1);
+      if (this.frames[frameIndex + 1].strike === true) {
+        this.frames[frameIndex].score += (this.frames[frameIndex + 2].bowl1);
+      }
+      else {
+        this.frames[frameIndex].score += (this.frames[frameIndex + 1].bowl2);
+      }
+    }
+  }
+};
```
spike01: Again, this method works, but it very difficult to reason about

src/interface.js
```diff
@@ -0,0 +1,32 @@
...
+    $('#frame1').text(scorecard.frames[1].score);
+    $('#frame2').text(scorecard.frames[2].score);
+    $('#frame3').text(scorecard.frames[3].score);
+    $('#frame4').text(scorecard.frames[4].score);
+    $('#frame5').text(scorecard.frames[5].score);
+    $('#frame6').text(scorecard.frames[6].score);
+    $('#frame7').text(scorecard.frames[7].score);
+    $('#frame8').text(scorecard.frames[8].score);
+    $('#frame9').text(scorecard.frames[9].score);
+    $('#frame10').text(scorecard.frames[10].score);
```
spike01: Can you think of a way to DRY this up?

src/interface.js
```diff
@@ -0,0 +1,32 @@
...
+  $('#score').text(scorecard.score);
+  });
+
+
+
+
+
+
```
spike01: Clean up your whitespace!

spike01:
This is a great effort, especially since you also created a UI layer that was easy to use and that worked. Like I said in the line notes, try to decompose methods into smaller, easier to follow methods. However, as you were testing at a higher level of abstraction (behaviour over implementation), the complexity wasn't as much of a problem.
The only glaring omission was a reset button for when I finished playing a game! :checkered_flag:

PR #4, opened 2015-03-15
Kevin Lanzon partial solution

kevinlanzon:


js/bowling.js
```diff
@@ -0,0 +1,43 @@
...
+  this.frameNumber = 1;
+  this.rollNumber = 1;
+  this.strikes = 0;
+  this.spares = 0;
+  this.currentScore = 0;
+};
+
+Bowling.prototype.pinsHit = function(number) {
+  if(number === 10)
+    {this.strikes += 1;
```
spike01: Watch formatting - if statements should look like this:
```javascript
if (condition) {
  // code to execute
} else {
 // other code to execute
}
```

js/bowling.js
```diff
@@ -0,0 +1,43 @@
...
+    this.rollNumber += 1;
+    this.scoreTotal();
+    this.frameNumber += 1;}
+  else{this.pins = (this.pins -= number);
```
spike01: This can be simplified to `this.pins -= pins`

js/bowling.js
```diff
@@ -0,0 +1,43 @@
...
+Bowling.prototype.frameTotal = function() {
+  if(this.frameNumber === 10) {this.frameNumber == 1;}
+  else{this.frameNumber += 1;}
+};
+
+Bowling.prototype.scoreTotal = function() {
+  if(this.spares != 0) {(this.currentScore = (10 * this.spares) + this.currentScore);}
+  else if(this.strikes != 0) {(this.currentScore = (10 * this.strikes) + this.currentScore);}
+  else{(this.currentScore = (10 - this.pins) + this.currentScore);}
+};
```
spike01: This is a really good start, but read some JS style guides and see how to format your JS - I find this kinda difficult to read, even though it works.

Be careful with relying heavily on conditionals - they are difficult to reason about and should really have every leg of the conditional tested

js/bowling.js
```diff
@@ -0,0 +1,43 @@
...
+
+Bowling.prototype.nextFrame = function() {
+  if(this.rollNumber === 1) {this.rollNumber += 1;}
+  else if(this.rollNumber === 2)
+    {this.rollNumber -= 1;
+    this.scoreTotal();
+    this.frameTotal();
+    this.pins = 10;}
+};
```
spike01: What executes if neither of these conditions is true? Or did you mean to use else?

spec/bowlingSpec.js
```diff
@@ -0,0 +1,55 @@
...
+    expect(bowling.frameNumber).toEqual(2);
+  });
+
+  it('can have a total score after each frame', function() {
+    bowling.pinsHit(2);
+    bowling.pinsHit(3);
+    expect(bowling.currentScore).toEqual(5);
+  });
+
+});
```
spike01: This tests are a good start, until you start having multiple expectations. If you are going to test the basic state of your object when it is created, then you don't need to test that again with every following test.

Also, try to move towards a more concise style for your test descriptions: it "moves to the next frame", it "has a total score after each frame".

spike01:
This is a good start, shame you didn't get to the more complex logic as you are on the right track. Be careful about style when formatting your code - making your code consistent with industry standards is important

:construction:

PR #5, opened 2015-03-15
all logic finished, tests passing. need to refactor and work on front en...

robertpulson:
...d, dynamically updating inputs etc.


spec/frameSpec.js
```diff
@@ -0,0 +1,47 @@
+describe('Frame', function() {
+
+  var frame;
+
+  beforeEach(function() {
+    frame = new Frame();
+  });
+
+  describe('when playing', function() {
```
silvabox: This is a superfluous `describe`; since there is no other context, 'when playing' is implicit.  Furthermore, 'when playing' is a context rather than a feature you are describing, so should be in a `context` method instead.  It's a semantic quibble, but nevertheless important.

spec/frameSpec.js
```diff
@@ -0,0 +1,47 @@
...
+
+    it('can receive a bowl', function() {
```
silvabox: In general, make sure the wording of your tests is consistent with the expectation.  Here, you are testing 'can receive a bowl' when in actual fact, the expectation is that following the very first bowl, `frame.firstBowl` is equal to the bowled value.

spec/frameSpec.js
```diff
@@ -0,0 +1,47 @@
...
+    it('knows when its result is a strike', function() {
+      frame.receiveBowl(10);
+      expect(frame.strike).toEqual(true);
+    });
+
+    it('skips the second bowl if the first bowl is a strike', function() {
+      frame.receiveBowl(10);
+      expect(frame.firstBowl).toBe(10);
+      expect(frame.secondBowl).toBe(0);
+      expect(frame.strike).toEqual(true);
```
silvabox: You have already tested `frame.strike` and `frame.firstBowl` in other tests.  Be careful to limit your expectations to the specific aspect you are testing.

spec/scorecardSpec.js
```diff
@@ -0,0 +1,189 @@
+describe('Scorecard', function() {
+
+  var scorecard;
+
+  beforeEach(function() {
+    scorecard = new Scorecard(Frame);
+  });
+
+  describe('when initialised', function() {
```
silvabox: use a `context` here

spec/scorecardSpec.js
```diff
@@ -0,0 +1,189 @@
...
+
+    it('has ten frames', function() {
+      expect(scorecard.frameRecord).toEqual([new Frame, new Frame, new Frame, new Frame, new Frame, new Frame, new Frame, new Frame, new Frame, new Frame]);
```
silvabox: How would you have tested 100 frames?  In order to verify this test is correct, I have to check that there are 10 `new Frame` entries in the expectation.  This makes the test unreadable and easily mistaken.  You should perhaps have 2 tests - one that tests there are 10 (using length) and another that each item is indeed a `Frame` (if you want to go that far).

spec/scorecardSpec.js
```diff
@@ -0,0 +1,189 @@
...
+    });
+
+  });
+
+  describe('when playing', function() {
```
silvabox: use a `context`.  Rule of thumb, if it starts 'when...' or 'if...' then it's a context!

src/frame.js
```diff
@@ -0,0 +1,30 @@
+var Frame = function() {
+  this.firstBowl = null;
+  this.secondBowl = 0;
+  this.result;
+  this.spare = false;
```
silvabox: with Booleans, use `is` or `has` prefixes to improve readability - rather like `?` methods in ruby.  So `isSpare` and `isStrike`.

src/frame.js
```diff
@@ -0,0 +1,30 @@
...
+  this.strike = false;
+};
+ 
+Frame.prototype.receiveBowl = function(pinsKnockedOver) {
```
silvabox: This method is too long.  Consider ways to separate the logic for handling which bowl it is from managing the result.  You have to some extent with calculateResult, but both these methods are dependent on `if/else if/else` statements which are ugly.

src/frame.js
```diff
@@ -0,0 +1,30 @@
...
+    this.calculateResult();
+  } else if (this.firstBowl != null) {
+    this.secondBowl += pinsKnockedOver;
+    this.calculateResult();
+  } else {
+    this.firstBowl = pinsKnockedOver;
+  };
+};
+
+Frame.prototype.calculateResult = function() {
```
silvabox: `result` is not set when there is a strike or spare.  Is this correct?  This could lead to unexpected behaviour and a disappointed consumer

src/frame.js
```diff
@@ -0,0 +1,30 @@
+var Frame = function() {
+  this.firstBowl = null;
+  this.secondBowl = 0;
+  this.result;
```
silvabox: This is redundant - it simply equates to `undefined`

src/scorecard.js
```diff
@@ -0,0 +1,73 @@
+var Scorecard = function(content) {
+  this.frameRecord = [];
+  this.totals = [];
+  this.DEFAULT_FRAMES = 10;
+  this.currentTurn = 0;
+  this.maxTurns = 20;
+  for (var i = 0; i < this.DEFAULT_FRAMES; i++) {
```
silvabox: Kudos for trying to use dependency injection here to remove the responsibility from scorecard.  However, it ends up being a confusing distraction and the two objects are so tightly coupled anyway I don't think it adds any benefit.  I'm sceptical of this approach for dependency injection anyway as scorecard is still responsible for calling `new content`, which feels neither here nor there.
Beware also unnecessarily eager-instantiating your objects.  Lazy instantiation can often lead to greater flexibility.  I.e. don't create the instances of `Frame` (`content`) until you need them.

src/scorecard.js
```diff
@@ -0,0 +1,73 @@
...
+    this.frameRecord.push(new content);
+  };
+};
+
+Scorecard.prototype.updateTotal = function() {
+  if (this.frameRecord[this.getCurrentFrame()].strike === true || this.frameRecord[this.getCurrentFrame()].spare === true) {
```
silvabox: DRY!!!!

How many times is this repeated? `this.frameRecord[this.getCurrentFrame()]` throughout this object?

`getCurrentFrame` should return the current frame, as the method name implies, not the current frame number.  This single change in your code would improve the readability of this object at least tenfold.

src/scorecard.js
```diff
@@ -0,0 +1,73 @@
...
+    this.totals.push(10);
```
silvabox: This whole conditional is unnecessary.  Scorecard is responsible for keeping the running score from the frames - the Frame is responsible for the interpreting the result of each frame.  Simply setting `result` to 10 in your frame for a spare or score would remove the need for this if statement and correctly encapsulate this feature.

src/scorecard.js
```diff
@@ -0,0 +1,73 @@
...
+
+Scorecard.prototype.checkCurrentFrameForStrikeWith = function(pinsKnockedOver) {
+  if (this.checkEven(this.currentTurn) === true && pinsKnockedOver === 10) this.currentTurn +=1;
+}
+
+Scorecard.prototype.checkIfFrameIsOver = function() {
+  if (this.checkEven(this.currentTurn) != true) this.updateTotal();
+}
+
+Scorecard.prototype.bowl = function(pinsKnockedOver) {
```
silvabox: I'm not even going to try to read this method and neither will anyone else.  It is brittle, has multiple responsibilities, makes multiple changes to the state of the object; and worst of all, none of which is tested.
Sorry that's harsh - but I want to get across how important that would be in the real world :wink:

src/scorecard.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+    if (this.checkLastFrameForSpare() === true) this.totals[this.getCurrentFrame() - 1] += pinsKnockedOver;
+    this.frameRecord[this.getCurrentFrame()].receiveBowl(pinsKnockedOver);
+    this.checkIfFrameIsOver();
+    this.currentTurn += 1;
+  };
+};
+
+Scorecard.prototype.getCurrentFrame = function() {
+  if (this.currentTurn >= 20) return 9;
```
silvabox: Calculating the current frame by checking if the current turn is even then subtracting 1 and dividing by 2 is highly brittle.  Why not just track the current frame separately?  This kind of anti-pattern quickly leads to problems where one variable is shoehorned into 2 responsibilities.

silvabox:
Hi Phil,

Well done on your bowling challenge.  I've made lots of comments, but overall it was a good effort and I liked the fact that you'd built a UI.

In general, be careful that your test expectation match the test description.  Also, note that while your UI might limit the range of values available, what happens when your objects are consumed directly?  Your frame does not check for invalid values or totals for example.

:checkered_flag:

silvabox:
Hi Phil,
Please note, I advised you to use `context` methods instead of `describe` methods in my comments.  I was confusing Jasmine with RSpec - my apologies - I'm getting old :wink:  Jasmine doesn't have contexts.
Ben

PR #8, opened 2015-03-16
moved file over from original folder

braunsnow:


spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,146 @@
...
+        spares(10);
+        scorecard.addScore(5);
+        expect(scorecard.currentScore()).toEqual(150);
+      });
+
+      it("should know the game has ended after the player rolls his 21st ball if he spares the 10th frame", function() {
+        spyOn(scorecard, "_alertGameIsFinished");
+        spares(10);
+        scorecard.addScore(5);
+        expect(scorecard._alertGameIsFinished).toHaveBeenCalled();
```
danldb: some repetition - think about extracting into methods

src/ScoreCard.js
```diff
@@ -0,0 +1,125 @@
...
+      case 12:
+        return true;
+      case 11:
+        return frameBeforeBefore[0] < 10;
+      case 10:
+        return (currFrame[0] + currFrame[1]) < 10;
+      default:
+        return false;
+    };
+  };
```
danldb: suggested refactor - extract the switch statements using a data construct such as an object

src/ScoreCard.js
```diff
@@ -0,0 +1,125 @@
...
+
+  ScoreCard.prototype._scoreCheck = function (rollScore) {
+    if ((currFrame[0] + rollScore) > 10 || rollScore > 10) {alert("cheater!")};
+  };
+
+  ScoreCard.prototype._alertGameIsFinished = function () {
+    alert("Game finished, your score is "+ this.currentScore());
+  };
```
danldb: this alert pops up during every single jasmine test - I would remove this from your logic and let the front end handle a completed game

src/ScoreTables.js
```diff
@@ -0,0 +1,65 @@
...
+        return i;
+      };
+      if (this.scoreCards[i].frameNo < currentFrameNo) {
+        this.scoreCards[i].playTurn = true;
+      };
+      if (this.scoreCards[this.scoreCards.length].playTurn === false && this.scoreCards[this.scoreCards.length].frameNo === currentFrameNo) {
+        currentFrameNo += 1;
+      };
+    };
+  };
```
danldb: multiple if statements - code smell. Consider storing state in an object.

src/ScoreTables.js
```diff
@@ -0,0 +1,65 @@
...
+      };  
+      scoreTable+= "<td style='width: 70px;'>" + total +"</td>";
+    };
+
+    scoreTable+="</table>";
+
+    return scoreTable;
+
+  };
+
```
danldb: I'm impressed by your ingenuity! This works, but it is bad practice to mix your view with your model. Separation of concerns.

src/Setup.js
```diff
@@ -0,0 +1,32 @@
+$(document).ready(function(){
+});
+
```
danldb: You should only have one document ready in an application

src/ScoreCard.js
```diff
@@ -0,0 +1,125 @@
...
+    return this.frameNo < 10; 
+  };
+
+  ScoreCard.prototype._isThe11thFrame = function () {
+    return this.frameNo === 10; 
+  };
+
+  ScoreCard.prototype._isThe12thFrame = function () {
+    return this.frameNo === 11; 
+  };
```
danldb: There is a code smell here - referring to frame more than self, suggests that there should be a Frame function

src/ScoreCard.js
```diff
@@ -0,0 +1,125 @@
...
+    if (rollScore === 10 || this._wasThe2ndRollOfFrame() || (this.frameNo === 10) ) {
+      this.frameNo += 1;
+      this.playTurn = false;
+      this.score[this.frameNo] = [];
+    };
+
+    if (this._gameIsFinished) {this._alertGameIsFinished()};
+
+  };
+
```
danldb: This function is way too big, illegible, and has too many responsibilities. Refactor into concise functions that each have a single responsibility.

danldb:
:checkered_flag: 

Hi Yannick, thanks for the submission!

Generally good, and I have awarded a pass. 

You have a mostly working application, but there are some errors. The biggest issue is regarding separating your concerns - you have mixed the view with your model. We will cover this in more detail this week.

Where is your readme? It would be great to see how to install and run the program, learning objectives, technologies used etc.

PR #10, opened 2015-03-16
Bowling Challenge

costassarris:


js/BowlingScorecard.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+  if (this.allFrames.length == 9) {
+    if ((this.currentFrame.length == 2) && ((this.currentFrame[0] + this.currentFrame[1]) < 10)) { this.allFrames.push([this.currentFrame[0] + this.currentFrame[1]])};
+    if (this.currentFrame.length == 3) { this.allFrames.push([this.currentFrame[0] + this.currentFrame[1] + this.currentFrame[2]])};
+    if (this.currentFrame.length > 3) throw new "The game is over";
+  }
+  this.beforeBefore = this.before;
+  this.before = result;
+};
+
```
danldb: long function, multiple 'if' is a code smell, refactor into a data structure and use something like a hash lookup.

spec/BowlingScorecardSpec.js
```diff
@@ -0,0 +1,70 @@
...
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    scorecard.bowl(10);
+    expect(scorecard.allFrames.length).toEqual(10);
+    expect(scorecard.allFrames[scorecard.allFrames.length - 1]).toEqual([30]);
```
danldb: lots of repetition - consider creating a function

danldb:
:construction: 

Hi Costas - thanks for the submission!

I have given you a :construction: tag rather than the checkered flag, as you have the majority of your logic in a single function. I am actually impressed that you managed to get that to work, but it breaks single responsibility principle. 

In general, more than 1 or 2 if statements in a row is a BIG code smell - it is a sign you should create an object to hold that state.

Let us know if you have any questions.

PR #11, opened 2015-03-16
My bowling challenge

katebeavis:
I'm sorry but this is unfinished as I struggled a bit with it. Also all in one file as I was going to refactor it into separate files when I finished it...which I didn't do :(

**I've just realised that I didn't delete 'pin' tests (I deleted the pin file as it was no longer needed) so there will be three failing tests at the start**


spec/FrameSpec.js
```diff
@@ -0,0 +1,76 @@
...
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
+      bowl.nextFrame();
```
spike01: Maybe a for loop here?

spec/FrameSpec.js
```diff
@@ -0,0 +1,76 @@
...
+      expect(bowl.frameNumber).toEqual(2);
+    });
+
+    it("should let rolls hit pins", function() {
+      bowl.roll1(1);
+      expect(bowl.pins).toEqual(9);
+    });
+
+    it("should move to the next frame when a strike is scored",
+      function() {
```
spike01: Watch your formatting

spec/FrameSpec.js
```diff
@@ -0,0 +1,76 @@
...
+    it("should increase the roll score when a roll hits a pin",
+      function() {
+        bowl.roll1(10);
+        expect(bowl.roll1).toEqual(10);
+      });
+
+    it("should only let the roll have a maximum score of 10", function() {
+      bowl.roll1(15);
+      expect(bowl.roll1).toEqual(10)
+    });
```
spike01: This is... fairly surprising! This would be a case where an error message or something like that would be suitable, as it's physically impossible to roll 15 - to have that automatically corrected to 10 is alarming!

src/Frame.js
```diff
@@ -0,0 +1,37 @@
...
+  this.maximumRollScore = this.pins;
+  this.frameNumber = 1;
+  this.maximumFrames = 10;
+  this.strikes = 0;
+  this.spares = 0;
+};
+
+Bowl.prototype.nextFrame = function() {
+  this.frameNumber++;
+  if (this.frameNumber > this.maximumFrames) this.frameNumber = this.maximumFrames;
```
spike01: Generally, stick to splitting your if statements over multiple lines, with brackets - it makes for clearer reading

spike01:
Good start! Can't really say much more at this point!

:construction:

PR #13, opened 2015-03-16
Javascript code works

wardymate:
Javascript code works - started to build HTML and jquery, adding extra functions to the Scorer class to assist with the jquery. Breaks the Single responsibility prinicple though. Will get the HTML JQuery and CSS working and then look to move the helper methods out into another class possibly.


spec/ScorerSpec.js
```diff
@@ -0,0 +1,115 @@
...
+      scorer.roll(2);
+      scorer.roll(10);
+      scorer.roll(7);
+      scorer.roll(1);
+      expect(scorer.total()).toEqual(172);
+    });
+
+  });
+
+  describe('helper methods', function() {
```
spike01: "helper methods" is probably not the right name for these tests - but they are good tests nonetheless

spec/ScorerSpec.js
```diff
@@ -0,0 +1,115 @@
...
+      expect(scorer.frameNumber()).toEqual(3);
+    });
+    it('allows for a 3rd ball in the tenth frame', function() {
+      multiRoll(12,10);
+      expect(scorer.frameNumber()).toEqual(10);
+    });
+
+  });
+
+});
```
spike01: :+1: This is awesome

src/Scorer.js
```diff
@@ -0,0 +1,41 @@
+var Scorer = function() {
+  this.rolls = [];
+  this.frameCounter = 0;
+};
+
+Scorer.prototype.roll = function(score) {
+  this.rolls.push(score);
+  if (this.frameCounter <10) {
+    if (score === 10) {
+      this.frameCounter +=1;}
```
spike01: Close your else bracket one line further down:

```javascript
if (condition) {
// code to execute
} else {
// other code to execute
}
```

src/Scorer.js
```diff
@@ -0,0 +1,41 @@
...
+    else {
+      this.frameCounter +=0.5;};}
```
spike01: Interesting approach! Good that it works, but be careful with floats, they can sometimes fail in sneaky ways

src/Scorer.js
```diff
@@ -0,0 +1,41 @@
...
+
+  var runningTotal = 0;
+  var rollNumber = 0;
+  var frameNumber = 1
+
+  for (var i = frameNumber; i < 11; i++) {
+    if(this.rolls[rollNumber] === 10) {
+      runningTotal += this.rolls[rollNumber] + this.rolls[rollNumber+1]+ this.rolls[rollNumber+2];
+      rollNumber +=1;}
+    else if (this.rolls[rollNumber] + this.rolls[rollNumber+1] === 10) {
```
spike01: Consider extracting these conditions to descriptively named functions

index.html
```diff
@@ -0,0 +1,163 @@
...
+
+});
+
+
+
+</script>
+
+
+</body>
+</html>
```
spike01: Yup... your game logic and jQuery are way too coupled here

spike01:
Good submission - clean, with just some small formatting problems. Fell down a bit on the UI layer, but only because you didn't create a layer between your game and your jQuery. :checkered_flag:

PR #14, opened 2015-03-16
Bowling Challenge

louisebeh:
This is very much a work in progress.


src/Frame.js
```diff
@@ -0,0 +1,36 @@
...
+  this.score = 0;
+  this.Pin = 10;
+};
+
+  Frame.prototype.score = function() {
+    this.score = this.roll1 + this.roll2;
+  };
+
+  Frame.prototype.isSpare = function() {
+  if (roll1 + roll2) === 10 {
```
spike01: Wrap the whole condition in brackets

src/Frame.js
```diff
@@ -0,0 +1,36 @@
...
+    return true
+  } else {
+    return false
+  }
+
+  Frame.prototype.isStrike = function() {
+    if roll1 === 10 {
```
spike01: Ditto here

src/Frame.js
```diff
@@ -0,0 +1,36 @@
...
+      return true
+    } else {
+      return false
+    }
+  };
```
spike01: Actually, if you find yourself explicitly returning true and false, your code probably needs to just be:
```javascript
Frame.prototype.isStrike = function() {
  return roll1 === 10;
}
```

src/Pin.js
```diff
@@ -0,0 +1,16 @@
...
+  };
+
+  Pin.prototype.state = function() {
+    if (this.standing === false) {
+      return 'knocked down';
+    } else {
+      return 'standing';
+    }
+  };
+
```
spike01: Choose one way of representing state, not two things that are basically the same thing

spike01:
Indeed it is
:construction:

PR #29, opened 2015-04-26
Pull Request

Nadav-Rosenberg:


spec/GameSpec.js
```diff
@@ -0,0 +1,111 @@
...
+
+  it('can roll a gutter game', function() {
+
+    rollMany(0,20);
+    expect(game.score()).toEqual(0);
+
+  });
+
+  it('can roll a game on ones', function() {
+
```
spike01: Be consistent with line spacing - you don't have breaks below, why do you have them here? (or vice versa)

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+      game.firstRoll = false;
+      game.firstRollScore = pins;
+    } else {
+      game.firstRoll = true;
+      game.firstRollScore = 0;
+    }    
+  };
+
+  function pinsError() {
+    var a = "Max second roll score is between 0 and "
```
spike01: `var a` and `var b` are terrible names - what do they mean?

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+    var b = (10 - game.firstRollScore).toString();
+    if (pins < 0) {
+      throw new Error("Min roll score is 0")
+    } else if (game.firstRoll == true && pins > 10) {
+      throw new Error("Max first roll score is 10")
+    } else if (game.firstRoll == false && pins > (10 - game.firstRollScore)) {
+      throw new Error(a += b)
```
spike01: `throw new Error("Max second roll score is between 0 and " + remainingPins)`

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+      game.firstRollScore = pins;
+    } else {
+      game.firstRoll = true;
+      game.firstRollScore = 0;
+    }    
+  };
+
+  function pinsError() {
+    var a = "Max second roll score is between 0 and "
+    var b = (10 - game.firstRollScore).toString();
```
spike01: `var remainingPins`

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+
+  this.rolls = [];
+  this.firstRoll = true;
+  this.firstRollScore = 0;
+
+};
+
+Game.prototype.roll = function(pins) {
+  var game = this;
+  pinsError();
```
spike01: This makes me think that there's an error every time

src/Game.js
```diff
@@ -0,0 +1,81 @@
+var Game = function() {
+
+  this.rolls = [];
+  this.firstRoll = true;
```
spike01: Is this the first roll of the game or the frame?

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+Game.prototype.roll = function(pins) {
+  var game = this;
+  pinsError();
+  this.rolls.push(pins);
+  rollOrder();
+
+  function rollOrder() {
+    if (game.firstRoll == true && pins < 10) {
```
spike01: `if(notStrike())`?

spike01:
```javascript
function notStrike() {
  return game.firstRoll == true && pins < 10
}
```

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+
+};
+
+
+
+
+Game.prototype.score = function() {
+  
+  var total = 0;
+  var i = 0;
```
spike01: i is usually for internal iterators in for loops - what is this?!

spike01: So call it `rollIndex`

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      normalScore();
+      i += 2;
+    }
+    
+  }
+
+  return total;
+
+  function Strike() {
```
spike01: Convention - camelCase means that this would strike, not Strike

spike01: Only use the capital if you are going to use the `new` keyword

spike01: For things that return booleans, in JavaScript, prefer isStrike() and isSpare()

src/Game.js
```diff
@@ -0,0 +1,81 @@
...
+
+
+
+Game.prototype.score = function() {
+  
+  var total = 0;
+  var i = 0;
+  var game = this;
+
+  for (f = 0; f < 10; f++) {
```
spike01: You're forgetting a var keyword - now you have a global `f`!
Use i for for loop interation (at least for the first loop) - or if you want to be really descriptive, you could use use `var frame`

spike01:
:checkered_flag: Great submission, work on clarity and naming

PR #37, opened 2015-04-27
Still more to do

massud:
I had a few attempts at creating a logic for my game, but came undone with bonus frames. I think my latest attempt is finally on the right path but unfortunately not enough time to get it done. Will try and finish this over the week and hopefully have a decent front end.


spec/frameSpec.js
```diff
@@ -0,0 +1,21 @@
+describe("Frame", function(){
+
+  var roll;
+  var score;
```
spike01: Keep master clean if you aren't using things

spec/frameSpec.js
```diff
@@ -0,0 +1,21 @@
...
+  
+  beforeEach(function(){
+    frame = new Frame();
+  });
+
+  describe('Frame', function(){
+    it('defaults to zero frames', function(){
+      frame = new Frame;
```
spike01: You've extracted this to a beforeEach hook, so take it out here

spec/rollSpec.js
```diff
@@ -0,0 +1,19 @@
...
+
+    it('should start with no pins knocked', function(){
+      expect(roll.pinsknockedtotal).toEqual(0);
+    });
+
+    it('can knock pins down after roll', function(){
+      roll.ball(4)
+      expect(roll.pinsknockedtotal).toEqual(4);
+    });
+  });
```
spike01: This is a very thin wrapper object around the concept of a number - maybe a bit too much

spec/scoreSpec.js
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  describe('roll counting', function(){
+
+    it('it can roll', function(){
+      expect(score.rolls.length).not.toEqual(0);
+    });
+
+    it('it can hit pins', function(){
+      score.rolls[0].ball(4)
```
spike01: Law of Demeter - talk to your friends, but don't talk to your friend's friends. In other words, don't chain dots!

Score -> Roll -> Ball method

spike01:
## Frame
- hold rolls (up to you to decide if you a roll object, or if it's just numbers) in an array

## Score
- provide a total score

## Here are a few questions
- should the frame know if it's a strike or a spare?
- how does the score calculate? By taking the values reported by the frame, or by calculating the strike and spare bonuses?
- do you want a game that is playable? Because if not, you could just supply an array of score
- how do we know if something is a strike/spare?
- where do I keep the bonus scores?
- how do we know how many frames are in the game?
- how do we know how many rolls are in the frame?

## Approach
- let's make a scorecard that can total ONE frame
- let's make scorecard that can total NINE frames
- let's make a score that can total TWO frames correctly with a spare
- THREE frames with a strike
- FOUR with 2 strikes
- then worry about the tenth frame

## Common problems
- how do we move through the frames?
- how do we do the tenth frame?

PR #40, opened 2015-04-29
bowling challenge, not as good as I wish it was

saramoohead:


spike01:
:construction: Almost there, let's work on refactoring

PR #42, opened 2015-05-07
Mark’s Bowling Submission

Gwasanaethau:
The challenge so far.


spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
+describe('Bowling Scorecard', function() {
+  var scorecard, frame;
+
+  beforeEach(function() {
+    frame = new Frame();
+  });
+
+  describe('bowls a single ball and –', function() {
+    describe('keeps track of the frame’s score, which:', function() {
```
sjmog: Nice use of `context` (or as close as Jasmine gets)

spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
...
+      it('is initally 0', function() {
+        expect(frame.frameScore).toEqual(0);
+      });
+
+      it('(potentially) increases as balls are bowled', function() {
+        frame.bowl();
```
sjmog: So frame is both bowling *and* keeping track of its `frameScore` - this looks like something that might be better for two separate prototypes

spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
...
+        expect(frame.frameScore >= 0).toBeTruthy();
+      });
+
+      it('does not exceed the initial number of skittles', function() {
+        initialNumberOfSkittles = frame.skittlesRemaining;
```
sjmog: Use `var` for declaring variables

spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
...
+        frame.bowl();
```
sjmog: Might have been tempted to stub the response to `bowl()` to be a positive integer - you run the risk of getting a 'false positive' result for this test if, say, `bowl()` returns `null`

spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
+describe('Bowling Scorecard', function() {
+  var scorecard, frame;
```
sjmog: `scorecard` isn't used - this file is really `FrameSpec.js` (a unit test for `Frame`)

spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+  });
+
+  describe('plays a single standard frame –', function() {
+    it('knocks down all skittles on first bowl', function() {
+      spyOn(frame, 'bowl').and.returnValue(10);
+      expect(frame.play()).toEqual('Strike');
+      expect(frame.frameScore).toEqual(10);
+      expect(frame.skittlesRemaining).toEqual(0);
+    });
```
sjmog: This makes good sense to me :+1:

spec/BowlingSpec.js
```diff
@@ -0,0 +1,64 @@
...
+
+    it('knocks down all skittles with two bowls', function() {
+      spyOn(frame, 'bowl').and.returnValue(5);
+      expect(frame.play()).toEqual('Spare');
```
sjmog:
```javascript
frame.play()
expect(frame.result).toEqual('Spare');
```

src/Frame.js
```diff
@@ -0,0 +1,41 @@
+var Frame = function() {
+  this.frameScore = 0;
+  this.skittlesRemaining = 10;
+};
+
+Frame.prototype.play = function() {
+  score = this.bowl();
```
sjmog: Use `var` to avoid leaking out

src/Frame.js
```diff
@@ -0,0 +1,41 @@
...
+  this.frameScore += score;
+  this.skittlesRemaining -= score;
+  if(this.skittlesRemaining === 0) {
+    return 'Strike';
```
sjmog: Would it not make more sense to set some state as `'Strike'` so it can be accessed later?

src/Frame.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+  score = this.bowl();
```
sjmog: Don't need `var` here as `score` is already scoped to `play()`

src/Frame.js
```diff
@@ -0,0 +1,41 @@
...
+  this.frameScore += score;
+  this.skittlesRemaining -= score;
+  if(this.skittlesRemaining === 0) {
+    return 'Spare';
+  } else {
+    return '' + this.frameScore;
```
sjmog: If you are going to set state with this somehow, can I suggest that you set it to one of three strings: `'Strike'`, `'Spare'`, or `'Open'` - then track the previous score in a different section of state

src/Frame.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+};
+
+Frame.prototype.bowl = function() {
+  score = Math.floor(Math.random() * (this.skittlesRemaining + 1));
```
sjmog: Use `var` here

src/Frame.js
```diff
@@ -0,0 +1,41 @@
...
+  this.frameScore += score;
+  this.skittlesRemaining -= score;
+  if(this.skittlesRemaining === 0) {
+    return 'Spare';
+  } else {
+    return '' + this.frameScore;
+  }
+};
+
+Frame.prototype.bowl = function() {
```
sjmog: Suggest rename this to `knock_pins`

src/Frame.js
```diff
@@ -0,0 +1,41 @@
+var Frame = function() {
+  this.frameScore = 0;
+  this.skittlesRemaining = 10;
```
sjmog: Maybe also have

```javascript
this.result = "";
this.previousResult = 0;
```

And *later on*, worry about splitting these concerns out into separate classes.

src/Scorecard.js
```diff
@@ -0,0 +1,3 @@
+var Scorecard = function() {
```
sjmog: Right direction - focus on keeping everything together for now, and we can worry about extra classes/prototypes later

SpecRunner.html
```diff
@@ -0,0 +1,25 @@
...
+    <link rel="shortcut icon" type="image/png" href="lib/jasmine-2.3.0/jasmine_favicon.png">
+    <link rel="stylesheet" href="lib/jasmine-custom.css">
+
+    <script src="lib/jasmine-2.3.0/jasmine.js"></script>
+    <script src="lib/jasmine-2.3.0/jasmine-html.js"></script>
+    <script src="lib/jasmine-2.3.0/boot.js"></script>
+
+    <!-- include source files here... -->
+    <script src="src/Frame.js"></script>
+    <script src="src/Scorecard.js"></script>
```
sjmog: Lose this for now (until `Scorecard` is defined)

SpecRunner.html
```diff
@@ -0,0 +1,25 @@
...
+
+    <!-- include spec files here... -->
+    <script src="spec/BowlingSpec.js"></script>
```
sjmog: Rename this to `FrameSpec.js` - it's kinda a unit test anyway (because you haven't done any 'feature testing' through an interface)

PR #52, opened 2015-06-07
Bowling challenge

AnnaKL:


AnnaKL:
I tried but unfortunately didn't manage to implement 10th frame functionality.

spec/feature/bowlingSpec.js
```diff
@@ -0,0 +1,64 @@
...
+
+it('can record total score of a game with strikes but without a bonus for last frame', function(){
+  for(var i = 0; i < 9; i++) {
+    var frame = new Frame;
+    scorecard.addFrame(frame);
+    frame.bowl(10);
+  }
+  var frame = new Frame;
+  scorecard.addFrame(frame);
+  frame.bowl(2);
```
sdorunga1: This should probably be `frame.roll(2)` just because it's more consistent with the frame object itself and the domain of bowling

spec/frameSpec.js
```diff
@@ -0,0 +1,78 @@
...
+    frame = new Frame();
+});
+
+it('can hold 10 pins', function(){
+   expect(frame.startingPins).toEqual(10);
+});
+
+it('can record a roll', function(){
+  frame.bowl(3);
+  expect(frame.bothRolls[0]).toEqual(3);
```
sdorunga1: `bothRolls` could probably just be `rolls` as it's clear enough

src/scorecard.js
```diff
@@ -0,0 +1,47 @@
+function Scorecard() {
+this.allFrames = [];
+};
+
+Scorecard.prototype.addFrame = function(frame) {
+  this.allFrames.push(frame);
+};
+
+Scorecard.prototype.isItSpare = function(index) {
```
sdorunga1: This and the `isStrike` method can just stay under Frame's responsibility

src/frame.js
```diff
@@ -0,0 +1,31 @@
...
+  this.rollNumber += 1
+  this.standingPins -= pins
+  this.bothRolls.push(pins)
+};
+
+Frame.prototype.isOver = function() {
+  return (this.rollNumber === 2 || this.standingPins === 0);
+};
+
+Frame.prototype.frameStrike = function() {
```
sdorunga1: should probably be `isStrike` for consistency with your other methods

PR #53, opened 2015-06-08
Bowling challenge for Dan J

DanielJohnston:


DanielJohnston:
OK. I've put a few notes in the README.md, but basically it's a complete, working simulator, with the unfortunate caveat that there's no frontend. The tests are all passing at my end, and you can interact with it using web console. I had to rewrite the whole thing a couple of times to handle the storage and access requirements without getting hideously complex, but it could be a lot more elegant and well structured. I'm a bit vague on how OOD could apply to this as it's really only one object, but the TDD kind of worked ok, give or take rewriting almost all of the code a couple of times.

src/Player.js
```diff
@@ -0,0 +1,22 @@
+function Player() {
```
sdorunga1: This should not be in here :). IT's especially confusing since it's a bowling app that has players.

spec/BowlingSpec.js
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  it("should score 5 when 3 then 2 pins knocked over", function() {
+    scorecard.bowl(3);
+    scorecard.bowl(2);
+    expect(scorecard.score()).toEqual(5);
+  });
+
+  it("should not accept 11 pins being knocked over in 1 bowl", function() {
+    expect(function() { scorecard.bowl(11); }).toThrow();
```
sdorunga1: It is a good practice to check for the actual error being thrown.

spec/BowlingSpec.js
```diff
@@ -0,0 +1,73 @@
...
+    scorecard.bowl(5);
+    expect(scorecard.whichFrame()).toEqual(2);
+  });
+
+  it("should be on frame 8 after 12 bowls, inc. 2 strikes", function() {
+    scorecard.multibowl(5,4,3,4,10,5,4,7,3,10,0,0);
+    expect(scorecard.whichFrame()).toEqual(8);
+  });
+
+  it("should have score 74 from 5,4,3,4,10,5,4,7,3,10,0,0", function() {
```
sdorunga1: Test is correct but could be more clear on the `why` of why it exists

spec/BowlingSpec.js
```diff
@@ -0,0 +1,73 @@
...
+    scorecard.multibowl(5,4,3,4,10,5,4,7,3,10,0,0);
+    expect(scorecard.score()).toEqual(74);
+  });
+
+  it("should have a gutter game score 0 from 20 lots of 0", function() {
+    scorecard.multibowl(0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0);
+    expect(scorecard.finalScore()).toEqual(0);
```
sdorunga1: Just reading the specs I wouldn't be able to tell the difference between `finalScore` and `score`

spec/BowlingSpec.js
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  it("should have a perfect game score 300 from 12 lots of 10", function() {
+    scorecard.multibowl(10,10,10,10,10,10,10,10,10,10,10,10);
+    expect(scorecard.finalScore()).toEqual(300);
+  });
+
+  // This is a dubious test as it checks an internal variable
```
sdorunga1: Comments =~ a bad thing.

src/bowling.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+// Check legality of the number of pins hit, store the bowl
+// Missing: 3 bowls on 10th
+Scorecard.prototype.bowl = function (pins) {
+
+  var shot = [this.currentFrame, this.currentRoll, pins];
+
+  // capture and fail on finished games
+  if (this.currentFrame >= 11) {
```
sdorunga1: It's not a good idea to doubt your code and check things twice.

src/bowling.js
```diff
@@ -0,0 +1,130 @@
...
+  // Move to the next roll or frame
+  this.nextRoll();
+
+};
+
+Scorecard.prototype.nextRoll = function () {
+  // Test game over
+  if (this.currentFrame >= 11) {
+    throw new Error("You can't roll any more. The game is over!");
+  // Test special case second roll in tenth frame
```
sdorunga1: Please do not duplicate code in methods that call each other

src/bowling.js
```diff
@@ -0,0 +1,130 @@
...
+// Same as score, but only returns a figure if the game's over
+Scorecard.prototype.finalScore = function () {
+  if(this.isGameOver) {
+    return this.score();
+  } else {
+    return "Game not over yet!";
+  };
+};
+
+Scorecard.prototype.isGameOver = function () {
```
sdorunga1: Good method, use it more :)

src/bowling.js
```diff
@@ -0,0 +1,130 @@
...
+  return this.currentFrame >= 11;
+};
+
+Scorecard.prototype.multibowl = function () {
```
sdorunga1: Good idea

src/bowling.js
```diff
@@ -0,0 +1,130 @@
...
+  var args = Array.prototype.slice.call(arguments);
```
sdorunga1: You could have avoided being clever by passing in an Array

PR #55, opened 2015-06-08
Final sub

timrobertson0122:


src/Frame.js
```diff
@@ -0,0 +1,71 @@
+function Frame() {
+  this.goCount = 0;
+  this.startingPins = 10;
+  this.standingPins = 10;
+  this.rolls = [];
+}
+
+Frame.prototype.registerGo = function(rollScore) {
```
sdorunga1: It would be nice to have functions named a little closer to the game domain itself.

spec/FrameSpec.js
```diff
@@ -0,0 +1,158 @@
+describe("Frame", function() {
+  var frame;
+
+  beforeEach(function() {
+    frame = new Frame();
+  });
+
+  describe('playing', function(){
```
sdorunga1: Descriptions inside the `describe` and `it` blocks could be a bit clearer/ more meaningfull

spec/FrameSpec.js
```diff
@@ -0,0 +1,158 @@
...
+
+    it("registers a gutter ball", function() {
+      frame.registerGo(0);
+      expect(frame.remainingPins()).toEqual(10); //count pins knocked down?//
```
sdorunga1: Don't leave comments just lying around. They can get very confusing and make everything quite cluttered

PR #63, opened 2015-07-16
Nicola Hughes Week 5 Challenge Submit

DataMinerUK:
Never user JavaScript for the backend. Had to redo it (hence bowling.js and bowling2.js). Got tied in knots. Why the f**k does `null + null` equal `0` and not `null`???? Gave up on testing. JQuery side was pretty measly. My api function is 120 lines long!!!!! Probably should have made another file for it. Why do you have to be so bloody explicit in your equalities???

This is why no one actually uses JavaScript, they use frameworks.

Can be viewed [here](http://datamineruk.github.io/bowling-challenge/)

My [api function](https://github.com/DataMinerUK/bowling-challenge/blob/master/src/bowling2.js#L87-L207) seems pretty ridiculous


ptolemybarnes:
CI is not running, but only because you need to pull in changes to the README from upstream :checkered_flag:

PR #65, opened 2015-07-19
Bowling-challenge

katsuraku:
See README.txt.


spec/bowlingSpec.js
```diff
@@ -0,0 +1,97 @@
...
+    it('ends the frame', function() {
+      scoreCard.roll(10);
+      scoreCard.roll(2);
+      expect(scoreCard.rollArray[1]).toBe(0);
+    });
+  });
+
+  describe('strike bonus', function() {
+    it('adds next two rolls as bonus points', function() {
+      scoreCard.roll(10);
```
ptolemybarnes: Have a helper method that will do all this rolling for you. Then you can do:
```ruby
manyRolls(10, 0, 3, 4,)
```
For example.

src/interface.js
```diff
@@ -0,0 +1,66 @@
...
+$( document ).ready(function() {
+
+  var showRollScores = function() {
+    $("#rolls").html(scoreCard.rollArray);
+    $("#frameScores").html(scoreCard.frameScores);
+    $("#totalScore").html(scoreCard.scoreTotal);
+
+  };
+
+  $("#0").click(function() {
```
ptolemybarnes: Very repetitive!

If you give your buttons a class, listen for clicks on that class, then use this.id to know which button was clicked on, you can reduce these down to a single listener.
=)

PR #73, opened 2015-07-20
Bowling business logic

duskyshelf:


spec/Bowlingspec.js
```diff
@@ -0,0 +1,144 @@
+describe('Bowling', function() {
+
+var scoresheet;
+
+  describe('Scoresheet', function() {
+
+    it('can store scores', function() {
+      scoresheet = new Scoresheet();
```
sdorunga1: Keep it DRY. Use a `beforeEach(function() {scoresheet = new Scoresheet();})`

spec/Bowlingspec.js
```diff
@@ -0,0 +1,144 @@
...
+      scoresheet.strikeCalc();
+      scoresheet.sum();
+      expect(scoresheet.total).toBe(300)
+    });
+
+    it('can score a running tally', function() {
+      scoresheet = new Scoresheet();
+      var frame_1 = [10,0];
+      var frame_2 = [2,3];
+      scoresheet.addFrame(frame_1);
```
sdorunga1: Maybe split out the two calculations into a helper method to explain what it does

src/Bowling.js
```diff
@@ -0,0 +1,55 @@
+var scoreFrame = function(frame) {
+  if (frame.length === 3) { return (frame[0] + frame[1] + frame[2]) };
```
sdorunga1: You can use reduce and get rid of the ifs

src/Bowling.js
```diff
@@ -0,0 +1,55 @@
...
+  this.round++;
+};
+
+Scoresheet.prototype.scoreFrames = function() {
+  this.tally = this.board.map(function(frame) {
+      return scoreFrame(frame)
+    });
+};
+
+Scoresheet.prototype.strikeCalc = function() {
```
sdorunga1: You could name this a bit clearer, something like `applyBonuses`

src/Bowling.js
```diff
@@ -0,0 +1,55 @@
...
+
+  if (this.board[8][0] === 10)
+    { this.tally[8] = this.tally[8] + this.board[9][0] + this.board[9][1] };
+  if (this.tally[8] === 10 && this.board[8][0] != 10)
+    { this.tally[8] = this.tally[8] + this.board[9][0] };
+};
+
+
+Scoresheet.prototype.sum = function() {
+  localSum = 0;
```
sdorunga1: You can use a reduce here

PR #75, opened 2015-07-20
Adds review guidelines

ptolemybarnes:


silvabox:
Good start - I would change 'Code free of comments' to 'Code free of unnecessary comments or commented-out sections'

There is a common issues page in the wiki which I've started adding to, but maybe I'm making it too detaied?

ptolemybarnes:
Updated :+1:

PR #80, opened 2015-08-22
Bowling challenge

adrianw1832:


src/bowlingGame.js
```diff
@@ -0,0 +1,104 @@
+function BowlingGame() {
```
zdajani: Think about separating out the responsibilities of the bowling game, ex. frame, scorecard

src/application.js
```diff
@@ -0,0 +1,67 @@
...
+  var buildButtons = function() {
+    for (rowID =0; rowID <= 10; rowID++) {
+      var input = $('<input type="submit">').val(rowID)
+        .on('click', $.proxy(registerValue, this));
+      var li = $('<li>').append(input);
+      $('.buttons').append(li);
+    }
+  };
+
+  var registerValue = function(cell) {
```
zdajani: better not to use general naming - maybe use registerRoll

ptolemybarnes:
Looks great @adrianw1832 :checkered_flag:

PR #87, opened 2015-08-31
for submission, not completed

winnieau:


lib/jasmine-2.3.4/boot.js
```diff
@@ -0,0 +1,121 @@
+/**
```
danldb: please don't commit these files, they are so disruptive to have to read through - instead use .gitignore or a CDN.

src/Scorecard.js
```diff
@@ -0,0 +1,23 @@
...
+  return this.totalScore = 0
+};
+
+ScoreCard.prototype.frameScore = function() {
+  return this.frameScore
+};
+
+ScoreCard.prototype.addFramePoints = function(number) {
+  return this.frameScore = this.frameScore + number
+};
```
danldb: This is really nice code - more of this please! :-D

danldb:
:construction: 

Looks good winnie - I would recommend you to finish this off at the end of the course, but in the meantime have a work through the kata that I sent you.

PR #88, opened 2015-08-31
Submitting bowling challenge

JoshuaTatterton:
The html page looks good, the javascript/jquery... not so much


lib/jasmine-2.3.4/boot.js
```diff
@@ -0,0 +1,121 @@
+/**
```
danldb: jasmine & jquery files should be in your .gitignore

server.rb
```diff
@@ -0,0 +1,3 @@
+require 'sinatra'
+
+set :public_folder, proc { File.join(root) }
```
danldb: nice static hosting

spec/bowlingScorerSpec.js
```diff
@@ -0,0 +1,136 @@
...
+      });
+      it("adds the score of the next frame", function() {
+        bowlingscorer.bowl(10);
+        bowlingscorer.bowl(8);
+        bowlingscorer.bowl(1);
+        expect(bowlingscorer.total).toEqual(28)
+      });
+    });
+    it("can score a perfect game", function() {
+      bowlingscorer.bowl(10);
```
danldb: use a function that makes the bowls, instead of typing each one out.

src/BowlingScorer.js
```diff
@@ -0,0 +1,179 @@
...
+        this.frameNo += 1;
+      }
+    }
+    return this.semifinalShot = pins
+  } else if(pins>=0&&pins<=10&&(pins%1===0)&&(this.frameNo<=12)&&(this.doubleStrike)) {
+    this.addPins(pins,this.frameNo-2);
+    this.frameNo += 1;
+    return pins;
+  }
+};
```
danldb: Too many conditionals - refactor so that the code is reusable. Consider using an object for the rules, or an index for the strike / spare status.

src/BowlingScorer.js
```diff
@@ -0,0 +1,179 @@
...
+
+// First attempt didn't work, ^^ is my second, keeping this for reference
+// I'll probably check what I did wrong at some point.
+
```
danldb: This file should be in your .gitignore. It is fine to keep it for yourself, but you should not share commented out code.

src/jqbowling.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    $("#total").text(scoreBoard.total);
+  
+    var i;
+    for(i=0; i < (scoreBoard.runningTotals.length); i++) {
+      var previousTotal = $("#frame"+i+"total").text();
+      if(previousTotal === "") { previousTotal = 0 }
+      var total = parseInt(previousTotal) + scoreBoard.runningTotals[i];
+      $("#frame"+(i+1)+"total").text(total);
+    }
```
danldb: Business logic looks very much coupled with your front end logic. The interface with DOM should not contain much in the way of logic, it shoudl purely be an interface between your business logic and the elements of the page.

styles/style.css
```diff
@@ -0,0 +1,102 @@
...
+.throwf10 {
+  vertical-align: middle;
+  float: right;
+  text-align: center;
+  margin: auto;
+  border-left: solid thin Black;
+  border-bottom: solid thin Black;
+  width: 33%;
+  height: 100%;
+}
```
danldb: class names are not descriptive enough.

src/jqbowling.js
```diff
@@ -0,0 +1,60 @@
...
+        $("#frame"+(scoreBoard.frameNo-1)+"shot1").text(scored)
+      } else if(scoreBoard.frameNo===10&&$("#frame9shot2").text!="") {
+        $("#frame10shot1").text(scored)
+      } else {
+        $("#frame"+(scoreBoard.frameNo)+"shot1").text(scored);
+      }
+    } else {
+      if(scoreBoard.spare) { scored = "/" }
+      $("#frame"+(scoreBoard.frameNo-1)+"shot2").text(scored);
+    }
```
danldb: This function is kind of insane. What happens if the rules change, or if you want to add a feature? You would have to navigate all this logic all over again. This is pure technical debt.

danldb:
:checkered_flag: 

Good attempt Josh, and worthy of a pass. However you really need to look at the over-use of if / else statements, and pay attention to single responsibility - a module that is responsible for the interface should only be responsible for the interface, and should have little-to-no logic.

PR #91, opened 2015-08-31
forgot to push from earlier could still use some refactoring

bgraves14:
a bit messy still some refactoring to do.


spec/bowlingFrameSpec.js
```diff
@@ -0,0 +1,79 @@
...
+    expect(function() {bowlingFrame.rollOne(-1); }).toThrow(new Error('Roll must be between 0 and 10'));
+  });
+
+  it('allows you to enter a second roll score', function() {
+    bowlingFrame.rollOne(7);
+    bowlingFrame.rollTwo(2);
+    expect(game.currentGame[1]).toEqual([7,2]);
+  });
+
+  it('raises an error if second score plus first score is above 10', function() {
```
ptolemybarnes: These 5 tests begin `raises an error`, which suggests you might group them within a describe block.
But also try to avoid specifying tests in too "codey" a way, with language such as 'raises an error'. You might phrase it as "doesn't allow" or something like that.

spec/bowlingFrameSpec.js
```diff
@@ -0,0 +1,79 @@
...
+
+  it('raises error if frames are not submitted in order', function() {
+    expect(function() { bowlingFrameThree = new BowlingFrame(3); }).toThrow(new Error('You must add to frames in correct order'));
+  });
+
+  it('raises error if first roll submitted is not frame one, roll one', function() {
+    expect(function() { bowlingFrame.rollTwo(7); }).toThrow(new Error('You must add to frame one, roll one first'));
+  });
+
+  it('allows a bonus roll if first two rolls are larger or equal to 10', function() {
```
ptolemybarnes: There is an opportunity to refactor this out to a helper method. Have a look at the method at the bottom of here: https://github.com/ptolemybarnes/bowling-scorecard/blob/master/spec/Scorecard.spec.js and how it's used.

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,162 @@
...
+    bowlingFrameTwo.rollTwo(1);
+    expect(game.currentScore()).toEqual(15);
+  });
+
+  it('current score should start at 0', function() {
+    expect(game.currentScore()).toEqual(0);
+  });
+
+  describe('Spare Logic', function() {
+    it('current score should not show if roll one of next frame has not been rolled', function() {
```
ptolemybarnes: Shouldn't the score be 10 at this stage though?

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,162 @@
+describe('Bowling Game', function() {
+
+  beforeEach(function() {
+    game = new BowlingGame();
+    bowlingFrame = new BowlingFrame(1);
+  });
+
+  it('can create a new game', function() {
```
ptolemybarnes: This is the sort of test you would probably delete after it's passing ;-)

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,162 @@
...
+      bowlingFrame.rollOne(10);
+      expect(game.currentScore()).toEqual(0);
+    });
+
+    it('should add whole next frame score to current score if next frame is not a strike', function() {
+      bowlingFrame.rollOne(10);
+      bowlingFrameTwo = new BowlingFrame(2);
+      bowlingFrameTwo.rollOne(3);
+      bowlingFrameTwo.rollTwo(5);
+      expect(game.scoreGrid[1]).toEqual(18);
```
ptolemybarnes: Here you're breaking encapsulation by reaching into the game object to retrieve a value stored in an array. It would be better to have a public interface to that array, a method like `game.getRollScore(10)` or something.

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,162 @@
...
+      bowlingFrameTwo = new BowlingFrame(2);
+      bowlingFrameTwo.rollOne(10);
+      bowlingFrameThree = new BowlingFrame(3);
+      bowlingFrameThree.rollOne(10);
+      expect(game.scoreGrid[1]).toEqual(30);
+    });
+  });
+
+  describe('Bonus Roll Logic', function() {
+    it('should add bonus roll to score if first two rolls are greater or equal to ten', function() {
```
ptolemybarnes: Again, some iterative helper methods would help clean this up a little.

src/bowlingFrame.js
```diff
@@ -0,0 +1,61 @@
...
+    throw new Error('Your must add to frame one first');
+  };
+
+  if (number !== 1 && game.currentGame[number - 1][0] == null) {
+    throw new Error('You must add to frames in correct order');
+  };
+};
+
+BowlingFrame.prototype.rollOne = function(input) {
+  number = this.number;
```
ptolemybarnes: eeegad, global variables!

src/bowlingFrame.js
```diff
@@ -0,0 +1,61 @@
...
+  if (game.currentGame[number][0] != null) {
+    throw new Error('You already have a score here');
+  };
+
+  if (input > 10 || input < 0) {
+    throw new Error('Roll must be between 0 and 10');
+  } else {
+    rollIndex = this.rollIndex;
```
ptolemybarnes: And here! >_<

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,162 @@
+describe('Bowling Game', function() {
+
+  beforeEach(function() {
+    game = new BowlingGame();
```
ptolemybarnes: >_< Global variable.

src/bowlingFrame.js
```diff
@@ -0,0 +1,61 @@
...
+    game.currentGame[number][0] = input;
+  };
+
+  game.currentScore();
```
ptolemybarnes: The only reason this is working is because your test defines a global variable, `game`. This code wouldn't work in production.

src/bowlingFrame.js
```diff
@@ -0,0 +1,61 @@
...
+
+};
+
+BowlingFrame.prototype.rollTwo = function(input) {
+  number = this.number;
```
ptolemybarnes: http://c2.com/cgi/wiki?GlobalVariablesAreBad

src/bowlingFrame.js
```diff
@@ -0,0 +1,61 @@
...
+  } else {
+    rollIndex = this.rollIndex;
+    game.currentGame[number][0] = input;
+  };
+
+  game.currentScore();
+
+};
+
+BowlingFrame.prototype.rollTwo = function(input) {
```
ptolemybarnes: There's quite a lot of duplication between the rollOne and rollTwo functions

src/bowlingGame.js
```diff
@@ -0,0 +1,99 @@
+var BowlingGame = function() {
+
+  this.scoreGrid = {1: null,
```
ptolemybarnes: It might be easier here to use an array.

src/bowlingGame.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+BowlingGame.prototype.currentScore = function() {
+  currentGame = this.currentGame;
+  scoreGrid = this.scoreGrid;
+  for (i = 1; i <= 9; i++) {
+    nextFrame = i + 1;
+    nextNextFram = i + 2;
+    nextNextFrame = i + 2;
+    if (game.isNotSpareOrStrike(i) && game.currentGame[i][1] != null) {
```
ptolemybarnes: All these ifs are really hard to understand. Try to extract out some of this logic to other descriptively-named functions.

src/bowlingGame.js
```diff
@@ -0,0 +1,99 @@
...
+  this.score = 0;
+  for (i = 1; i < 11; i++) {
+    this.score += this.scoreGrid[i];
+  };
+
+  return this.score;
+};
+
+BowlingGame.prototype.isNotSpareOrStrike = function(i) {
+  if ((game.currentGame[i][0] + game.currentGame[i][1]) !== 10 && game.currentGame[i][0] !== 10) {
```
ptolemybarnes: Couldn't you do `(!this.isSpare(i) && !this.isStrike(1))` here?

ptolemybarnes:
Not bad! I agree, it is a little messy. Think about how you can extract some of the logic in the larger functions out to smaller functions, where it can be reused etc.

PR #92, opened 2015-08-31
strikes and spares still not working 

sivanpatel:


spec/BowlingGameSpec.js
```diff
@@ -0,0 +1,97 @@
...
+
+  beforeEach(function(){
+    game = new BowlingGame();
+  });
+
+  it('starts with score = 0', function() {
+    expect(game.score).toEqual(0);
+  });
+
+  it('has 10 frames', function() {
```
ptolemybarnes: I can see why you would write this sort of test but it's not testing the public interface to game so you might want to delete it at some point.

spec/BowlingGameSpec.js
```diff
@@ -0,0 +1,97 @@
...
+
+  it('can only take rolls that are 10 or less', function() {
+    expect(function() { game.roll(11) } ).toThrow(new Error ("Illegal throw"));
+    expect(game.score).toEqual(0);
+  })
+
+  it('a roll in input into the frame scores', function() {
+    game.roll(5);
+    game.roll(4);
+    expect(game.allRolls[0]).toEqual([5,4]);
```
ptolemybarnes: This is breaking encapsulation: you're expectation is against an array that is used internally to game.

spec/BowlingGameSpec.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  it('can only roll twice before moving onto another frame', function() {
+    game.roll(2);
```
ptolemybarnes: It would be really nice if you had a helper method for doing a succession of rolls. So you could do:
```javascript
rolls([2, 2, 2, 2]);

File: spec/BowlingGameSpec.js [hunk 4]
```diff
@@ -0,0 +1,97 @@
...
+    expect(game.allRolls[0]).toEqual([2,2]);
+    expect(game.allRolls[1]).toEqual([2,2]);
+    expect(game.score).toEqual(8);
+  });
+
+  it('only rolls once if a strike is scored', function() {
+  game.roll(10);
+  game.roll(5);
+  expect(game.allFrames[0]).toEqual(10);
```
ptolemybarnes: Again, breaking encapsulation here. You could have a method like `game.getFrame(0)`;

spec/BowlingGameSpec.js
```diff
@@ -0,0 +1,97 @@
...
+    game.roll(2);
+    game.roll(6);
+    game.roll(2);
+    game.roll(6);
+    expect(game.allFrames[0]).toEqual(8);
+    expect(game.allFrames[1]).toEqual(8);
+    expect(game.score).toEqual(16);
+  });
+  //
+  // it('can give bonus points for strikes', function() {
```
ptolemybarnes: Try and get these tests passing and I'll review again :-)

PR #93, opened 2015-08-31
Finished the back-end. Jquery is hard

leon-wee:


ptolemybarnes:
Looks great @leon-wee! :checkered_flag:

PR #94, opened 2015-08-31
Submission for Bowling Challenge

katylouise:


index.html
```diff
@@ -0,0 +1,67 @@
+<!DOCTYPE html>
```
ptolemybarnes: You might use a templating engine (haml / erb) to clean this up a bit. Not sure how that would play with jasmine-jquery though... you would need to precompile. So maybe vanilla HTML is better actually. :+1:

spec/FeatureSpec.js
```diff
@@ -0,0 +1,93 @@
...
+
+  beforeEach(function(){
+    jasmine.getFixtures().fixturesPath = '.';
+    loadFixtures('index.html');
+    $.holdReady(false);
+    scorecard = new ScoreCard();
+  });
+
+  it("displays score in table when a button is clicked", function() {
+    $("button[data-pins=2]").click();
```
ptolemybarnes: ```$("button[data-pins=2]").click();``` <-- this is repeated a lot. Could you have a helper function to clean this up? The interface might look like: ```clickPin(3)```.

spec/FeatureSpec.js
```diff
@@ -0,0 +1,93 @@
+describe('BowlingApp',function(){
```
ptolemybarnes: Overall, great test coverage and well done for using jasmine-jquery. Use variables or helper functions to avoid the fiddly and repetitive references to elements on the DOM.

spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,159 @@
+describe("ScoreCard", function() {
+  var scorecard;
+
+  beforeEach(function() {
+    scorecard = new ScoreCard();
+  });
+
+  it("should keep track of the frame", function() {
```
ptolemybarnes: Avoid starting test specifications with modal verbs like 'should'. For me, the modality is implied by it being a test.

spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,159 @@
...
+  it("is limited to 11 frames", function() {
+    for(var i = 0; i < 11; i++) {
+      scorecard.nextFrame();
+    }
+    expect(scorecard.nextFrame()).toEqual("End of game!");
+  });
+
+  it("can move to the next roll", function() {
+    scorecard.nextRoll();
+    expect(scorecard.roll).toEqual(2);
```
ptolemybarnes: Not sure about the naming here: what's the logical connection between ```roll``` and 2?

spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  it("can return the points for a frame", function() {
+    scorecard.scoreForRoll(4);
+    scorecard.scoreForRoll(3);
+    expect(scorecard.scoreForFrame(1)).toEqual(7);
```
ptolemybarnes: I like how you're providing an interface allowing me to get the score of each frame here.

spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  it("adds points to scores each roll", function() {
+    scorecard.scoreForRoll(4);
+    scorecard.scoreForRoll(5);
+    expect(scorecard.scores[1]).toEqual([4, 5]);
```
ptolemybarnes: This, however, is breaking encapsulation: we're reaching into scorecard to extract a particular value (an array) from its scores array. This would be much better done by an interface as you've done above (line 37).

spec/ScoreCardSpec.js
```diff
@@ -0,0 +1,159 @@
...
+
+  it("only gives a third roll after a spare or strike in the tenth frame", function() {
+    scorecard.frame = 10;
+    scorecard.scoreForRoll(4);
+    scorecard.nextRoll();
+    scorecard.scoreForRoll(3);
+    scorecard.nextRoll();
+    expect(scorecard.roll).toEqual(1);
+  });
+});
```
ptolemybarnes: This is pretty good! However, as we discussed, I do think #nextRoll should not be part of the public interface to ScoreCard.

src/BowlingApp.js
```diff
@@ -0,0 +1,110 @@
+$(document).ready(function() {
+  var scorecard = new ScoreCard();
+
+  $("#play_again").click(function() {
+    $("#game_over").hide();
+    location.reload();
+  });
+
+  $("button").click(function(event) {
```
ptolemybarnes: As discussed, it would be great to break some of the stuff in here out to separate functions.

src/BowlingApp.js
```diff
@@ -0,0 +1,110 @@
...
+      gameOver();
+    }
+
+    var score = scorecard.scoreForRoll($(this).data("pins"));
+    var frame = scorecard.frame;
+    $("td[id='" + frame + "']").append('<td>' + score +'</td>');
+    hideButtons(scorecard.pinsLeft(score));
+
+    //check for spare on previous frame
+    if (frame > 1 && scorecard.isPreviousFrameSpare()) {
```
ptolemybarnes: Have a look at the "Tell, Don't Ask" principle of OOP for more on why this is undesirable.

src/ScoreCard.js
```diff
@@ -0,0 +1,77 @@
+function ScoreCard() {
+  this.frame = 1;
+  this.pins = 10;
+  this.roll = 1;
+  this.scores = {1:[], 2:[], 3:[], 4:[], 5:[], 6:[], 7:[], 8:[], 9:[], 10:[]};
```
ptolemybarnes: It seems like you're getting a vanilla js object to behave like an array here. Why not just use an array?

ptolemybarnes:
Nice attempt!! :smile: 
We had a code review of this with @natstar93 and discussed how this might be refactored. At the moment Scorecard handles all of the business logic. It would be good to break this up into separate classes. Consider having a Frame class, for example.
Also, the code doesn't quite meet the specification as you can't get a perfect game. Try to write a test for this and make it pass. :construction:

PR #95, opened 2015-08-31
Bowling challenge submission

christopheralcock:
Engine, no front end


ptolemybarnes:
Nice work! We had a review of this with @natstar93 and discussed the possibility of splitting out Scorecard into separate classes. Also, it seems to me that the tests are not limited to Scorecard's public interface. This makes them potentially quite brittle: if I decide to change the internal implementation of ScoreCard, the tests might break even if the behaviour is the same. Check out the chapter on testing in POODR (or any Sandy Metz lecture on the subject) for a lengthy explanation about why we should exercise only the public interface at the unit test level.
:checkered_flag:

ptolemybarnes:
Looks great @christopheralcock , though failing CI:
image
Looks like an easy fix. :-)

christopheralcock:
passing now, ta

christopheralcock:
(just straight up had to remove the featurespec - i dont have a front end, it was just a hangover of when i started this outside-in)

PR #100, opened 2015-09-01
Not final version - please don't review until version from home computer is comitted!!

UpstatePedro:


ptolemybarnes:
Looks like CI is still failing! Have you any insight into why?

PR #106, opened 2015-10-03
Usman - Bowling Challenge

UsmanJ:


spec/GameSpec.js
```diff
@@ -0,0 +1,156 @@
...
+  it("should start with zero points", function() {
+    expect(game.score).toEqual(0);
+  });
+
+  it("should be able to increase the score when ball is rolled", function() {
+    game.rollBall(4);
+    expect(game.score).toEqual(4);
+  });
+
+  it("should change frame after two rolls", function() {
```
sdorunga1: Confusing spec name

spec/GameSpec.js
```diff
@@ -0,0 +1,156 @@
...
+  it("frame score resets when new frame starts", function() {
+    game.rollBall(6);
+    game.rollBall(4);
+    game.rollBall(4);
+    game.rollBall(6);
+    expect(game.frameScore).toEqual(0);
+  });
+
+  it("can roll upto two additional balls if strike on last frame", function() {
+    game.rollBall(10);
```
sdorunga1: You might want to DRY this up by doing something like this:

```
 function rollBalls(score, numberOfRolls) {
   for(i=0, i< numberOfRolls, i++) {
     game.rollBall(score);
   }
}
```

spec/GameSpec.js
```diff
@@ -0,0 +1,156 @@
...
+    game.rollBall(10);
+    game.rollBall(10);
+    game.rollBall(10);
+    game.rollBall(10);
+    game.rollBall(10);
+    game.rollBall(1);
+    game.rollBall(1);
+    game.rollBall(1);
+    expect(game.scoreBoard[10][2]).toEqual(undefined);
```
sdorunga1: The more important thing seems to be taht the game would be over

src/Game.js
```diff
@@ -0,0 +1,90 @@
...
+  this.spareBall = false;
+  this.strikeBall = false;
+  this.tenthFrameBonus = false;
+  this.gameOver = false;
+  this.scoreBoard = {1:[], 2:[], 3:[], 4:[], 5:[], 6:[], 7:[], 8:[], 9:[], 10:[]}
+};
+
+Game.prototype.rollBall =  function(hit) {
+
+  if ((this.frameNumber < 10) && (this.firstThrow === false) && (this.frameScore + hit > 10)) {
```
sdorunga1: You coud extract this out into a function, with a more explicit name

src/Game.js
```diff
@@ -0,0 +1,90 @@
...
+    this.scoreBoard[this.frameNumber].push(hit*2);
+    this.score += (hit*2);
+    this.spareBall = false;
+  } else {
+    this.scoreBoard[this.frameNumber].push(hit);
+    this.score += (hit);
+  };
+};
+
+Game.prototype.isFirstThrow = function() {
```
sdorunga1: It's not a very good idea to name things that change state with `is` because that suggests that you are checking something

src/Game.js
```diff
@@ -0,0 +1,90 @@
...
+
+  } else if (this.frameNumber < 10) {
+      this.addToScoreBoard(hit);
+      this.frameScore += hit;
+      this.isSpare(hit);
+      this.isFirstThrow();
+      this.isStrike(hit);
+      this.resetFrameScore();
+
+  } else if (this.frameNumber === 10 && this.gameOver === false) {
```
sdorunga1: This condition could be extracted to an `isFinalFrame` function

PR #116, opened 2015-10-09
Harriet's Bowling Challenge

harrietc52:


src/BowlingScore.js
```diff
@@ -0,0 +1,59 @@
...
+      currentFrame.push(this.rolls[i])
+      this.frameScores.push(currentFrame);
+    };
+  };
+};
+
+BowlingScore.prototype.frameBonus = function () {
+  for (i = 0; i < this.frameScores.length; i++) {
+    if (this.frameScores[i][0] === 10 && this.frameScores[i+1][0] != 10) {
+      this.bonusScores.push(this.frameScores[i+1][0]);
```
stephenlloyd: bit of duplication of this. Could turning it into a function with the differences passed as arguements clean it up.

src/BowlingScore.js
```diff
@@ -0,0 +1,59 @@
...
+BowlingScore.prototype.frameTotal = function () {
+  for (i = 0; i < this.frameScores.length; i++) {
+    this.frameTotals.push(this.frameScores[i][0] + this.frameScores[i][1]);
+  };
+};
+
+BowlingScore.prototype.scores = function () {
+  this.score = (this.frameTotals.reduce(function (x,y) {
+      return x+y
+    })
```
stephenlloyd: x + y would be bar in ruby. I'm persoanlly ok with it but others might say you should be more expressive.

stephenlloyd:
I really like you're approach and i haven't seen it before. I'm going to chat you about it purely because I like it. Is there a reason you don't have the last frame or have a final test adding 12 strikes to 300? Otherwise it looks really good and if you came to this solution yourself you've got a great programming mindset :)

stephenlloyd:
:construction:

PR #118, opened 2015-10-10
Initial submit - refactoring to be done

aaronkendall:


spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+  beforeEach(function() {
+    bowling = new Bowling();
+  });
+
+  describe("Frames", function() {
+    it("frame count to be equal to one at the start of a game", function() {
+      expect(bowling.frameCount).toEqual(1);
+    });
+
+    it("starts with ten empty frame arrays", function() {
```
sdorunga1: This test does not do what it says

spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+      expect(bowling.frames[1]).toEqual([0,0]);
+    });
+
+    it("updates frameCount to next frame", function() {
+      bowling.updateFrame();
```
sdorunga1: This doesn't really update the frame

spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+      expect(bowling.frameCount).toEqual(2);
+    });
+
+    it("returns true when it is the 10th frame", function() {
+      bowling.frameCount = 10;
```
sdorunga1: This isn't really what a user of the class would do

spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+  });
+
+  describe("Score", function() {
+    it("starts with a score of 0", function() {
+      expect(bowling.score).toEqual(0);
+    });
+
+    it("bowl should add scores to the relevant frame array", function() {
+      bowling.bowl(3,6);
```
sdorunga1: For the purposes of thiss application you don't really care about this

spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    it("should update the score with the values bowled", function() {
+      bowling.bowl(3,6);
+      expect(bowling.score).toEqual(9);
+    });
+
+    it("returns true if previous frame was a strike", function() {
+      bowling.bowl(10,0);
+      expect(bowling.isStrike()).toBe(true);
```
sdorunga1: `isSrike` could be `lastFrameIsStrike`

spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+
+    it("adds two scores to the 10th frame array when it is the 10th frame but no bonus", function() {
+      bowling.frameCount = 10;
+      bowling.tenthFrameBowl(2,3);
+      expect(bowling.score).toEqual(5);
+    });
+
+    it("adds three scores to the 10th frame array if a strike is scored in last frame", function() {
+      bowling.frameCount = 10;
+      bowling.tenthFrameBowl(10,5,6);
```
sdorunga1: Same as above don't have code that's specifically for testing

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
+function Bowling() {
+  this.frameCount = 1;
```
sdorunga1: something like `frameCount` should be dynamically counted rather than tracked as state

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+                };
+  this.score = 0;
+};
+
+  Bowling.prototype.updateFrame = function() {
+    this.frameCount++;
+  };
+
+  Bowling.prototype.bowl = function(val1,val2,val3) {
+    this.checkScores(val1,val2);
```
sdorunga1: `checkRollsValid`

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+    this.frames[this.frameCount][0] = val1;
+    this.frames[this.frameCount][1] = val2;
+    if(this.isTenthFrame() == true){
```
sdorunga1: `if(this.isTenthFrame){}`

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+  Bowling.prototype.tenthFrameBowl = function(val1,val2,val3) {
+    this.frames[this.frameCount][0] = val1;
+    this.frames[this.frameCount][1] = val2;
+    this.frames[this.frameCount][2] = val3;
+    this.updateScoreTenth(val1,val2);
+    if(this.finalRoll()){this.score += val3};
+    return "game over";
+  };
+
+  Bowling.prototype.finalRoll = function() {
```
sdorunga1: something like `bonusRollAvailable`

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+    if(this.frames[10][0] == 10 || this.frames[10][0] + this.frames[10][1] == 10){
```
sdorunga1: You could use similar logic to `isStrike` / `isSpare`

sdorunga1: `return (this.frames[10][0] == 10 || this.frames[10][0] + this.frames[10][1] == 10)`

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+      return true;
+    };
+  };
+
+  Bowling.prototype.isTenthFrame = function() {
+    if(this.frameCount == 10){return true};
```
sdorunga1: Same as above

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  Bowling.prototype.updateScoreTenth = function(val1,val2) {
+    if(this.isStrike() == true && val1 == 10){
```
sdorunga1: Don't need to compare to true. Maybe name `val1` something `roll1`

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+      var orig = this.frames[this.frameCount - 1][0];
```
sdorunga1: orig should prob be `ninthFrame`

spec/BowlingSpec.js
```diff
@@ -0,0 +1,92 @@
...
+      bowling.frameCount = 10;
+      bowling.tenthFrameBowl(5,5,5);
+      expect(bowling.score).toEqual(15);
+    });
+
+    it("throws error if inputted values are higher than 10", function() {
+      expect( function() {bowling.bowl(5,6); } ).toThrow(new Error("frame score cannot exceed 10"));
+    });
+  });
+
```
sdorunga1: You might want a perfect game test

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+      this.frames[this.frameCount - 1][0] += val1 * 2;
```
sdorunga1: Extract this logic into something like `applyStrike/SpareBonus

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+    else{
+      this.score += val1 + val2;
+    };
+  };
+
+
+  Bowling.prototype.isStrike = function() {
+    if(this.frameCount > 1){
+      if(this.frames[this.frameCount - 1][0] == 10){return true}
```
sdorunga1: this whole if can be just `this.frames[this.frameCount - 1][0] == 10`

src/Bowling.js
sdorunga1: You should probably count from 0 just to make array lookups easier

src/Bowling.js
```diff
@@ -0,0 +1,101 @@
...
+      else{return false};
+    };
+  };
+
+  Bowling.prototype.isSpare = function() {
+    if(this.frameCount > 1){
+      if(this.frames[(this.frameCount) - 1][0] + this.frames[(this.frameCount) - 1][1] == 10 && this.frames[(this.frameCount) - 1][1] > 0){
```
sdorunga1: Same as above, plus some of these truth checks i'm sure are a concept

PR #123, opened 2015-10-11
Tim Hyson's Bowling Challenge

timhyson:


sdorunga1:
As a first bit of general feedback. I wouldn't have gone with writing a UI for this before the whole functionality was complete

spec/bowlingScoreSpec.js
```diff
@@ -0,0 +1,117 @@
+describe('BowlingScore', function() {
+  var game = new BowlingScore();
```
sdorunga1: This should be wrapped into a `before` block, as right now you're creating a state dependency between your tests. For example if you move one test out of order it will break another one. You want each test to stand as an individual statement about what your application can do

spec/bowlingScoreSpec.js
```diff
@@ -0,0 +1,117 @@
...
+      game.addNewBonusScore();
+      game.makeRoundScores();
+      expect(game.roundScores).toEqual([[9], [9], [9], [19], [9]]);
+    });
+  });
+
+  describe('#makeGameScore', function() {
+    it('calculates gameScore by totalling roundScores', function() {
+      game.makeGameScore();
+      expect(game.gameScore).toEqual([55]);
```
sdorunga1: I would expect `gameScore` to just be a number

src/bowlingScore.js
```diff
@@ -0,0 +1,61 @@
+function BowlingScore() {
+  this.rawScores = [];
+  this.frameScores = [];
+  this.bonusScores = [];
+  this.roundScores = [];
```
sdorunga1: What is a round?

spec/bowlingScoreSpec.js
```diff
@@ -0,0 +1,117 @@
...
+
+  describe('#makeFrameScores', function() {
+    it('restructures rawScores into frameScores', function() {
+      game.makeFrameScores();
+      expect(game.frameScores).toEqual([[6, 3], [7, 2], [4, 5], [10, null]]);
+    });
+  });
+
+  describe('#makeRoundScores', function() {
+    it('calculates roundScores by adding frameScores to bonusScores', function() {
```
sdorunga1: Try to avoid explaining how functionality works and focus more on the outcome.

spec/bowlingScoreSpec.js
```diff
@@ -0,0 +1,117 @@
...
+
+    it('correctly pushes second frame into raw score array', function() {
+      game.addNewRollScore(7);
+      game.addNewRollScore(2);
+      game.addNewRollScore(4);
+      game.addNewRollScore(5);
+      expect(game.rawScores).toEqual([6, 3, 7, 2, 4, 5]);
+    });
+
+    it('correctly pushes strike into raw score array', function() {
```
sdorunga1: You don't necessarily care about this in terms of the functionality of this object. This is more of an internal thing. What you want this object to be able to do is to calculate a total score.

PR #141, opened 2015-11-13
testing jshint config

tansaku:


danldb:
Lots of semi-colon noise. I suppose it may push them to use jshint on their text editor, but I'm not sure that this is worth potentially masking actual code smells in a sea of semicolon messages.

PR #158, opened 2015-11-22
bowling_challenge

Alaanzr:


tansaku:
You had your bowling_challenge reviewed by **Adrian Booth**.
### The good points are:
* All unnecessary components removed
* Appropriate use of describe blocks, Tests have descriptive titles, Repetition is DRYed to beforeEach functions, Operation of code clear from tests themselves
* Rules are modeled correctly
* Functions are named for the actions they do
* Open for extension, Closed for modification
* Repetition refactored out
* Any duck punching is done carefully

* Little or no testing of edge cases
* Use Spec 'Helpers' Appropriately
* Poor Encapsulation

### Additional comments:
Too many magic numbers that could be placed in the function constructer.
For loop to create frame array rather than coding in the 10 frames. May be hard coding but the frame amount of bowling does not change.


see https://github.com/makersacademy/bowling_challenge/blob/master/docs/review.md for more details on the items above

PR #162, opened 2015-11-22
Mari-Ann's Weekend Challenge

mariann013:


tansaku:
You had your bowling_challenge reviewed by **Dom**.
### The good points are:
* Tests have descriptive titles, Repetition is DRYed to beforeEach functions, Operation of code clear from tests themselves
* Edge cases are tested effectively
* Rules are modeled correctly
* Internal variables are encapsulated effectively
* Functions are named for the actions they do
* Responsibilities separated over multiple objects, Dependency Injection used appropriately, Open for extension, Closed for modification
* Repetition refactored out
* Any duck punching is done carefully

* Unnecessary Components should be Removed
* Use Spec 'Helpers' Appropriately

### Additional comments:
Very impressive, a lot better than my attempt at this challenge. Only slight gripe i can raise is that there are some indentation issues in the frame and bowling spec. all good apart from that


see https://github.com/makersacademy/bowling_challenge/blob/master/docs/review.md for more details on the items above

PR #163, opened 2015-11-22
bowling challenge

jrose111284:


tansaku:
You had your bowling_challenge reviewed by **Rob Youngman**.
### The good points are:
* Appropriate use of describe blocks, Tests have descriptive titles, Repetition is DRYed to beforeEach functions, Operation of code clear from tests themselves
* Edge cases are tested effectively
* Spec 'Helpers' used appropriately
* Rules are modeled correctly
* Internal variables are encapsulated effectively
* Dependency Injection used appropriately
* Repetition refactored out

* Unnecessary Components should be Removed
* Poor Function Naming
* Duck Punch carefully

### Additional comments:
Could use DRY in test cases.

Break out code into more than one object.

More instructions in the read me.

Final frame rules are missing or not correct.


see https://github.com/makersacademy/bowling_challenge/blob/master/docs/review.md for more details

PR #166, opened 2015-11-23
pull request for bowling challenge

ChukaEbi:


MoeSadoon:
2) Make sure you delete the player.js testing spec, its going to confuse people downloading this if you've still got jasmine test files that are testing songs :)

MoeSadoon:
3) You have some decent tests Chuka, but you should also implement tests which check for unique cases in a game too, eg:
 - it should roll a gutter game (everything is 0)
-  it should roll a perfect game
-  it should have a score of 12 after two games , etc

MoeSadoon:
4) Consider cleaning up your 'roll' function, perhaps extract a final score function out of it instead of populating the whole lot in the same function

PR #274, opened 2016-05-22
Bowling challenge

lexiht:


danldb:
:eyes:

Nested if statements - alarm bells should be ringing in your head when you see these. How can you refactor to remove them? You could use an object to map out the logic. You could use polymorphism to inject the correct functionality. Maybe you could use another class with a simpler responsibility, and compose your domain in a modular fashion. There is always a way to avoid these nested if statements, and it will really help you in your job search if you can hone in on this refactoring skill.

PR #276, opened 2016-05-22
Backend completed

letianw91:


danldb:
:eyes:

Watch out for if/else statements - conditionals can be a sign of a missing class, particularly when nested.

I really liked the simplicity of your jQuery controller.

PR #281, opened 2016-05-22
bowling challenge

MaxProvin:


src/Game.js
```diff
@@ -0,0 +1,90 @@
+'use strict'
+
+function Game() {
+  this._frames = {1:[], 2:[], 3:[], 4:[], 5:[], 6:[], 7:[], 8:[], 9:[], 10:[]}
```
danldb: just use an array, it is accessed in exactly the same way, but indexed at 0.

PR #290, opened 2016-05-23
Work done by Monday 23th

Adaymesa:


spec/frameSpec.js
```diff
@@ -0,0 +1,25 @@
+describe("Frame", function() {
+  var frame
+  var game
+
+  beforeEach(function () {
+    frame = new Frame(game)
+    game = new Game(frame)
```
danldb: using classical add instead of london style

danldb:
:eyes:

I like this solution a lot. Simple objects with clear responsibilities.

PR #291, opened 2016-05-23
Bowling Challenge

nmrichards:


src/bowling.js
```diff
@@ -0,0 +1,90 @@
...
+Bowling.prototype.calculateScore = function() {
+  var self = this;
+  self._score = 0;
+  if (self._frames.length > 0){
+    self._frames.forEach(function(frames,i){
+      if(self.isStrike(frames)){
+        self.strikeScore(i)
+      }else if(self.isSpare(frames)){
+          self.spareScore(i)
+        }else{
```
danldb: pyramid of doom! nested if/else conditional will cause headaches when you need to make any changes to this class.

danldb:
:eyes:

Watch your complexity levels - nested if/else is a huge code smell, and there is always a better way. It could be that you are missing an object - maybe you could build a map using a js object?

PR #361, opened 2016-10-29
Let's go bowling!

MalinnaLeach:


public/js/Game.js
```diff
@@ -0,0 +1,107 @@
...
+  if (!this.gameOver) {
+    this.resetPins(hits);
+    this.resetMode(hits, this.scoreMode);
+    this.resetRolls(hits);
+    this.lastRollScore = hits;
+  }
+};
+
+Game.prototype.score = function(hits) {
+  if (this.scoreMode === "normal") {
```
maryrosecook: Can `"normal"` be a constant?

public/js/Game.js
```diff
@@ -0,0 +1,107 @@
...
+
+Game.prototype.hits = function() {
+  return Math.floor(Math.random() * (this.pins + 1));
+};
+
+
+Game.prototype.roll = function() {
+  var hits = this.hits();
+  this.score(hits);
+  this.resetStatus(hits);
```
maryrosecook: Could `resetStatus` have a clearer name?

public/js/Game.js
```diff
@@ -0,0 +1,107 @@
...
+   this.currentFrame = 1;
+   this.currentRoll = 1;
+   this.lastRollScore = 0;
+   this.pins = 10;
+   this.scoreMode = "normal";
+   this.gameOver = false;
+ }
+
+Game.prototype.hits = function() {
+  return Math.floor(Math.random() * (this.pins + 1));
```
maryrosecook: It's probably worth passing in each roll (eg 5) rather than generating it inside the game.  Or, if you'd like to randomly generate the roll, it might be worth moving the code out of the Game - it feels like a separate responsibility.

public/js/Game.js
```diff
@@ -0,0 +1,107 @@
+ function Game(){
+   this.scoreTotal = 0;
+   this.currentFrame = 1;
+   this.currentRoll = 1;
+   this.lastRollScore = 0;
+   this.pins = 10;
```
maryrosecook: Is it possible store every roll, and from that be able to derive `gameOver`, `currentRoll`, `currentFrame` etc from the rolls? This would reduce the coordination necessary inside Game.

public/js/Game.js
maryrosecook: Game feels like a large unit.  Could some of its responsibilities be moved to another unit? (For example, by refactoring to keep every roll that has happened and then refactoring some of the logic for managing the scoring of frames to a new unit?)

PR #362, opened 2016-10-29
Bowling challenge

wirsindpapst:


src/game.js
```diff
@@ -0,0 +1,99 @@
+"use strict"
+
+function Game() {
+  this._game = []
```
silvabox: shouldn't this be `_frames`?

src/game.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+};
+
+Game.prototype.calculateSpareBonus = function(){
+  if (this.returnLastElement(2, this.spareOrStrike) === "spare") {
+    this.returnLastElement(2, this.bonuses).push(this.returnLastElement(1, this._game)[0]);
+  }
+};
+
+Game.prototype.calculateStrikeBonus = function(){
```
silvabox: This function is virtually unreadable!  Even if the logic is correct within the context of the design, it's incredibly difficult for a human to parse.

src/frame.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+function Frame(game) {
+  this.score = []
+}
+
+Frame.prototype.bowlFrame = function(game){
+  var bowl = Math.floor((Math.random() * 11));
```
silvabox: there was no requirement to randomly generate the values passed to the score calculator.  If there was, then this would still be a questionable design; ideally you should separate the generation of the random number from storing it in the frame.

PR #365, opened 2016-10-29
Bowling Score Card

crsanghani:


src/scorecard.js
```diff
@@ -0,0 +1,78 @@
...
+
+Scorecard.prototype.getCurrentFrame = function () {
+  return this.frame;
+}
+
+Scorecard.prototype.getCurrentRoll = function () {
+  return this.roll;
+}
+
+Scorecard.prototype.updateScore = function (number) {
```
silvabox: you are using the argument name `number` in a lot of your functions, but it's not consistent with the domain language - you don't often use the term `number` when bowling.

PR #366, opened 2016-10-29
Frances Maxwell Bowling You Over

francesmx:


src/Frame.js
```diff
@@ -0,0 +1,49 @@
+function Frame() {
+  this.score = 0;
```
silvabox: Avoid storing multiple versions of the truth.  Strikes, spares, completion and scores are all derived from the other properties - so these should all be functions that calculate the answers on the fly.

PR #367, opened 2016-10-30
Added two test for gutter and first roll

amidabrian51:
Added test with Jasmine. Did a test for a gutter roll and a first roll that returned zero.


src/Bowling.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+function Bowling (){
+  this.score = 0;
+};
```
amidabrian51: Thanks

src/interface.js
```diff
@@ -0,0 +1,3 @@
+$(document).ready(function() {
+  var thermostat = new Bowling();
```
silvabox: ahem, er, `thermostat`...?

src/Bowling.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+    result += getNormalScore();
+    rollIndex += 2;
+    }
+
+  }
+
+  return result;
+
+  function isStrike() {
```
silvabox: Did you mean for these functions to be defined inside the `score` function?

PR #374, opened 2016-10-30
bowling challenge

antonyCastineiras:


silvabox:
Hi Antony,
Once again, I can't review your code because your text editor is still using tabs for indentation.  I'm also worried that there is only one commit.  You need to be committing much more frequently; trust me, that's not a lesson you want to learn the hard way.

PR #380, opened 2016-10-31
Challenge incompleted

michkles:
Solution coming from the source described in the files.


silvabox:
Hi Michal, thank you for referencing the source files in your solution.  Did you attempt your own approach before going to the video?

michkles:
Hi Ben, yes I did attempt to do it on my own but I thought it would be to
difficult. Then I decided to follow the steps from that video. When I
realised that I am not really progressing by writing somebody's else code I
decided to stop doing the challenge. I wanted to finish it next week and
write everything from scratch but didn't have enough time. Now after
spending more time with JavaScript I know the challenge is not as difficult
as it seemed to be in the beginning.

silvabox:
OK, thanks for letting me know.  I'm glad to hear you feel better equipped to approach the challenge now.

PR #381, opened 2016-10-31
Bowling Challenge

B-raw:


src/interface.js
```diff
@@ -0,0 +1,113 @@
+$(document).ready(function() {
+  var game = new Game();
+
+  $('#total_score').text(game.calculateScore());
+
+  $('#zero').on('click', function() {
```
silvabox: you don't need to repeat this 11 times as the behaviour is the same for all.  Use a class to add a `click` handler to all the buttons at once and pass the number of skittles to `submitSkittles` instead of a CS selector.

PR #383, opened 2016-10-31
Not finished, still need to sort out bonus for a strike/spare 

James-SteelX:


silvabox:
Hi James, I can't review your code as your CI is failing.  I think there is an incompatibility between the config and your folder structure.

PR #387, opened 2016-11-06
Weekend challenge 5

DagmaraSz:


app/game.js
```diff
@@ -0,0 +1,86 @@
...
+    pins_down += this.getRoll(i+1);
+    if (this.bonus) {
+      this.applyBonus(pins_down);
+    };
+  };
+  this._score += pins_down;
+  if (pins_down === this._pins) this.setBonus();
+};
+
+Game.prototype.getRoll = function(i) {
```
maryrosecook: Watch out for this function's name. It implies that it just returns a value, but it also changes data.

app/game.js
```diff
@@ -0,0 +1,86 @@
...
+  this._balls -= 1;
+  return this.results[i];
+};
+
+Game.prototype.roll = function () {
+  this._balls -= 1;
+  return _.sample([1,2,3,4,5,6,7,8,9,10]);
```
maryrosecook: It would probably be better to write your own version of sample, rather than include the whole of Underscore in order to use this one function.

app/game.js
```diff
@@ -0,0 +1,86 @@
...
+    this.applyBonus(pins_down);
+  };
+  if (pins_down < this._pins) {
+    pins_down += this.getRoll(i+1);
+    if (this.bonus) {
+      this.applyBonus(pins_down);
+    };
+  };
+  this._score += pins_down;
+  if (pins_down === this._pins) this.setBonus();
```
maryrosecook: Try to always use this format for conditionals:

```
if (...) {
  ...
}
```

app/game.js
```diff
@@ -0,0 +1,86 @@
...
+};
+
+Game.prototype.getRoll = function(i) {
+  this._balls -= 1;
+  return this.results[i];
+};
+
+Game.prototype.roll = function () {
```
maryrosecook: Is this function used anywhere?

PR #389, opened 2016-11-06
JS Bowling Challenge

fmlharrison:


src/Frame.js
```diff
@@ -0,0 +1,112 @@
...
+
+Frame.prototype._isNotStrike = function () {
+  return this.firstRoll < 10;
+};
+
+Frame.prototype._isSecondRollNeeded = function () {
+  return ((this._isNotStrike() && this._isNotLastFrame()) || this._isLastFrame());
+};
+
+Frame.prototype._isThridRollNeeded = function () {
```
maryrosecook: (Typo aside) I love this function! So clear.

src/Frame.js
```diff
@@ -0,0 +1,112 @@
+var Frame = function (game) {
```
maryrosecook: Really love that you've extracted the frame logic out from the game.  Is there a way to make it not know about the game? Have a look at the exemplar for how to tell frames about the other frames they need to know about.

src/BowlingGame.js
```diff
@@ -0,0 +1,66 @@
+var BowlingGame = function () {
+  this.frames = [];
+  this.totalScore = 0;
+  this._gameSetup();
+};
+
+BowlingGame.prototype.totalScoreBonuses = function () {
+  this._frameScoreBonuses();
+  var allBonuses = this.frames.reduce(function (totalBonus, frameBonus) {
```
maryrosecook: This line could be `return this.frames...` to avoid having to create a new var that is returned.

src/BowlingGame.js
```diff
@@ -0,0 +1,66 @@
...
+    return totalBonus + frameBonus.bonusScore;
+  }, 0);
+  return allBonuses;
+};
+
+BowlingGame.prototype.totalBaseScore = function () {
+  this._framesBaseScores();
+  var allScores = this.frames.reduce(function (totalScore, frameScore) {
```
maryrosecook: Could the code have a `sum` function that could take an array of numbers and return a sum? This function could be used by `totalBaseScore` and `totalScoreBonuses`.

PR #392, opened 2016-11-06
Bowling Weekend Challenge

Swinston88:


spec/frameSpec.js
```diff
@@ -0,0 +1,59 @@
...
+
+  it('updates the frame score for two rolls', function() {
+    frame.bowl(6);
+    frame.bowl(3);
+    expect(frame.score).toEqual(9);
+  });
+
+  it('allows a maximum of 10 pins to be knocked down per frame', function() {
+    frame.bowl(6);
+    expect(function(){frame.bowl(5)}).toThrowError('Please re-enter correct score')
```
silvabox: Maybe the error message could be more specific?

src/frame.js
```diff
@@ -0,0 +1,34 @@
...
+Frame.prototype.bowl = function(pins) {
+  if (this.score + pins > 10) {
+    throw new Error('Please re-enter correct score');
+  }
+  this.rolls.push(pins);
+  this.score += pins;
+  this.assignBonus();
+};
+
+Frame.prototype.isEndFrame = function() {
```
silvabox: `isEndFrame` says to me that this function tells me whether this frame is the last frame in the game, rather than that this frame is over.  Just `isOver` would be much more intuitive.

src/frame.js
```diff
@@ -0,0 +1,34 @@
...
+
+Frame.prototype.isStrike = function() {
+  return this.rolls[0] === 10;
+};
+
+Frame.prototype.isSpare = function() {
+  return this.rolls[0] + this.rolls[1] === 10;
+};
+
+Frame.prototype.assignBonus = function() {
```
silvabox: This is pointless.  There is no need to store two versions of the truth.  Since `strike` and `spare` are both derived, they are unnecessary.  If I want to know if a frame is a strike, I can just call `isStrike()`.

spec/gameSpec.js
```diff
@@ -0,0 +1,76 @@
...
+    expect(game.frameNum).toEqual(2);
+  });
+
+  describe('Bonus Scores', function() {
+
+    beforeEach(function() {
+      game.recordRoll(9);
+      game.recordRoll(1);
+      game.recordRoll(3);
+      game.addSpareBonus();
```
silvabox: It should not be necessary for me to call `addSpareBonus()` manually - the game and frames should handle that automatically.

src/game.js
```diff
@@ -0,0 +1,50 @@
...
+Game.prototype.addScore = function() {
+  this.allFrames.push(this.currentFrame);
+  this.totalScore += this.currentFrame.score;
+};
+
+Game.prototype.isOver = function() {
+  return this.allFrames.length === 10
+};
+
+Game.prototype.addSpareBonus = function() {
```
silvabox: what about strikes?

PR #398, opened 2016-11-07
Unable to get logic to pass Jasmine tests when calculating a strike followed by a spare.

stejgregory:


src/rolls.js
```diff
@@ -0,0 +1,82 @@
...
+  this.currentGame = [];
+  this.score = 0;
+  this.frameNum = 1;
+  this.rollOneTaken = false;
+  this.rollOneScore = 0;
+  this.rollTwoScore = 0;
+  this.frameScore = 0;
+}
+
+var strike = 10
```
silvabox: these should not be defined at the top level; they are essentially now global variables.

src/rolls.js
```diff
@@ -0,0 +1,82 @@
...
+  } else {
+    this.currentGame.push(this.rollOneScore + this.rollTwoScore)
+  }
+  this.addingFramesUp();
+};
+
+
+//CALCULATING THE SCORE
+
+Rolls.prototype.addingFramesUp = function() {
```
silvabox: This is way too complicated.

src/rolls.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+function Rolls() {
```
silvabox: Use domain language to name your objects.  We don't talk about _rolls_ in bowling, we talk about frames and games.

PR #408, opened 2016-11-07
tests and some methods

DavidAMAnderson:
happy with the tests, but solution is nowhere near complete


src/Game.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+Game.prototype.addFrame = function(frame) {
+  var a = this.scorecard;
+  var b = frame.scorecard;
+  if(frame.length != 2) {
+    throw new Error('frame is not valid');
+  }
+  else {
+    a.push(b);
```
silvabox: you are breaking encapsulation here by storing the inner array instead of the frame object.  What happens when you come to calculate the total score?

PR #445, opened 2017-02-05
Tamar's Bowling Challenge

tamarlehmann:


neoeno:
Hi Tamar!

Nice work here! I especially like how you had a go at tweaking the travis config to get it to work. Good decision to try something — good decision to not get too wrapped up in it when it didn't have any impact. If we had more time (or, perhaps, if you really liked making Travis work) we might have pushed further.

A few notes:

* I really like the fact there's a web interface here. It's not overly complicated but it makes it easier to see if your code is doing what you expect it to.
* We get the same effect in your tests too — which I feel pretty good about. I really like the `describe` and `it` descriptions — they make the tests and the code much easier to understand. I also like how we've used `for` loops to remove needless repetition from the tests. We could take this even further by creating a utility method in our tests, something like `repeatRoll({value: 5, times: 10});`
* I'd spend a bit of time focusing on consistent formatting, indentation for example. It matters less what the formatting is, but more than it's totally consistent, without a single inconsistency. The difference in professional appearance is massive and it's an easy way to disqualify yourself in a tech interview. Giving this a read might be good motivation: http://stackoverflow.com/questions/127916/is-programming-style-important-how-important

If you feel the desire, spend a bit more time on this project. Don't worry about implementing any new functionality, don't change the behaviour, but focus just on refactoring and improving readability. Make the code the best possible code you can. As a beginner it's sometimes hard to find the time to focus on that when we keep asking functionality of you! But it's super important (and, I find, sometimes, a bit relaxing — might just be me).

Good work here Tamar — keep it up!

PR #448, opened 2017-02-05
end of 1st weekend 

frankiefy:


frankiefy:
resubmitted pull request, still haven't managed to get strikes and spares to work properly

PR #454, opened 2017-02-09
Challenge submission

kathicks:


src/js/model/Score.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+Score.prototype.scorePins = function(roll, current_frame) {
+  if (current_frame.isNew()) { this._scoreboard.push(roll); }
+  else { this._scoreboard[this._scoreboard.length - 1] += roll; }
+}
+
+Score.prototype.scoreSpare = function(roll, current_frame, last_frame) {
+  if (typeof last_frame !== 'undefined') {
+    if (last_frame.isSpare() && current_frame.isNew()) { this._scoreboard[this._scoreboard.length - 2] += roll;; }
```
neoeno: The double semicolon! Extra safe. Consider dropping the habit though — maybe `jshint` would help by nagging you out of it?

I'd also consider putting a few more linebreaks into the ifs/functions in the code. It feels kind of 'neat' to do oneliners, but it's usually much more readable to have stuff visually broken up. Here's how I'd write it:

```js
Score.prototype.scoreSpare = function(roll, current_frame, last_frame) {
  if (typeof last_frame == 'undefined') { return; }

  if (last_frame.isSpare() && current_frame.isNew()) {
    this._scoreboard[this._scoreboard.length - 2] += roll;
  }
};
```

If the body of the if is just `return`, making it a guard clause, you can just about get away with it — and some JS devs won't even like that.

neoeno: Super nice use of private methods and variable names to enhance readability here though. Makes it really easy to see what's going on. Though, perhaps, the `this._scoreboard` line which is a little more obscure.

I'd also look into JS variable name conventions — usually they're written as `currentFrame` rather than `current_frame`. JSHint or similar might be able to remind you of this too.

src/js/model/Game.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+  if (this.getLength() >= 12 && this.lastFrame().isStrike()) {
+    throw Error('Game over!');
+  }
+};
+
+Game.prototype.isOver = function() {
+  try { this.isComplete(); }
+  catch(err) { var errorMessage = err; }
+}
```
neoeno: I tried deleting the body of this method and I think the whole thing still worked — might be worth thinking about.

src/js/model/Game.js
```diff
@@ -0,0 +1,65 @@
...
+  if (this.getLength() >= 10 && !this.extraFrame()) {
+    throw Error('Game over!');
+  }
+  if (this.getLength() >= 11 && this.lastFrame().isSpare()) {
+    throw Error('Game over!');
+  }
+  if (this.getLength() >= 12 && this.lastFrame().isStrike()) {
+    throw Error('Game over!');
+  }
+};
```
neoeno: I'd usually expect `isComplete` to return a boolean, and maybe `assertComplete` or `throwIfIncomplete` to throw exceptions. The `is` prefix is a pretty strong convention in JS much like `method?` is in ruby — meaning they return true or false and nothing else.

src/js/model/Frame.js
```diff
@@ -0,0 +1,47 @@
...
+  return (this.getFrame().length === 2);
+};
+
+Frame.prototype.isStrike = function() {
+  return (this._frame[0] === 10);
+};
+
+Frame.prototype.isSpare = function() {
+  if (!this.isStrike()) {
+    return (this.sumRolls() === 10);
```
neoeno: The brackets here in the `return` are interesting — were you worried about precedence? I.e. were you trying to avoid `(return this.sumRolls()) === 10`? Or something else?

src/index.html
```diff
@@ -0,0 +1,376 @@
...
+        </table>
+      </div>
+
+
+      <footer id="footer" class="footer py-2 px-3">
+      </footer>
+    </div>
+
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/tether/1.4.0/js/tether.min.js"></script>
```
neoeno: Every week I ask Rick this question and every week he refuses to answer! Maybe you can: why does the application need `tether`?

PR #472, opened 2017-02-13
Rick Clark pull request

rkclark:


rkclark:
Self review:
I have learnt a lot, but not produced much. I've covered: node.js, express, selenium webdriver w/jasmine, jasmine-node and mongodb!
Problems were many, including:
- Splitting up the express routers into individual modules
- Getting selenium to work
- Trying to decide where to use Promises
- Running out of time

My index.js router is an example of my router code pre re-factor into use of promises and functions in the rolls model.

The rolls.js router is where I have tried to refactor.

I'd really appreciate some advice on what is good and bad in these two routers, and the rolls.js model.

Sorry its such a mess! I know there is loads of commented out code. I've not removed it as the app is very much still in progress.

neoeno:
Hi Rick!

Wow it seems like you went pretty deep into this one! I'm sure you learned a chunk out of it too. That's good of course, but since this is a code review and not a learning review I may be a little harder on the results. It should be said — the learning is super great on its own merits!

That all said — here's the code review.

* Commented out code! You already know it's there — but it's really considered a bad sign if you have any commented out code checked into source control in any commit, even if you're in the middle of something. If you need to save code — use git to do that for you. Got some code that you need to delete as part of a refactor but you know you'll need to add back in in a few commits time? Don't comment it out — commit the code before the refactor, then delete the code as part of the refactor, then commit, then when you need it again look back at your git history and copy it out of there.
* The use of promises is interesting here! I think I misunderstood you before — you're creating new promises, not using ones that are given to you. In this case, I think you can achieve what you need to without creating your own promises. (At this point I called you over to talk about it!)
* The models look more like the [Data Access Object](https://en.wikipedia.org/wiki/Data_access_object) pattern than classic models — no problem with that as such, just so's you know a little more about them.
* Nice work on the feature tests — handy to have in your toolbelt!
* The routes seem more like controllers to me — maybe they're routes in node parlance. They also seem a little bit bogged down with detail, and I'd consider extracting the logic to other places. You might find it interesting to look at the difference between the code using libraries and the code not using libraries:

```js
    if (err) {
      res.render('index', { title: 'Ten Pin Bowling Scores', rolls: [] });
    } else {
      var rolls = [];
      if (result[0].rolls){
        rolls = result[0].rolls
      }
      res.render('index', { title: 'Ten Pin Bowling Scores', rolls: rolls });
    }
```

I find lines 2 and 8 easy to understand — the methods have names that communicate intent. The rest of the code I have very little ability to understand without following through the life of `rolls` through the application, consulting the database, etc. Try to make your code be as readable, communicate as much intent as the libraries you're using — ideally even moreso.

It's super cool you've taken the time to explore a bit into an area that interests you. I'd suggest taking it even further. Maybe you could have satisfied yourself by doing a very simple bowling implementation, just a few files that implements it with no visual interface or anything — then take as much time as you like exploring node, etc, setting stuff up exactly how you want it. You'll know best, but I suspect you'd have learned even more that way!

I have to remind you though — when it comes to doing code for an employer, or doing a tech test, they won't really accept running out of time because you wanted to explore using a new technology. You'll have to perform well, write code at your best, use the technologies you think appropriate, within the time window. So make sure you're comfortable doing that too! I know I know — lecture over — I wouldn't be doing my job if I didn't mention it though!

PR #476, opened 2017-03-20
Js logic, unfinished

evebalog:
(For Mary to have a look pls, thanks heaps!)


spec/FrameSpec.js
```diff
@@ -0,0 +1,113 @@
...
+  var frame;
+
+  beforeEach(function() {
+    frame = new Frame();
+  });
+
+  describe("Calculates scores", function() {
+    it("can put the score after a roll into the roll array", function() {
+      frame.roll(7);
+      expect(frame._rolls).toEqual([7]);
```
maryrosecook: Try to avoid testing state.  Test behaviour instead.  Is there a public behaviour that you can test that happens because 7 has been rolled?

spec/FrameSpec.js
```diff
@@ -0,0 +1,113 @@
+'use strict';
+
+describe("Frame", function() {
```
maryrosecook: Really nice, clean, easy to read tests!

src/Frame.js
```diff
@@ -0,0 +1,51 @@
...
+  return this._rolls.length === 2 && this.calculateFrameScore() === 10;
+};
+
+Frame.prototype.isComplete = function () {
+  return this._rolls.length === 2 || this.isStrike();
+};
+
+Frame.prototype.bonus = function(value) {
+  if (this.isStrike()) {
+  this._bonus = value.reduce(function(score1, score2) {
```
maryrosecook: Try and keep indentation consistent with JS idioms.

maryrosecook: Could `score1` and `score2` have names that more clearly explain their role in the code?

src/Frame.js
```diff
@@ -0,0 +1,51 @@
+function Frame(finalFrame) {
+  this._rolls = [];
+  this._bonus = null;
```
maryrosecook: Watch out for having state that can have a value that would cause an error.  eg in this case, what happens if someone uses `_bonus` when it's still `null`?

src/Frame.js
```diff
@@ -0,0 +1,51 @@
...
+  } else if (this.isSpare()) {
+  this._bonus = value[0];
+  };
+};
+
+Frame.prototype.calculateFinalScore = function () {
+  return this.calculateFrameScore() + this._bonus;
+};
+
+Frame.prototype.alert = function () {
```
maryrosecook: Where will the `alert()` method be used? Grab me and let's have a chat about it.

src/Frame.js
```diff
@@ -0,0 +1,51 @@
...
+  this._bonus = null;
+  this._finalFrame = finalFrame;
+};
+
+Frame.prototype.roll = function(value) {
+ this._rolls.push(value);
+};
+
+Frame.prototype.calculateFrameScore = function () {
+  return this._rolls.reduce(function(score1, score2) {
```
maryrosecook: Nice reduce!

PR #478, opened 2017-03-20
Week 5 weekend challenge

joemaidman:


src/frame.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+};
+
+Frame.prototype._completeFrame = function () {
+  this.complete = true;
+  this.pins = 0;
+};
+
+Frame.prototype._calculateScore = function () {
+  return this.balls.reduce(function(a, b) {
```
maryrosecook: Nice!

src/frame.js
```diff
@@ -0,0 +1,79 @@
...
+  return score;
+};
+
+Frame.prototype._getPins = function () {
+  return this.pins;
+};
+
+Frame.prototype._checkPins = function(score){
+  this.pins -= score;
+  if(this.balls.length === 1 && this.balls[0] === 10){
```
maryrosecook: Are there ways to increase the readability of expressions like `this.balls.length` and `this.balls[0]`?

src/game.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+Game.prototype.wasPerfect = function (score) {
+  return score === 300 ? true : false;
+};
+
+Game.prototype._getAllBalls = function () {
+  var balls = [];
+  this.frames.forEach(function(frame){
+    frame.balls.forEach(function(ball){
```
maryrosecook: Why might `frame.balls.forEach...` be breaking an OO guideline?

PR #484, opened 2017-03-20
(WIP) Bowling Challenge

ayanit1:


ayanit1:
WIP, I need to change add logic, to show a working scoreboard. I wanted to play around with jQuery so my calculator is able to do a final total score, but not an accumulative score, which is not what I want. Also if I had more time, I'd work more with guard cases so users aren't able to break boundaries.

PR #489, opened 2017-03-26
Bowling Challenge

ledleds:
10th Frame calculation not yet implemented.


src/Frame.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+function Frame(id){
+  this.MAX_SCORE = 10;
+  this.id = id;
+  this.totalFrame = [];
```
maryrosecook: Could totalFrame have a name that expresses its intent more clearly?

src/Frame.js
```diff
@@ -0,0 +1,35 @@
...
+}
+
+Frame.prototype.inputBowls = function(first, second) {
+  var pins = first + second;
+  this.checkNumberOfPins(pins);
+  this.totalFrame.push(first, second);
+};
+
+Frame.prototype.checkNumberOfPins = function(pins) {
```
maryrosecook: Could the name of checkNumberOfPins be even more specific? What is it checking about the number?

src/Frame.js
```diff
@@ -0,0 +1,35 @@
...
+  return this.totalFrame.reduce(function (first, second) {
+    return first;
+  });
+};
+
+Frame.prototype.isStrike = function() {
+  return this._firstRoll() === this.MAX_SCORE;
+};
+
+Frame.prototype.isSpare = function() {
```
maryrosecook: Love these helper methods!

src/Frame.js
```diff
@@ -0,0 +1,35 @@
...
+  this.checkNumberOfPins(pins);
+  this.totalFrame.push(first, second);
+};
+
+Frame.prototype.checkNumberOfPins = function(pins) {
+  if (pins > this.MAX_SCORE) { throw new Error ("Score cannot be over 10"); }
+};
+
+Frame.prototype._firstRoll = function() {
+  return this.totalFrame.reduce(function (first, second) {
```
maryrosecook: Could this be done without using a reduce?

src/Frame.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+function Frame(id){
+  this.MAX_SCORE = 10;
+  this.id = id;
```
maryrosecook: Is id necessary?

src/Game.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+function Game(){
+  this.MAX_FRAMES = 10;
+  this.frames = [];
+  this.newFrame = new Frame(1);
+  this.totalscore = 0;
+}
+
+Game.prototype._canBowl = function() {
```
maryrosecook: This method throws an exception if there's a problem, but also returns true if there's no problem.  This feels like two responsibilities.  Can we use the lack of an exception as a sign that all is well, instead?

src/Game.js
```diff
@@ -0,0 +1,37 @@
...
+  if (this._canBowl()) {
+    this.newFrame.inputBowls(first, second);
+    this.frames.push(this.newFrame);
+    this.newFrame = new Frame(this.frames.length + 1); }
+};
+
+Game.prototype.calculateScore = function() {
+  return this.frames.reduce(function(memo, frame, index, arr) {
+    if (arr[index].isStrike()) {
+      return (memo + frame.totalFrameScore(arr[index + 1])) + arr[index + 1].totalFrameScore();
```
maryrosecook: Can the logic for calculating a single frame be moved to the frame? Possibly by passing in the other frames that are required to calculate that frame?

src/Game.js
```diff
@@ -0,0 +1,37 @@
...
+
+Game.prototype.addNewFrame = function(first, second) {
+  if (this._canBowl()) {
+    this.newFrame.inputBowls(first, second);
+    this.frames.push(this.newFrame);
+    this.newFrame = new Frame(this.frames.length + 1); }
+};
+
+Game.prototype.calculateScore = function() {
+  return this.frames.reduce(function(memo, frame, index, arr) {
```
maryrosecook: Really love this reduce!

PR #501, opened 2017-03-27
Makers Weekend Challenge Pull Request: CCB

ClemCB:


src/Score.js
```diff
@@ -0,0 +1,31 @@
...
+    this.runningScore += ((2 * frameScore[0]) + frameScore[1]);
+  } else {
+    this._checkForStrike(frameScore);
+    this._checkForSpare(frameScore);
+    this.runningScore += (frameScore[0] + frameScore[1]);
+    return frameScore[0] + frameScore[1];
+  }
+};
+
+Score.prototype._checkForStrike = function (frameScore) {
```
maryrosecook: Could the `Score` object provide a method that returns whether it represents a strike, rather than calculating that in advance and then exposing the state?

PR #502, opened 2017-03-27
Bowling challenge (WIP)

azntastic:


src/frame.js
```diff
@@ -0,0 +1,14 @@
+//each frame initializes with empty array
+
+function Frame(){
+  this.outcome = []
```
maryrosecook: The Bowl abstraction feels really useful.  However, it feels like Frame objects break the encapsulation of Bowl objects by copying each roll from bowl into outcome.

src/frame.js
```diff
@@ -0,0 +1,14 @@
...
+}
+
+Frame.prototype.start = function(){
+  bowl = new Bowl()
```
maryrosecook: Should the frame do the bowling (by calling bowl.roll, or should a roll be created with the rolls and then passed into a new frame?

src/game.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+Game.prototype.play = function(){
+  while (this.frames.length < 10){
+    this.advance()
+  }
+}
+
+// Probably best to remove scoring to another class; implement strikes etc
```
maryrosecook: Sounds like a promising idea!

src/scoring.js
```diff
@@ -0,0 +1,19 @@
+function scoring() {
+  this.score = 0
+}
+
+scoring.prototype.outcome = function(outcome){
+  return outcome
+}
+
+scoring.prototype._isstrike = function(x, y){
```
maryrosecook: Could _isstrike take a frame, rather than the individual rolls (to preserve encapsulation).

src/game.js
```diff
@@ -0,0 +1,40 @@
+// could build a scoring class but call those methods within this/frame function
+function Game(){
+  this.frames = []
+  // scoring = new scoring()
+}
+
+Game.prototype.advance = function(){
```
maryrosecook: Could advance be named to more clearly express its intent?

src/frame.js
```diff
@@ -0,0 +1,14 @@
+//each frame initializes with empty array
+
+function Frame(){
+  this.outcome = []
+}
+
+Frame.prototype.start = function(){
```
maryrosecook: Could start be named to more clearly express its intent?

PR #577, opened 2017-07-09
Bowling-challenge

mihobo:
Bowling-challenge


lib/jasmine-2.6.4/console.js
```diff
@@ -0,0 +1,190 @@
+/*
```
maryrosecook: Could the Jasmine library have been committed separately from the code that used it? What advantages would this have brought to the readability/changeability of the commit history?

spec/gameSpec.js
```diff
@@ -0,0 +1,85 @@
+describe("Game", function(){
+
+  var game;
+
+  beforeEach(function(){
+    game = new Game();
+  });
+
+  it("has no intial rolls", function(){
+    expect(game.rolls).toEqual([]);
```
maryrosecook: It feels like this is testing state, rather than behaviour.  Why is this not ideal? Are there other places that are testing behaviour instead of state?

spec/gameSpec.js
```diff
@@ -0,0 +1,85 @@
...
+    it("calculates a spare", function(){
+      game.roll(9);
+      game.roll(1);
+      game.roll(3);
+      expect(game.score).toEqual(16);
+    });
+  });
+
+  describe("#finalFrame", function(){
+    it("lets you roll 3 times in the 10th frame if you get strikes or spares", function(){
```
maryrosecook: Love this clear test description!

src/game.js
```diff
@@ -0,0 +1,60 @@
...
+  this.rolls = [];
+  this.score = 0;
+  this.frame = 1;
+  this.frameRollCount = 1;
+}
+
+Game.prototype.roll = function(roll) {
+  if (this.frame >= 10) {
+    this.finalFrame(roll);
+    } else {
```
maryrosecook: Watch out for this indentation.

src/game.js
```diff
@@ -0,0 +1,60 @@
+function Game() {
+  this.rolls = [];
+  this.score = 0;
+  this.frame = 1;
```
maryrosecook: At the moment, the code incrementally calculates the `frame` and `score` as rolls come in. Why might it be better for the `score` to be calculated whenever it's asked for (by making `score` a method)?

src/game.js
```diff
@@ -0,0 +1,60 @@
+function Game() {
+  this.rolls = [];
```
maryrosecook: Is it dangerous for the code to signal that `rolls` is public information?

src/game.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+Game.prototype.roll = function(roll) {
+  if (this.frame >= 10) {
+    this.finalFrame(roll);
+    } else {
+    if (roll === 10){
+      this.rolls.push("X");
+      this.frame++;
+    } else if ((roll + this.rolls[this.rolls.length -1] === 10) && (this.frameRollCount == 2)) {
```
maryrosecook: Could this code be made clearer by writing helper functions that would describe the intention of conditional statements like `roll + this.rolls[this.rolls.length -1] === 10`?

src/game.js
```diff
@@ -0,0 +1,60 @@
...
+  this.frame = 1;
+  this.frameRollCount = 1;
+}
+
+Game.prototype.roll = function(roll) {
+  if (this.frame >= 10) {
+    this.finalFrame(roll);
+    } else {
+    if (roll === 10){
+      this.rolls.push("X");
```
maryrosecook: Does the string `X` reveal intention? Is there a way to make it clearer?

src/game.js
```diff
@@ -0,0 +1,60 @@
...
+      this.frame++;
+    } else if ((roll + this.rolls[this.rolls.length -1] === 10) && (this.frameRollCount == 2)) {
+      this.rolls.push("/");
+      this.frame++;
+      this.frameRollCount = 1;
+    } else {
+      this.rolls.push(roll);
+      this.frameRollCount === 1 ? this.frameRollCount = 2 : (this.frame++, this.frameRollCount = 1);
```
maryrosecook: This line feels a bit tricky to read.

src/game.js
```diff
@@ -0,0 +1,60 @@
...
+    };
+    this.calculateScore();
```
maryrosecook: It feels a little hard to read that `calculateScore` is called from different places, depending on whether it's the final frame.  Is there a way to DRY up this call to clarify the code?

src/game.js
```diff
@@ -0,0 +1,60 @@
+function Game() {
+  this.rolls = [];
+  this.score = 0;
+  this.frame = 1;
+  this.frameRollCount = 1;
+}
+
+Game.prototype.roll = function(roll) {
```
maryrosecook: There's a lot of mentions of "frame", both in the code and in the variable names.  Does this tell give us a hint about whether the code is adhering to the SRP?

PR #585, opened 2017-08-07
bowling-challenge

oleglukyanov:


dearshrewdwit:
I'd squash the commits that are the same, and note what the changes are:

Quick one: your first commit hash is 62532ef

`> git rebase -i 62532ef`

And then use the docs to `s` squash or `f` force squash the commits so 

```
  |   | Update README.md |   |   | f3fe833
  |   | Update README.md |   |   | f8da21e
  |   | Update README.md |   |   | ccd9fba
```
doesn't happen

js/frame.js
```diff
@@ -0,0 +1,76 @@
...
+  this._increaseScore(pins);
+  this.bonusRollsNeeded--;
+};
+
+Frame.prototype.getScore = function() {
+  return this.score;
+};
+
+Frame.prototype.isComplete = function() {
+  return this.completion;
```
dearshrewdwit: Looks like you could remove some lines and call `_completeFrameIfEligible()` here?

js/game.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+function Game(){
+  this.MAX_FRAMES = 10;
+  this.completion = false;
+  this.currentFrame = new Frame();
+  this.frames = [];
+  this.currentFrameNo = 1;
```
dearshrewdwit: feels like you could use `frames.length()` as an alternative here

js/game.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+function Game(){
+  this.MAX_FRAMES = 10;
+  this.completion = false;
```
dearshrewdwit: same thought here about using `_completeTheGameIfRequired()`

js/game.js
```diff
@@ -0,0 +1,112 @@
...
+  this._startNewFrameIfRequired();
+  this._completeTheGameIfRequired();
+};
+
+Game.prototype.getScore = function() {
+  return this.frames.reduce(this._sumFrameScores, 0);
+};
+
+Game.prototype.isComplete = function() {
+  return this.completion;
```
dearshrewdwit: same thought here about not needing this property and using `_completeTheGameIfRequired()`

js/game.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+// Private implementation
+
+Game.prototype._getFrame = function(n) {
+  return this.frames[n-1];
+};
+
+Game.prototype._giveBonusPointsToEligibleFrames = function(pins) {
```
dearshrewdwit: this feels like it should belong in another object to handle this logic

js/game.js
```diff
@@ -0,0 +1,112 @@
...
+
+Game.prototype._maxFramesAchieved = function() {
+  return this.frames.length === this.MAX_FRAMES;
+};
+
+Game.prototype._saveFrame = function(frame) {
+  this.frames.push(frame);
+};
+
+Game.prototype._frameCountUp = function() {
```
dearshrewdwit: we shouldn't need this if we adding to frames array

spec/frameSpec.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    it('Receives two rolls and returns total number of pins knocked', function() {
+      for(var i = 0; i <= 10; i++) {
+        var roll1score = Math.floor(Math.random() * 11);
+        frame.roll(roll1score);
+        var roll2score = Math.floor(Math.random() * (11 - roll1score));
+        frame.roll(roll2score);
+        expect(frame.getScore()).toEqual(roll1score + roll2score);
+        frame = new Frame();
```
dearshrewdwit: why is this needed?

js/frame.js
```diff
@@ -0,0 +1,76 @@
...
+
+Frame.prototype.roll = function(pins) {
+  if(pins + this.getScore() > this.MAX_PINS || pins < 0) { throw('Received wrong number of pins') };
+  this._rollCountUp();
+  this._increaseScore(pins);
+  this._setBonusIfEligible();
+  this._completeFrameIfEligible();
+};
+
+Frame.prototype.needsBonus = function() {
```
dearshrewdwit: This is another indication to me that this should be extracted. If something 'needs' something else that sounds like a relationship between two different things

dearshrewdwit:
I like how easy it is to read your tests!

PR #633, opened 2017-10-30
Alexi's bowling scorecard

alexiscarlier:


spec/gameSpec.js
```diff
@@ -0,0 +1,84 @@
...
+  var testFrameFive;
+  var testFrameSix;
+  var testFrameSeven;
+  var testFrameEight;
+  var testFrameNine;
+  var testFrameTen;
+  var frames;
+//
+  beforeEach(function() {
+    testGame = new game();
```
mdsomerfield: This feels like a lot of test setup. We're duplicating a lot of effort by instantiating individual frames and assigning individual scores. That's usually a sign that a classes dependencies are too tightly coupled.

If we think about this in a TDD fashion, out tests should look a lot simpler. For example, if we want to test some logic to calculate a frame bonus score then the things we should care about are (a) we have a frame that should get a bonus score (the condition) (b) we have a roll which will generate that bonus score (the action) and (c) an expectation of what that bonus score should be (the result). 

To achieve a really clean test, the only dependency that we should need to inject into our game is a frame that requires a bonus score.

spec/gameSpec.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+  it('calculates the post bonus score of a Strike', function() {
+    frames = [testFrameOne, testFrameTwo, testFrameThree, testFrameFour, testFrameFive, testFrameSix, testFrameSeven, testFrameEight, testFrameNine, testFrameTen];
+    expect(testGame.frameScoreWithBonus(frames, 5)).toEqual(11);
+    expect(testGame.frameScoreWithBonus(frames, 9)).toEqual(20);
+  });
+  it('calculates the post bonus score of a Spare', function() {
+    frames = [testFrameOne, testFrameTwo, testFrameThree, testFrameFour, testFrameFive, testFrameSix, testFrameSeven, testFrameEight, testFrameNine, testFrameTen];
+    expect(testGame.frameScoreWithBonus(frames, 3)).toEqual(15);
+    expect(testGame.frameScoreWithBonus(frames, 4)).toEqual(20);
```
mdsomerfield: In general, you only need a single test case to test a single piece of logic.

src/frame.js
```diff
@@ -0,0 +1,50 @@
+function frame() {
+  this._firstScore = null;
```
mdsomerfield: It feels like an array might be a more suitable structure here than separate instance variables.

src/frame.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+frame.prototype.thirdScore = function() {
+  return this._thirdScore;
+};
+
+//
+
+frame.prototype.isAStrike = function() {
+  return (this.firstScore() === BONUS_SCORE);
```
mdsomerfield: This is good encapsulation of game logic.

src/game.js
```diff
@@ -0,0 +1,46 @@
...
+game.prototype.strikeScoring = function(frames, selectedFrameNumber) {
+  finalFrameScore = 0;
+  if(selectedFrameNumber === frames.length - 1) {
+    finalFrameScore = frames[selectedFrameNumber - 1].preBonusScore() + frames[selectedFrameNumber].firstScore() + frames[selectedFrameNumber].secondScore();
+  }
+  else {
+    finalFrameScore = frames[selectedFrameNumber - 1].preBonusScore() + this.frameScoreWithBonus(frames, selectedFrameNumber + 1);
+  }
+  if(finalFrameScore > 30) {
+    finalFrameScore = 30;
```
mdsomerfield: It feels like we are silently swallowing an error case here - the final frame score should never be more than 30 and if it is then we should throw an exception because that indicates a bug somewhere in the game logic.

src/interface.js
```diff
@@ -0,0 +1,288 @@
+$(document).ready(function() {
+
+  var bowlingGame = new game();
+  var frameOne = new frame();
```
mdsomerfield: I think use of an array here would really simplify some of this logic.

PR #640, opened 2017-11-26
Gabriela's bowling challenge submission 

gabrielabud:


neoeno:
Hi Gabriela! Plenty of time to work on this yet, so I'll give you some responses to your questions that I hope will give you something to work on:

1. Good to work on `eslint` — shows attention to quality :). Your problem is related to your configuration file — it is not in the right format. I would suggest using `eslint` on the command line and getting it working there before moving onto text editor integration. That should make it easier to locate your mistakes.
2. Your mocking of `frame` isn't the issue causing your test to fail. I'd examine it closer — use `console.log`, think carefully about the code, use `console.log` again — etc.
3. Currently `Game` contains all the logic in the program, while `Frame` only contains a bit of state. This would not be an example of the single responsibility principle, as `Game` is responsible for almost everything. Can you think of some ways of extracting that logic? Try putting 'meaning' to the side and trying to think of any possible way of breaking the responsibilities up, come up with a few options, see how they feel.

Keep going. You're putting in some really hard work here and persevering with these questions will really pay off, even if it feels tough. Come find me if you start to get really frustrated :)

gabrielabud:
Added interface!

index.html
```diff
@@ -0,0 +1,82 @@
...
+      <h1>Bowling Scores</h1>
+      <div id="final" class="final"></div>
+    </header>
+    <div class="scores">
+      <div class="frame">
+        <div class="number">1</div>
+        <input type="text" class="first" />
+        <input type="text" value="" maxlength="2" class="second" />
+        <div class="total"></div>
+      </div>
```
dearshrewdwit: I love the consistency of the HTML: the repeated structure and the minimum number of names.

src/Game.js
```diff
@@ -0,0 +1,103 @@
+'use strict'
+
+function Game(){
+  this._frames = [];
+  this._rolls = [];
+};
+
+var game;
```
dearshrewdwit: What's going on here?

What's your intention behind lines 8-9?

src/Frame.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+
+function Frame(nr, rolls, score) {
```
dearshrewdwit: A couple questions:

What behaviour does this object have?

What should it be responsible for?

src/Game.js
```diff
@@ -0,0 +1,103 @@
...
+  if ( ((game._frames).length === 10) && (game.score() === 0)) {
+    return true;
+  }
+  else {
+    return false;
+  }
+};
+
+Game.prototype.isPerfectGame = function() {
+  if ( ((game._frames).length === 10) && (game.score() === 300)) {
```
dearshrewdwit: you don't need to explicitly return true or false, because the comparison itself returns a boolean.

it's enough to do `return game._frames).length === 10) && (game.score() === 300`

src/Game.js
```diff
@@ -0,0 +1,103 @@
...
+game = new Game();
+
+Game.prototype.updateRolls = function (rollsList) {
+  if(rollsList.length > 21){
+    throw new Error('cannot have more than 21 rolls in a game')
+  }
+  game._rolls = rollsList;
+};
+
+Game.prototype.score = function () {
```
dearshrewdwit: I'd like to see this function broken down into its parts, and wrapped in functions to be called (and unit- tested)

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
+  jasmine.addMatchers({
+    toBePlaying: function () {
```
dearshrewdwit: nice!

spec/GameSpec.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(game.isPerfectGame()).toBeTruthy();
+  });
+
+  it('calculates correctly the score', function() {
+    game.updateRolls([3,0,10,10,7,3,5,3,10,8,0,6,0,3,7,10,10,10]);
+    game.score();
+    expect(game.score()).toEqual(155);
+  });
+
+  // not passing, i think is related to doubling frame;
```
dearshrewdwit: Let's avoid commented-out code in submissions

spec/GameSpec.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+describe('Game', function(){
+  var game;
+  var frame;
```
dearshrewdwit: How do the Game and Frame related?

PR #643, opened 2017-11-26
Submission

Eustaquio122:


src/validator.js
```diff
@@ -0,0 +1,76 @@
...
+                this._buildFrame(firstFrame, firstRolls(rolls, 3));
+            } else if (this._isSpare(rolls)) {
+                firstFrame = firstRolls(rolls, 2);
+                this._buildFrame(firstFrame, firstRolls(rolls, 3));
+            } else {
+                firstFrame = firstRolls(rolls, 2);
+                this._buildFrame(firstFrame, firstRolls(rolls, 2));
+            }
+
+            return this._validateFrame(firstFrame) && this._validateFrames(removeRolls(rolls, firstFrame.length));
```
sjmog: In response to your question – "is there a way to make this clearer?", I think this is pretty clear. The recursion is neat, and I don't have a big problem with the `if/else` statement – each one is clearly handling a different kind of roll. It's pretty easy to read and understand.

Some quick wins – there are some magic numbers in here, and your code will be cleaner if you extract them. What is `1`, `3`, `2`? This might lead, in turn, to different versions of the `firstRolls` method.

You might also gain by thinking in a slightly more OO way. For instance – you have some ways of representing a frame that also knows about other rolls (`this._buildFrame(firstFrame, firstRolls(rolls, 3))`). Could these be special kinds of object? For instance:

```javascript
if (this._isStrike(rolls)) {
  this._buildStrikeFrame(rolls)
}
```

Then, in `buildStrikeFrame`:

```javascript
_buildStrikeFrame: function(rolls) {
  this.frameBuilder.buildStrikeFrame(rolls, this._numberOfFrames++)
}
```

And in `buildStrikeFrame`:

```javascript
buildStrikeFrame: function(allRolls, frameNumber) {
  var firstFrame = firstRolls(rolls, 1)
  var otherAffectedFrames = firstRolls(rolls, 3)
  return new Frame(firstFrame, otherAffectedFrames)
}
```

One rule-of-thumb I'm using here is 'if you are pulling information out of a data structure and passing that to an object, just pass the whole data structure to the object instead'. So just pass `rolls` to the `FrameBuilder` and let it figure out how to deal with them.

src/validator.js
```diff
@@ -0,0 +1,76 @@
...
+        },
+
+        _buildFrame: function(frameRolls, scoringRolls) {
+            this._numberOfFrames++;
+            this.frameBuilder._buildFrame(frameRolls, scoringRolls, this._numberOfFrames);
+            return;
+        },
+
+        _isStrike: function(rolls) {
+            return rolls[0] === 10;
```
sjmog: This is a good opportunity to introduce a constant (`STRIKE_SCORE` or something), to avoid the magic number '10'.

src/interface.js
```diff
@@ -0,0 +1,160 @@
...
+        if (roll > 19) {
+            updateLastSelectors();
+        } else {
+            updateRollSelectors(rollScore(roll));
+        }
+    });
+
+// Helpers -------------------------------------------- Helpers //
+
+    function createSelector(id_number, limit) {
```
sjmog: jQuery is generally going to end up like this – even good jQuery. That is: a nice-ish interface and a whole bunch of helpers.

It can be refactored by using one of the framework architectures you know (i.e. MVC) and building a controller that uses jQuery methods, but it'll probably end up much the same.

An alternative thing to do is create a 'view model' – that is, Objects that use jQuery in their methods. You identify objects in the view just like you do in the rest of your program: what nouns are you using? I see: `SpiritBowler`, `Total`, `FrameScore`. If you move the logic for rendering and updating each of these 'view objects' into their own object, you start moving towards a _component-based_ architecture.

sjmog:
Oh, and one general point – watch out for sticking closely to requirements! It's awesome to experiment now and flex your creative muscles, but as a tech test submission this runs a risk of failing because it totally over-delivers on the requirements. Excess delivery = excess time paying devs, from a company point-of-view!

(But I like it ;) )

PR #647, opened 2017-11-27
Initial submission for bowling challenge

antoniobelmar:


dearshrewdwit:
Love the commit messages!

app/public/src/Frame.js
```diff
@@ -0,0 +1,32 @@
+function Frame(roll1, roll2) {
+  this._rolls = [roll1, roll2]
+  this._MAX_PINS = 10
+};
+
+Frame.prototype.getRolls = function() {
+  return this._rolls;
+};
```
dearshrewdwit: In line 2, you declare `_rolls` as a private array. (a piece of state internal to the class). All your `Frame` methods control access to this array, which is great.

EXCEPT this one. This one just says "you know that private array I made earlier? It's not actually private. Any other class can use it. `Game` can use it. `Interface` can use it. Whatever"

This is super-confusing for me as a dev tasked with extending the code. On one hand, I understand that `Frame` should do all the business logic relating to managing an array of rolls. On the other hand, I'm being told from lines 6 to 8 that _any object_ can do the logic relating to managing the array of rolls.

The best bit is I know why you did this. You did it to make `Frame` testable (it's the only part of your code that uses this method). You wanted to test that frames contain rolls. Here's the secret – you don't need to test this. **Test behaviour, not state.** If the `Frame` object functions like it should, I don't care whether it contains rolls as an array, as a string, as an object, or as an AJAX call to some remote server sending out rolls.

app/public/src/Roll.js
```diff
@@ -0,0 +1,10 @@
+function Roll(pins) {
+  if(pins>10 || pins<0 || isNaN(pins)) {
```
dearshrewdwit: So two different perspectives here:

PERSPECTIVE 1:
The point of an initialiser is to set up a valid object. The object has been created and state is then set. I tend to have clean initialisers that expect valid input. I want input validation to happen as soon as possible to the user giving input rather than all the way down the chain to when the roll object is created. At the moment this means that if I were to change the rules I have to change the if condition of this object, which 

This also means a bunch of errors will be thrown in my program which I need to handle in order for my program not to break every single time bad input is given. Ideally, in production, i'd like to validate the input on the client side first, and then have a validation layer before calling new Roll().

PERSPECTIVE 2:

The responsibility of this object is to validate input for a roll, so it's incorrectly named. It should be some kind of RollBuilder or some such. Throwing errors is ok because it's a small specific error case that won't bubble up if handled at the correct layer.

app/public/src/Frame.js
antoniobelmar: Hi Ed,
I think I get this but I am just trying to see what you mean by "use it", other classes can certainly see the array but not really use it right? I understand that they don't really have a need for it(as you correctly point out I don't use get Rolls in other classes...that said, I do use it in the following prototype method of a Frame:

Frame.prototype.getRollKnockedPins = function(index) {
  return this.getRolls()[index].getKnockedPins();
};

If I shouldn't be making a getter for this should I be accessing my rolls property directly then?

Thanks!

app/public/src/Roll.js
antoniobelmar: Understood perspective 1, I guess the problem is me building this before building the interface which means that I didn't really take into account I could have validations on the actual user interface.

Understand perspective 2, would I be able to just get away with one object then? I seem to keep having issues with naming my objects correctly (Round class in rock/paper/scissors comes to mind), any tips for that? I know it seems stupid but when I realize a pattern I get worried!

PR #653, opened 2017-12-03
Bowling challenge progress

Yolantele:


spec/BowlingSpec.js
```diff
@@ -0,0 +1,74 @@
+describe('Bowling', function () {
+
+  var myGlobal
+  var bowling
+  var f
+  var f2
```
mdsomerfield: It feels like you don't need to declare all these frames here - they are only used in a single scope so why not make them local variables? (e.g. like on line 28)

spec/BowlingSpec.js
```diff
@@ -0,0 +1,74 @@
...
+
+    bowling = new Bowling()
+  })
+
+  it('should add to the list of frames', function () {
+    bowling.addFrame(f)
+    expect(bowling.frames).toContain(f)
+  })
+  it('should only add max 10 frames', function () {
+    var f = new Frame ('frame2')
```
mdsomerfield: This is a good test, but could it be more general? i.e. instead of testing that we can't add more than 10 frames, could we simply test that we couldn't add more than X frames, where X could be 5, 10, 20, 100? 

Obviously this would require a change in the code to allow the max number of frames to be configurable - but whenever we can move from a specific to a general case our code becomes more resilient to change.

spec/BowlingSpec.js
```diff
@@ -0,0 +1,74 @@
...
+  describe('#strikeBonus', function () {
+    it('counts strikeBonus', function () {
+      f5 = new Frame (10, 0)
+      f6 = new Frame (1, 0)
+      f7 = new Frame (1, 0)
+      expect(bowling.strikeBonus()).toEqual(2)
+    })
+    it('unless player rolls another strike', function () {
+      f5 = new Frame (5, 4)
+      f6 = new Frame (10, 0)
```
mdsomerfield: I'm a bit confused in these tests - you create frames f5, f6, f7 etc, but you don't ever call bowling.addFrame like you do in the earlier tests. Do these tests pass?

spec/frameSpec.js
```diff
@@ -0,0 +1,19 @@
+describe('Frame', function () {
+  var frame
```
mdsomerfield: I feel like theses tests are all state rather than behaviour.

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
...
+  },
+  rollScore: function () {
+    return this.first + this.second
+  }
+}
+
+function Bowling () {
+  this.frames = []
+  this.totalPoints = 0
+  this.MAX_FRAMES = 10
```
mdsomerfield: I mentioned in a comment on your tests that we could go from specific to general - to highlight what I mean by that, consider whether you could inject a variable maxFrames here.

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
+'use srict'
+
+// var Frame = require('../src/frame')
+function Frame (first, second) {
```
mdsomerfield: Usually a good idea to keep one constructor function/prototype definition to a single file.

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
...
+  this.rolls = [first, second]
```
mdsomerfield: Remember when using prototype it's a good idea to use the _ convention to indicate intended encapsulation

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
...
+Bowling.prototype = {
+
+  error: function () {
+    if (this.frames.length > this.MAX_FRAMES) {
+      throw ('Max Frames Added')
+    }
+  },
+
+  addFrame: function (generatedFrame) {
+    this.error()
```
mdsomerfield: this.validateFrames() or something similar would be a better name here - this.error() feels like you are creating an exception not validating the number of frames.

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
...
+    this.frames.forEach(function (frame) {
+      var total = 0
+      console.log(frame)
+      for (var i in frame) { total += frame[i] }
+      console.log(total)
+      this.totalPoints += total
+    })
+  },
+  spareBonus: function (frame) {
+    return 4
```
mdsomerfield: Why 4?

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
...
+    var i
+    this.frame[i] = frame
+    return this.frame[i + 2] // check if it returns the next to following frame
+  },
+  _isStrikeInARow: function () {
+    return false
+  },
+  strikeBonus: function () {
+    if (this._isStrikeInARow === false) {
+      return 2 // return normal strike bonus
```
mdsomerfield: Why 2?

src/Bowling.js
```diff
@@ -0,0 +1,92 @@
...
+      console.log(total)
+      this.totalPoints += total
+    })
+  },
+  spareBonus: function (frame) {
+    return 4
+  },
+  _followingFrame: function (frame) {
+    var i                         // check if it returns the next frame
+    this.frames[i] = frame
```
mdsomerfield: Is 'i' not undefined at this point?

PR #656, opened 2017-12-03
Robert Clayton Bowling Challenge

RobertClayton:


public/views/jQuery.js
```diff
@@ -0,0 +1,141 @@
+$(document).ready(function() {
+
+  scorecard = new BowlingScorecard();
+
+  var outputCounter = 0;
+  var numberChecker = 0;
+  printTotal();
+
+  $("#button0").on('click', function() {
```
mdsomerfield: I think there are several improvements we can make here. 

First, rather than assigning an anonymous function as a click handler, it's usually better to use a named function, e.g. 

```
function handleButtonClick() {
   ..code...
}

$(document).ready(function() {
   $("#button0").on('click', handleButtonClick);
});
``` 

The advantages of this are that the handleButtonClick function is now testable independently of the JQuery that it is handling - we've separated the concern of reacting to an event from how we react to that event.

Second, there's a lot of repetition here. Rather than assigning the click handlers individually, we could generalise, e.g.

``` 
   $(document).ready(function() {
      var scoreButtons = $('.scoreButton');
      for (var index = 0; index < scoreButtons.length; index++ {
         scoreButtons[0].on('click', handleButtonClick);
      }
   });
``` 

But how do we assign different values to each of the buttons in this case? Well, there's nothing to stop us passing information into handleButtonClick:

``` 
   function handleButtonClick(buttonElement, rollValue) {
      $(buttonElement).html(rollValue);
      printButton(rollValue);
      etc...
   }

   $(document).ready(function() {
      var scoreButtons = $('.scoreButton');
      for (var index = 0; index < scoreButtons.length; index++ {
         scoreButtons[0].on('click', function() {
            return handleButtonClick(scoreButtons[0], scorecard.roll(index));
         });
      }
   });

```

public/views/jQuery.js
```diff
@@ -0,0 +1,141 @@
...
+  $("#button10").on('click', function() {
+    if (isGameOver(outputCounter) == false) {
+      $("#button10").html(scorecard.roll(10));
+      printButton(10);
+      printTotal();
+    }
+  });
+
+  function printButton(number) {
+    outputCounter += 1;
```
mdsomerfield: I'm not sure I totally understand what this is doing, but I think this is trying to work out which frame to display the score on. If that's the case, a better separation of concerns would be:

```
var currentFrame = scorecard.getCurrentFrame();
$('#' + currentFrame).text(number);
```

i.e. your view logic doesn't need to keep track any outputCounter - it just asks the scorecard 'what's the current frame?', and when the score card tells it, it understand that it needs to apply the last roll to the view element representing that frame.

Event better though, if you follow an OO approach and you treat each of the individual frames as objects, you can get the UI to listen for changes in the frames; 

```
// Define a frame class
function Frame() {
   var score = 0;
   var listener = null;

   function setScore(roll) {
      score = roll;
      if (listener !== null) {
         listener(score);
      }
   }

   function getScore() {
      return score;
   }

   function addListener(scoreListener) {
      listener = scoreListener;
   }
   
   return {
      setScore: setScore,
      addListener: addListener
   }
}

...
// In your view logic
var buttons = $('.buttons);
for (var index = 0; index < buttons.length; index++) {
   var frame = scoreCard.getFrame(index);
   frame.addListener(function(score) {
      updateScore(buttons[index], score);
   });
}
```

This last example is pretty advanced and may be difficult to follow, but it's a good place to end up if you can :)

public/views/jQuery.js
```diff
@@ -0,0 +1,141 @@
...
+function isGameOver(outputCounter) {
+  if (outputCounter > 21) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+function printTotal() {
+  $('#total1').text(scorecard.printTotal[0]);
```
mdsomerfield: We can use a similar iteration to one of the comments above:

```
var totalElements = $('.totals);
for (var index = 0; index < totalElements.length; index++) {
   totalElements[index].text(scorecard.printTotal(index));
}
```

PR #658, opened 2017-12-04
Bowling-Challenge Marco

Vanals:


spec/bowlingSpec.js
```diff
@@ -0,0 +1,171 @@
...
+      bowling.recordInFrame(5);
+      bowling.reducePins(5)
+      bowling.increaseActualFrame();
+      bowling.throw(5);
+      bowling.recordInFrame(5);
+      bowling.reducePins(5)
+      bowling.increaseActualFrame();
+      bowling.throw(5);
+      bowling.recordInFrame(5);
+      bowling.spareBonus();
```
mdsomerfield: If I have to call a lot of methods on my test object in order to achieve the right test setup, I see that as a sign that my code either has a lot of tight coupling or that I haven't effectively modelled the behaviour.

For example, in this case:
- 'throw', 'recordInFrame' and 'reducePins' are all the result of the same action. I would expect to be able to call 'throw' and for 'recordinframe' and 'reducePins' to happen automatically. The client calling bowling shouldn't need to know how to do those things - bowling needs to know that when we throw down some pins, we also need to record the score and reduce the number of pins at the target.
- 'increaseActualFrame' is a a game rule rather than a direct result of the throw, but it's also something that we shouldn't need to know about. The bowling object should know that when we've finished a throw, we move on to the next one. 
- the fact that we need to do all this setup in the first place feels a lot harder than if we just created a bowling object and injected a scorecard which was already initialised with a spare.

If we address all of these problems then our test looks like

```
//Arrange
var frame = new Frame(5, 5);
var scoredCard = new ScoredCard([frame]);
var bowling = new Bowling(scorecard);

// Act
bowling.throw(5);

// Assert
expect(... test condition...)
```

public/src/jQuery.js
```diff
@@ -0,0 +1,32 @@
+$(document).ready(function() {
+
+  bowling = new Bowling();
+
+  $('#throwBall').click(function() {
+    var score = $('#scoreChoosen').val();
+    if ( bowling.points.length <= 19) {
```
mdsomerfield: We've got a lot of logic inside an anonymous function which is tightly coupled to our view. That makes it very difficult to test. A better approach is to name the function e.g.

```
function throwBallClicked() {
   ... code here ...
}

$(document).ready(function() {
  $('#throwBall').click(throwBallClicked);
}
```

You could then further improve it by moving the throwBallClicked function to a module or a class - either way, you are now able to test this functionality without needing to involve JQuery.

public/src/jQuery.js
```diff
@@ -0,0 +1,32 @@
...
+      bowling.throw_record(parseInt(score));
+      writePtAndScore();
+      writeSpareBonus();
+      bowling.increaseActualFrame();
+    } else if (bowling.points.length === 20) {
+      bowling.throw_record(parseInt(score));
```
mdsomerfield: We went through this a bit in the Monday code review, but whenever you find yourself using if/else for game rules or business logic, ask whether you might get better results via Polymorphism or Strategy patterns. See:

https://refactoring.guru/replace-conditional-with-polymorphism

PR #659, opened 2017-12-04
bowling challenge

SuzanneHuldt:
back end implementation only 
all tests passing except for perfect game which returns 10 points too many


SuzanneHuldt:
all tests now passing

jasmine/spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
+  jasmine.addMatchers({
```
mdsomerfield: I wasn't aware you could do this with Jasmine, but I like it - nice!

lib/score_card.js
```diff
@@ -0,0 +1,119 @@
...
+  this.frame = new Frame()
+  this.frames = []
+  this.frameCount = 1
+}
+
+ScoreCard.prototype.recordRoll = function(points){
+  this.frameCheck()
+  if(this.cheatGuard(points)==true){
+    return 'illegal roll'
+  } else {
```
mdsomerfield: Two things - firstly, we don't really need the 'else' and secondly we should encapsulate the validation logic within the method. I would expect to see something like:

```
if (this.isLegalRoll(points)) {
  return 'Illegal Roll';
}

... code here ... 
``` 
Or even better

```
assertIsLegalRoll(points);

... code here ... 
``` 

Where assertLegalRoll throws an exception if it is not legal, and the calling code is responsible for deciding what to do with that exception (because asserting a rule and handling the result of that assertion are two different responsibilities)

lib/score_card.js
```diff
@@ -0,0 +1,119 @@
...
+    if(this.frameCount > 10){
+      this.addBonus(points)
+    } else {
+      this.addPoints(points)
+}
+}
+}
+
+
+ScoreCard.prototype.addBonus = function(points){
```
mdsomerfield: It feels like there's a lot of if/else here, in the next method and elsewhere in the code. A lot of what you're doing in the scorecard is manipulating the frame based on different game states. In Ruby, you may have come across the principle 'Tell, don't ask' (see https://robots.thoughtbot.com/tell-dont-ask) - the same principle applies. 

Rather than 'If we have a spare bonus, then, if we have a strike bonus and it's the last frame, then' etc, you should be aiming for 'Hey frame, here's the score - deal with it'. 

You can achieve that by better encapsulating this logic within the frame - so you end up just calling frame.roll(numberOfPinsKnockedDown) and letting frame dealing with the rest.

Of course, you still need to deal with bonuses, and one frame doesn't need to know that another frame is waiting for a bonus. This is where we need to create game rules (as we discussed in code review on Monday), we we could have a StrikeBonusRule and a SpareBonusRule and our code becomes:

```
frame.roll(numberOfPinsToKnockDown)
rules.forEach((rule) => {
  rules.apply(frames, numberOfPinsKnockedDown)
});
```
And an example rule might be
```
function StrikeBonusRule() {
   function apply(frames, numberOfPinsKnockedDown) {
      var framesAwaitingBonus = frames.filter((frame) => {
         return frame.needsStrikeBonus();
      })
      framesAwaitingBonus.addStrikeBonus(numberOfPinsKnockedDown);
   }
   return {
      apply: apply
   }
}
```

spec/score_card_spec.js
```diff
@@ -0,0 +1,169 @@
...
+    scorecard.recordRoll(10)
+    console.log(scorecard.bonusFrame)
+    scorecard.recordRoll(10)
+    console.log(scorecard.bonusFrame)
+    scorecard.recordRoll(10)
+    console.log(scorecard.bonusFrame)
+    console.log(scorecard.total)
+    console.log(scorecard.frame.points)
+    scorecard.recordRoll(10)
+    console.log(scorecard.bonusFrame)
```
mdsomerfield: Any time we need to call more than a couple of methods for test setup, we should be asking ourselves whether we've effectively separated our concerns. Usually a long test setup is a sign of (a) tightly coupled classes, (b) lack of dependency injection or (c) testing the wrong thing. 

Here, I think there's a mix of all three. Do we need to test that it returns 300 for a perfect game? Or is it enough to test that strike bonus gets applied correctly to the first frame, the 9th frame and the 10th frame (i.e. the three different cases). I would argue that the second option is all that is required, and is in fact better because if the tests fail it's immediately clear where the hole in the logic is. 

But to achieve that test, the way your code is currently set up, you'd need to play 8 frames to set up the 9th frame test and 9 frames to reach the 10th frame test. We should be able to inject state into our score card such that our test is something like:

```
var numberOfFramesInGame = 3;
var frames = [new Frame(10)];
var currentFrame = 1;
var scoredCard = new Scorecard(numberOfFramesInGame, currentFrame, frames);
```

With this test setup, we can immediately test the condition that the penultimate frame handles the strike bonus correctly, and we could do the same for the first and last frames in a game.

(Also, try not to check in console log statements)

PR #663, opened 2017-12-31
Bowling challenge(Xin Wang)

Xin00163:


index.html
```diff
@@ -0,0 +1,89 @@
...
+  <script src='src/jquery-3.2.1.min.js'></script>
+  <script src='src/game.js'></script>
+  <script src='src/interface.js'></script>
+</head>
+
+<body>
+  <div id='title' class='title'><h1>Bowling!</h1></div>
+  <div id='buttons' class='buttons'>
+    <p>Enter on your roll</p>
+    <button class="rollScore"> 0 </button>
```
mdsomerfield: You mentioned on your review request form that you struggled with the UI. I think we covered some of this in the code review on Monday morning, but I'll cover here again for reference. 

Here, rather than 10 buttons, we could use an iterator. If you have an element:

```
<div id='button-container'></div>
```

Then you could add your buttons using a for loop:

```
function buildButtons() {
    for (var index = 0; index <= 10; index++) {
        var button = $('<button class="rollScore"></button>');
        button.text(index);
        button.click(function() {
            setScore(index);
        });
        $('#button-container').append(button);
    }
}

function setScore(score) {
   .. set the score
}
```

index.html
```diff
@@ -0,0 +1,89 @@
...
+    <button class="rollScore"> 7 </button>
+    <button class="rollScore"> 8 </button>
+    <button class="rollScore"> 9 </button>
+    <button class="rollScore"> 10 </button>
+    <br>
+  </div><br>
+
+  <div id='scoresheet'>
+    <table id='scoresheetTable' class='scoresheet' cellpadding='1' cellspacing='0'>
+      <tr>
```
mdsomerfield: An alternative to treating this as a table is to use a series of divs. If you refactor you game code to use frame classes such that you have an array of frames, you could write a method here like:

```
function buildGameCard() {
    $container = $('#score-card-container');
    game.getFrames().forEach((frame) => {
        var frameView = buildFrameView(frame);
        $container.append(frameView);
    });
}

function buildFrameView(frame) {
    var frameView = $('
        <div class='frame'>
        <span class='roll1'></span><span class='roll2'>
        <span class='score'></score>
        </div>');
    frameView.find('.roll1').text(frames.getFirstRoll());
    frameView.find('.roll2').text(frames.getSecondRoll());
    frameView.find('.score').text(frames.getScore());
    return frameView;
}
```

PR #679, opened 2018-02-04
Complete to starting to check for bonus points

AttyC:
My work, not helped by others.
Currently passing two values to Roll function - this will be refactored to pass in values one by one, once the scoring / bonus mechanism is finished.


jasmine/spec/GameSpec.js
```diff
@@ -0,0 +1,78 @@
...
+    frameScore = [];
+
+  });
+
+  describe('beginning of Game', function(){
+    it('starts a new game', function(){
+        expect(game).toEqual(game)
+    });
+    it('is empty before game starts', function(){
+        expect(game.frames).toEqual([]);
```
mdsomerfield: This feels like testing state not behaviour. 

If we were to test various cases such as rolling a 7, rolling a 7 then a 3, rolling a 7 then a 3 and then a 5 (for example) then we can implicitly test this - because in order for the outcome of those tests to be correct then our start state must be as we expect it. 

Arguably, even if it isn't, this isn't something we need to test because that kind of state should be encapsulated - we should only be concerned with outcomes, not how those outcomes are achieved.

jasmine/src/roll.js
```diff
@@ -0,0 +1,19 @@
...
+
+Roll.prototype.roll = function(score1, score2) {
+  if (score1 == 10) { this.isStrike = true; } //strike
+  else if (score2 + score1 == 10){ this.isSpare = true; } // spare
+  this.addScore(score1, score2, this.isStrike, this.isSpare);
+};
+
+Roll.prototype.addScore = function (score1, score2, strikeStatus, spareStatus) {
+  this.scoreSoFar += (score1 + score2); // total for frame (before any bonuses added)
+  this.frameScore = {'Ball 1: ' : score1, 'Ball 2:' : score2, 'Strike: ' : strikeStatus, 'Spare: ' : spareStatus }; // push values to Frame Score array
```
mdsomerfield: Whenever you find yourself using a hash in a case where you are likely to have more than one instance and those instances will have a consistent data structure, think about using a class instead. 

Having a class usually (a) makes your code easier to follow, but also (b) gives you a mechanism for encapsulation. For example, you may decide that it is the responsible of a frame class (rather than a roll) to know whether it is a strike or a spare, so you could have a method on the frame:

```
Frame.prototype.isStrike = function() {
   return this.score1 === 10
}
Frame.prototype.isSpare = function() {
   return !this.isStrike() && (this.score1 + this.score2 === 10)
}
```

This refactor isn't necessary, this is just for illustration - injecting the strike and spare values from a roll class is also a perfectly reasonable approach. However, if you were to start integrating this with a UI, I suspect that you would start seeing the value of being able to attach logic to an instance of the frame class rather than just having an array of hashes.

jasmine/src/roll.js
```diff
@@ -0,0 +1,19 @@
+function Roll(game, frame){
+  this.game = game;
+  this.frameScore = null;
```
mdsomerfield: By convention, these values should have underscores and we should try not to access them from outside this class, i.e. 

```
this._isStrike
this._isSpare
```

Also, do isStrike and isSpare need to be class variables? You only use them in the roll function, so I would suggest that you could get rid of them and your roll function could be:

```
Roll.prototype.roll = function(score1, score2) {
  var isStrike = (score1 == 10);
  var isSpare = !isStrike && (score2 + score1 == 10)
  this.addScore(score1, score2, isStrike, isSpare);
};
```

PR #680, opened 2018-02-04
Bowling-Challenge - week 6 - UI & game logic 

ghost:


spec/FrameSpec.js
```diff
@@ -0,0 +1,299 @@
...
+      frame = new Frame();
+    });
+
+    it("should be 10 before play starts", function() {
+      expect(frame.pinsRemaining()).toEqual(10);
+    });
+
+    it("should be 5 if 5 already down", function() {
+      frame.play(5);
+      expect(frame.pinsRemaining()).toEqual(5);
```
mdsomerfield: Hmm. So I have written and rewritten this comment a couple of times. I feel like there are a lot of tests here and some of them are duplicating but its hard for me to pin down exactly what I would change. 

I feel like these three tests are a good example - I think there's a single test here (if I play 4 and then 3, there are 3 pins remaining, for example) which basically covers all of the cases - at least, I can't imagine many regressions I could introduce which would cause any of these tests to fail and not that one.

There are a couple of other examples below which I will comment. Testing is good, but too much testing makes code brittle - if I have multiple tests testing a single flow through my application, if I want to make a change to the application then I have to fix multiple tests when I refactor. I want my tests to help identify regressions, but I don't want them to hinder me from refactors, so there is a balance there.

spec/FrameSpec.js
```diff
@@ -0,0 +1,299 @@
...
+    });
+
+    it("should be 0 for a strike", function() {
+      frame.play(10);
+      expect(frame.pinsRemaining()).toEqual(0);
+    });
+
+    it("should be 10 remaining on each ball in last frame if 3 strikes", function() {
+      lastFrame = new Frame(true);
+      expect(lastFrame.pinsRemaining()).toEqual(10);
```
mdsomerfield: We have many expectations in the same test - this is a form of duplication. We've already tested elsewhere that pinsRemaining behaves as expected. 

If we were to test that we could roll 3 strikes on the last frame (but no more), I think we'd have confidence that this code was working. I think we do test that further down, so I don't feel like these tests add anything.

spec/FrameSpec.js
```diff
@@ -0,0 +1,299 @@
...
+      lastFrame = new Frame(true);
+    });
+
+    it("should not allow scores over 10", function() {
+      frame.play(1)
+      expect(function() {frame.play(10);}).toThrow(new Error('Score for this game should not exceed 10'));
+    });
+
+    it("should allow scores up to 20 after second ball on last frame", function() {
+      lastFrame.play(10)
```
mdsomerfield: Are there any circumstances in which the tests on line 239 and 268 pass and this test fails?

interface.js
```diff
@@ -0,0 +1,75 @@
+$(document).ready(function() {
+  var match;
+
+  new_match();
+
+  $('#score0').on('click', function() { play(0); })
```
mdsomerfield: We could make this more DRY by assigning a class to the score buttons and doing something like: 

```
for (var i = 0; i < 10; i++) {
  $('.score-button')[i].on('click', function() { play(i); }
}
```

interface.js
```diff
@@ -0,0 +1,75 @@
...
+
+  function new_match() {
+    for ( i=0; i<=10; i++ ) {$('#score' + i).fadeIn();}
+    match = new Match(Frame);
+    $('#newGame').hide();
+    clear_scores();
+  }
+
+  function play(score) {
+    match.play(score);
```
mdsomerfield: So, rather than needing to use indexes in IDs, we could be a little more OO about this by thinking of our UI elements as instances of a class. For example:

```
function FrameView(element, frame) {
  this.frameElement = element;
  this.frame = frame;
}

FrameView.prototype.update() = function() {
  this.frameElement.find('.ball1').text(this.frame.ball1);
  this.frameElement.find('.ball2').text(this.frame.ball2);
  this.frameElement.find('.score').text(this.frame.score());
}

function updateCurrentFrameScores() {
  frameViews.forEach((frame) => { frame.update(); });
}

var frames = $('.frame');
var frameViews = [];

for (var i = 0; i < 10; i++) {
  frameViews.push(new FrameView(frames[i]));
}
```

The nice thing about this approach is that we can then also refactor our view so that it's not repeated:

```
FrameView.prototype.initialiseView() = function() {
  this.frameElement.html('<div class="balls"><span class="ball" id="ball1"></span><span class="ball" id="ball2"></span></div><span class="game_score" id="score"></span>');
}
```
And can remove a lot of the repeated HTML from the index file above.

mdsomerfield (commented):
Hi Phil,

I think this is a really good effort and you've got all of the elements there. There's also a lot to review, so I've tried to focus on two main elements:

1) You mentioned that you wanted to think more about HTML/CSS, so I've demonstrated a couple of ways you can make your HTML a bit more OO with javascript, which can help reduce duplication and improve readability/encapsulation.

2) You're writing a lot of tests, which is great, but it's also good to think about what tests you write. I found it harder to articulate this - if my comments aren't helpful please do let me know, happy to discuss this in more detail.

PR #681, opened 2018-02-04
Completed bowling challenge - Ed Litten 

ealitten:


lib/interface.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  function updateTotalScore() {
+    $("#current-score").text(game.score());
+  };
+
+  function writeRolls() {
+    for (n = 0; n < game.currentFrameNo(); n++) {
+      var frame = game.getFrames()[n]
+      $("#frame" + (n + 1) + "roll1").text(getFirstRoll(frame))
```
mdsomerfield: Rather than giving everything ids like 'frame1' and 'frame2', what if you conceived of your frame interface as a component with it's own class? So then rather than a table, you end up with something like:

```
function FrameView() { }

FrameView.prototype = {
  initialise: function() {
    this._element = $(this.frameViewElement).html(
      '<div class='frame-view'>' +
        '<div class='frame-number'></div>' +
        '<div class='frame-roll1'></div>' +
        '<div class='frame-roll2'></div>' +
      '</div>'
    );
  },

  appendTo: function(parent) {
    $(parent).append(this._element);
  }

  recordRolls: function(frame) {
    $(element).find('.frame-roll1').text(getFirstRoll(frame));
    $(element).find('.frame-roll2').text(getSecondRoll(frame));
  }
}

var parentElement = $('#score-card');
var frameViews = [];

for (var i = 0; i < MAX_FRAMES; i++) {
  var frameView = new FrameView();
  frameView.initialise();
  frameView.appendTo(parentElement);
  frameViews.push(frameView);
}

...

for (n = 0; n < game.currentFrameNo(); n++) {
  frameViews.setScores(get.getFrames()[n]);
}

```

I've missed some code above and I'm sure I've made a mistake or two typing it out, but the point is to demonstrate that UI components can also be considered in an OO way and this can help you reduce repetition and encapsulate some of your logic, which will scale better when building more complex UIs.

lib/game.js
```diff
@@ -0,0 +1,52 @@
...
+
+  currentFrame: function() {
+    return this.frames[this.frames.length - 1];
+  },
+
+  currentFrameNo: function () {
+    return this.frames.length
+  },
+
+  addFrame: function (frame) {
```
mdsomerfield: Adding frames requires something outside to know about the internal implementation of game. I would prefer to inject an array of frames at the start of the game and have the game manage which one is the currently active frame as the game develops. The way you've done this creates a coupling between game and whatever is calling it. 

Injecting an array of frames also means you can initialise your game in any state you choose, which can help with test setup.

spec/gameSpec.js
```diff
@@ -0,0 +1,108 @@
+describe('Game', function () {
+
+  var game;
+
+  beforeEach(function () {
+    game = new Game
+    frame = jasmine.createSpyObj('frame', {
```
mdsomerfield: Rather than generating all your test objects here, I would get into the habit of defining helper functions 'buildFrame()', 'buildStrikeFrame()' etc and calling them in your tests. Some reasons for this:

1) It cuts down on the amount of test setup you have to do for each test
2) It's clearer in the test where your objects are coming from
3) You can extend the helper functions to allow for slightly different setups, e.g:

```
function buildFrame(firstRoll = 3, score = 7) {
  return {
    'score': score,
    'isSpare': false,
    'firstRoll': firstRoll
  }
}
```

You can then use this more easily when you want create to frame stubs with different values (

lib/game.js
```diff
@@ -0,0 +1,52 @@
...
+
+  score: function () {
+    return this.getFrames().reduce(
+      (total, frame, index) => total + frame.score() + this.bonusScore(frame, index),
+      0
+    );
+  },
+
+  // works but needs some refactoring
+  bonusScore: function (frame, index) {
```
mdsomerfield: It feels to me like this should be a factor of the frame. Rather than calculating the bonus score here and holding that state outside the frame, why not tell the preceding frames 'here's a new roll' and let the frame decide whether it is waiting for a bonus? That way you can keep all of the frame score logic inside the frame. 

The fact that you struggled to a isolate your Game class (as you mention in your self review) is a side effect of your approach here - your Game is too tightly coupled with your Frame. Whenever you find yourself struggling to isolate logic for testing, ask yourself if you've really got the right design and consider entering a refactor step.

lib/game.js
ealitten: `addFrame` was actually only ever called by the game itself, so it was effectively a private method (just not marked as such). The outside world called addRoll and then the game decided whether it needed a new frame or not. However, I see your point re injecting an array of empty frames making testing much easier, so I've updated this

PR #688, opened 2018-02-11
completed all logic - Ben Pourian

bpourian:
I have completed all the logic testing for various outcomes and possibilities.


spec/gameSpec.js
```diff
@@ -0,0 +1,200 @@
...
+      game.pinsDroppedEachRoll(5);
+
+      expect(game.scoreArray).toEqual([4,5]);
+    });
+
+  });
+
+  describe('Total score after each frame', function () {
+    it('Return total score for 3 frames: 4 and 5 rolls', function () {
+      game.startGame(5);
```
mdsomerfield: It feels odd that we can start the same game 6 times. Worth thinking about method naming.

spec/gameSpec.js
```diff
@@ -0,0 +1,200 @@
...
+      game.startGame(1);
+
+      expect(game.finalScore).toEqual(49);
+    });
+
+    it('Return total score for 10 frames without a tenth frame bonus', function () {
+      game.startGame(0);
+      game.startGame(10);
+
+      game.startGame(10);
```
mdsomerfield: Long tests like this are usually either an indication that (a) we're testing the wrong thing or (b) our classes are too tightly coupled. I think it's a bit of both here. 

Regards (a) - are we really testing anything here that we aren't effectively testing in other tests? Specifically, the test on line 46 looks like a shorter version of this. I think what we want to test is that we can calculate the total score - in which case, if it works for 2 or 3 frames, testing for more frames doesn't really add anything. 

Regards (b) - you only have one class, so suggesting that classes are tightly coupled may sound strange, but it feels like there is logic here that should be refactored out of Game. If I had a separate Frame class, for example, and if I was injecting that into Game, then I could write a test something like:

```
var frames = [new Frame(0, 10), new Frame(7, 2)];
var totalNumberOfFrames = 2;
var game = new Game(frames, totalNumberOfFrames);

var totalScore = game.getTotalScore();

expect(totalScore).toBe(19);
```

Here I am effectively testing the same case but it is terser and easier to read and also more flexible.

src/interface.js
```diff
@@ -0,0 +1,76 @@
...
+      $('.box' + (j+ 1)).text(bowling.screenArray[j]);
+    }
+  }
+  };
+
+  $('#newgame').click(function() {
+    location.reload();
+  });
+
+  $('#btn0').click(function(){
```
mdsomerfield: This doesn't feel DRY - what about adding a class 'score-button' to all of the score buttons and doing something like:

```
for (var i = 0; i < 10; i++) {
  $('.score-button')[i].click(function() {
    bowling.pinsDroppedEachRoll(i);
    addScore()
  });
}
```

src/game.js
```diff
@@ -0,0 +1,162 @@
...
+      self.scoreArray.push(pinsdropped);
+      return true;
+    }
+  }
+};
+
+Game.prototype.strikeBonus = function (index) {
+  var self = this;
+
+  if (self.tenthFrame() && self.screenArray[index] === 'X') {
```
mdsomerfield: Whenever we see a big if statement - particularly a nested if statement - we should always be thinking about a refactor. Reasons why this is sub-optimal:

a) It's difficult to read
b) It's difficult to change
c) It's complicated to test all the different outcomes.

I started trying to write an example of how to refactor this, but it was going to be a pretty big endeavour - I think it's worth sitting down and pairing on this at some point during the week.

PR #695, opened 2018-02-21
Very late bowling challenge pull request

OutWooods:


spec/frameSpec.js
```diff
@@ -0,0 +1,207 @@
+/* eslint-env jasmine */
+const Frame = require('../src/frame');
```
mdsomerfield: Your unit tests look pretty clean and well written - good. However, there do seem to be a lot of them! I think it's great that you're writing such detailed tests, but if I think of the cases that I think I want to test between game and frame, it's probably not much more than:

- a single roll produces the score for a single roll
- if I roll again, that roll gets assigned to the next available slot
- once my turn is complete, the next frame becomes active
- if I fail to knock down the pins, there is no bonus
- if I half strike, there is a single roll bonus
- if I strike, there are two roll bonuses
- the overall score calculates the sum of the frames
- if I need a bonus on the last frame, I get the correct number of relevant rolls

I may have missed a couple of cases, but not too much. So if I found myself writing a lot more tests than that, I might start thinking 'is my code properly encapsulated?'.

src/format.js
```diff
@@ -0,0 +1,105 @@
...
+
+  let setTotal = () => {
+    document.getElementById('currentScore').innerHTML = game.score();
+  };
+
+  let individualScores = () => {
+    let collumnID = 1;
+    let scores = game.runningScores();
+    scores.forEach((score) => {
+      document.getElementById(`scoreFrame${collumnID}`).innerHTML = `${score}`;
```
mdsomerfield: String interpolation - woo! That's the first time I've seen this used (it's new in ES6 and I haven't done a lot of coding recently). 💯 

More seriously, it's great that you're building the view dynamically and in DRY fashion. For next level javascript, think about how you could treat each of the view elements as a class/object itself. E.g:

```
function FrameView(element) {
   this.element = element;
}

FrameView.prototype.setScore(score) {
  this.element.innerHTML = score;
}
```

src/format.js
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+
+  let createFieldWithID = (id) => {
+    const field = document.createElement('td');
+    field.id = id;
+    return field;
+  };
+
+  let createScoreCard = () => {
```
mdsomerfield: I would tend to write functions using the standard 

```
function createScoreCard() {
  ..
}
```

notation. The lambda style notation (I'm actually not sure if they are called lambda's in javascript, that's a C#ism) is used more with callbacks - they are a bit like Ruby code blocks. This code feels confusing to me because it looks like it's a very long method when really it's a set of function declarations.

src/frame.js
```diff
@@ -0,0 +1,55 @@
+const STRIKE = 10;
+const STANDARD_ROUND = 2;
+const STRIKE_ROUND = 1;
+const STRIKE_SCORE_ROUND = 3;
+
+class Frame {
+  constructor(strikeRound = STRIKE_ROUND, spareRound = STANDARD_ROUND) {
+    this.rounds = [];
+    this.rules = {
```
mdsomerfield: I like how you've turned the rules into a dictionary - that's known as a Strategy pattern and it's good practice with this kind of logic.

I would suggest however that you don't fall into the trap of treating all rolls the same. A roll may just be a number programmatically, but a first or second roll is different from a bonus roll in *purpose*. We should always think about what a things represents, not what represents the thing, and in this case I would suggest that if you treat rolls and bonuses and different things you'll end up with an even cleaner implementation.

src/game.js
```diff
@@ -0,0 +1,56 @@
...
+    return this.board.reduce((acc, frame) => acc + frame.score(), 0);
+  }
+
+  runningScores() {
+    let results = [];
+    this.board.forEach((frame) => { results.push(frame.score()); });
+    return results;
+  }
+
+  view() {
```
mdsomerfield: It feels like this is not the responsibility of this class.

PR #696, opened 2018-03-04
Bowling challenge PR - Daniel Keen

DKeen0123:


README.md
```diff
@@ -1,77 +1,28 @@
...
+I ran out of time to calculate a total and integrate spare and strike logic.
 
-In the image below you can find some score examples.
+## Personal Reflection
 
-More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
+This was a tough week for me. I started by thinking about what the application would look like, and had a complete mental block on how a person would interact with it. I was unable for some reason to see past just randomising the output of rolls, and only realised late Sunday evening of the first week that my interface could be interactive.
 
-![Ten Pin Score Example](images/example_ten_pin_scoring.png)
+I struggle greatly with planning, and feel lost until I start coding, and I am not really sure how to approach full stack applications like this in terms of planning. Should I have planned out the interface first and worked back to the logic? I started on the logic first as this was what was most comfortable for me as I was not used to browser-based interaction.
```
samjones1001: I would probably take the same approach.  In this case, I would expect your front end interactions to be pretty dependent on your logic, so this seems like the natural order to do things in.  For example, if you have a button which, when clicked, represents the number of pins knocked down, its not going to be able to do anything unless you can hook it up to a function which deals with knocking down pins.

README.md
```diff
@@ -1,77 +1,28 @@
...
 
-In code review we'll be hoping to see:
+I also could not figure out how to score strikes and spares properly, and ran out of time to do so.
 
-* All tests passing
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+## Key questions
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Note that referring to this rubric in advance may make the challenge somewhat easier.  You should be the judge of how much challenge you want.
+How do I go about planning building this type of application? How do I break it down? How do I visualise it?
```
samjones1001: Try to think about the simplest possible case for finishing a game of bowling - perhaps a game in which no pins are knocked down.  Think about how you could create a system which could handle this case.  Once it works, think about another simple case (a game of all ones?) and build upon your existing code.  If you keep repeating this until your code handles all the cases you can think of, you can gradually build up complexity, rather than trying to build a fully-realised, complex system in a single shot.

index.html
```diff
@@ -0,0 +1,52 @@
...
+      <button id="7" type="button">7</button>
+      <button id="8" type="button">8</button>
+      <button id="9" type="button">9</button>
+      <button id="10" type="button">10</button>
+    </div>
+
+    <!-- <input id="player-name" type="text" placeholder="Your name...">
+      <button id="player-name-submit"type="button" name="button">Start Game!</button>
+
+      <button id="bowl-button" type="button" name="button">Bowl</button> -->
```
samjones1001: This is a minor point, but try to get in the habit of removing commented out code before you submit it.  If you were to leave it in when submitting a tech test, for example, it just looks a little messy.

src/game.js
```diff
@@ -0,0 +1,43 @@
...
+
+  strikeTracker: function(frame) {
+    if (frame.firstRollScore === 10) {
+      frame.secondRollScore = 0;
+      return true;
+    }
+    return false;
+  },
+
+  spareTracker: function(frame) {
```
samjones1001: Is this function ever use anywhere?  I'm guessing that you had plans to utilise it later, but at the moment I think its just taking up lines :)

src/game.js
```diff
@@ -0,0 +1,43 @@
...
+    if (frame.total() === 10) {
+      return true;
+    }
+    return false;
+  },
+
+  createNewFrame: function() {
+    this.frameTwo = new BowlingFrame();
```
samjones1001: This feels a little confusing to me - after the first frame is completed, it seems like every subsequent frame is going to be stored in the frameTwo instance variable which doesn't feel right.  I think you could probably do away with the frameOne and frameTwo instance variable altogether and just rely on the currentFrame to track where you are in the game.

src/interface.js
```diff
@@ -0,0 +1,76 @@
...
+  //     console.log('first');
+  //     var $score = game.bowl(game.currentFrame);
+  //     $('#scoreboard').append(
+  //       `<tr id="${frameNumber}"><td>${frameNumber}</td><td>${game.currentFrame
+  //         .rollNumber - 1}</td><td>${$score}</td></tr>`
+  //     );
+  //   }
+  // });
+
+  $('.buttons')
```
samjones1001: I would definitely agree with you that this could do with some refactoring and breaking out into separate functions.  It's good that you identified this as something you would like to have done, and I'm sure given more time you would have been able to.

PR #697, opened 2018-03-04
Bowling challenge completed - Full reflection done in README

marcusfgardiner:


spec/scoreCalculatorSpec.js
```diff
@@ -0,0 +1,183 @@
+describe('ScoreCalculator', function(){
+
+  var scoreCalculator;
+
+  beforeEach(function(){
```
EdwardAndress: If you're only using each game in one test it'd be clearer to assign it right next to where it's being used (and they don't all need to be assigned before every test).  Should you find that you're using a given game in more than one test and you want to DRY up your test suite, it might be clearer to use a descriptive variable name – these are quite complex data sets though so that could be hard :)

spec/scoreCalculatorSpec.js
```diff
@@ -0,0 +1,183 @@
...
+      });
+      it('can add up a 10th frame score with a strike in the 10th frame only', function() {
+        scoreCalculator.calculateTotalScore(game10);
+        expect(scoreCalculator.totalScore).toEqual(33);
+      });
+    });
+
+    describe('Strike scoring', function(){
+
+
```
EdwardAndress: sounds like this method is for calculating the score when two strikes are made in a row :)

views/index.html
```diff
@@ -0,0 +1,71 @@
...
+  <body>
+    <header>
+      <h1> Bowling Scoresheet <h1>
+    </header>
+
+
+    <div id='scorecard'>
+      <img class='strike' src="../public/Strike.gif">
+
+  <table id='scorecardTable' class='scorecard' cellpadding='1' cellspacing='0'>
```
EdwardAndress: It's possible to build this table using js, leaving you with fewer lines of HTML to write :)

src/interface.js
```diff
@@ -0,0 +1,47 @@
+$(document).ready(function (){
+  var game = new Game();
+
+  $( '#bowlingBall' ).click(function() {
```
EdwardAndress: There's a 'smell' here.  The fact that you're calling several different methods that accept the `game` as a parameter, one after the other, makes me think that there's something which could be abstracted from here into another method.

Also, the interface seems to combine 'game engine' and presentation functionality.  Printing score is the latter.  Checking whether a roll is valid or the game is over is the former.  Interface sounds like a presentational layer so the engine-like bits feel out of place.  Perhaps some of them could be moved to the game?

views/index.html
marcusfgardiner: Hmm... I imagine there is a way of writing javascript code to do lots of HTML things for you, but haven't seen/ used this yet

src/interface.js
marcusfgardiner: Could an abstraction be a method that calls all these methods that use game? This method would then have game flowing through all of them (like how my score calculator works)

PR #699, opened 2018-03-04
Bowling Challenge PR + Personal review - Josué Estévez Fernández

Jestfer:


public/js/interface.js
```diff
@@ -0,0 +1,134 @@
+const game = new Game();
+let frame = new Frame();
+let i = 0;
```
samjones1001: This could be more descriptively named - 'i' doesn't really communicate what this variable is supposed to represent.  Maybe it could be called frameCount or something similar.

public/js/interface.js
```diff
@@ -0,0 +1,134 @@
...
+
+    if (count === 1) {
+      $('.first-roll').last().html(id);
+    } else if (count === 2) {
+      $('.second-roll').last().html(id);
+    } else if (count === 3) {
+      $('.third-roll').last().html(id);
+    }
+
+    if (game.frames.length < 9) {
```
samjones1001: Big blocks of nested if/else conditions like this can be difficult to follow and make it hard for other people to understand what your code is intended to do.  Can you think of any ways in which you could break it up a bit?  You'll have a chance to practice this in week 10, as one of the exercises you'll do is all about doing exactly this.

src/Frame.js
```diff
@@ -0,0 +1,59 @@
...
+
+  return pinsKnocked;
+};
+
+// PRIVATE METHODS
+
+Frame.prototype.isRollLegit = function isRollLegit() {
+  const newCounter = this.rollCounter + 1;
+
+  if (this.strike === false && this.spare === false && this.frameNum === 10 && newCounter === 3) {
```
samjones1001: Sometimes if I'm checking against a long list of conditions like this I like to create a descriptive variable name to make things a bit more readable.  So something like

var isNotStrikeOrSpareIsFinalFrameFinalRoll = this.strike === false && this.spare === false && this.frameNum === 10 && newCounter === 3

if (isNotStrikeOrSpareIsFinalFrameFinalRoll) {
    throw new Error('The game is over!');
}

This is just a personal preference, but I think it helps to explain what your code is trying to do.

src/Frame.js
```diff
@@ -0,0 +1,59 @@
...
+  } else if (this.strike === true && this.frameNum === undefined) {
+    throw new Error('You scored a strike. Start next frame!');
+  } else if (newCounter >= 3 && this.frameNum === undefined) {
+    throw new Error('You cannot roll again. Start next frame!');
+  }
+
+  this.rollCounter += 1;
+};
+
+Frame.prototype.isStrike = function isStrike(pinsKnocked) {
```
samjones1001: This function feels to me like it has multiple responsibilities at the moment.  an isStrike function sounds like it should just check whether a particular roll is a strike, this seems to do that, then calculate the score, then push the rolls into storage.  Could you perhaps break some of this functionality out into it's own function in order to keep to the single responsibility principle.

src/Game.js
```diff
@@ -0,0 +1,42 @@
+function Game() {
+  this.frames = [];
+  this.totalScore = 0;
+}
+
+Game.prototype.addFrame = function addFrame(frame) {
+  this.frames.push(frame);
+};
+
+Game.prototype.finalScore = function finalScore() {
```
samjones1001: As you said in your readme, this function is probably doing too much at the moment.  It would benefit from some refactoring and breaking out into some smaller functions.  It also has lots of nested if/else logic which makes it a bit difficult to follow.

public/js/interface.js
Jestfer: Hey Sam! Thanks a lot for your review again!
The variable 'i' represents an accumulator in both cases, a tricky helper to allow me to generate pins (first case) and hide them as they are knocked (second case).
In the second case it helps me in a for loop to iterate through pin and hide them until the pin reaches the id num of the knocked one.

I have renamed 'i' in both cases following your suggestion as follows:
1- **pinNum**
2- **pinIterator**

PR #703, opened 2018-03-04
Complete code + reflection in readme

Kotauror:


jasmine-standalone-2.3.4 (4)/src/Game.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// ADD FRAME
+
+Game.prototype._addFrame = function() {
+  let framesCount = this.frames.length;
+  let lastFrame = this._lastFrame();
+  let secondLastFrame = this._secondLastFrame();
+
+  if (framesCount === 0 || framesCount === 1) {
```
mdsomerfield: A multi-pronged if statement, and a nested if statement in the last fork! Whenever we see nested conditionals, or conditionals with more than 2 distinct paths, we should be thinking about whether we should refactor our code.

In this case for example, I might follow the process:

Q: What is the outcome of this method?
A: We push a number to the frames collection.

Q: What does the number represent?
A: The frame that the last roll relates to

Q: Why is this complicated?
A: Because we have multiple cases

Q: What are the cases?
A: We only iterate the frame when there have been two rolls, so a roll can be either the first or the second, and also there is a special case relating to the last frame. 

Q: So we have a two to one relationship rolls to frames... sounds like we could abstract that?
A: Hmm. Yes. If frames were a class, and we made it responsible for knowing if this was the first or second roll, we could lose one of the if statements.

Q: Great. And the last frame case... why is that special?
A: Because we can multiple additional rolls if we score a strike or a spare.

Q: Are those additional rolls treated the same as normal rolls? 
A: No. They are only applied as bonuses to the last frame. 

Q: Do we have bonus rolls anywhere else?
A: Yes. All frames can have bonus rolls.

Q: So it's not a special case.
A: But bonus rolls for the last frame don't count towards other frames, because there are no more frames.

Q: Maybe the problem is that we are treating bonuses and rolls as the same thing. 
A: Ah, yes, if our frame class had 'rolls' and 'bonuses' as separate properties, then I'm pretty sure I can treat all frames the same. Then the other conditional goes away. Awesome!

This is obviously an imaginary conversation. But the key things to highlight are:
- When you see nested conditionals or conditionals with multiple pathways, recognise this as a sign that something could be better.
- Refactor it early. The longer you leave it to refactor the harder it will be to unpick.
- Interrogate your code. You could try using the 'Five Whys'. Basically, the goal is to keep challenging your assumptions until you're able to unpick them to something cleaner.

jasmine-standalone-2.3.4 (4)/src/Game.js
```diff
@@ -0,0 +1,156 @@
...
+
+}
+
+// COUNTING METHODS
+
+Game.prototype.basicScore = function() {
+  var counter = 0;
+  for (i = 0; i < this.pairs.length; i ++) {
+   if (this._valueOfPair(i) !== undefined) {
+      if (i < 18) {
```
mdsomerfield: We have a lot of magic numbers here (18, 10, 19 etc). Generally we should avoid this. If we have a constant which represents something we should make it a named constant e.g. MAX_FRAMES = 10 and use that. If we are deriving other numbers from constant we should make it clear how they are derived e.g. (MAX_FRAMES * 2) - 2 and (MAX_FRAMES * 2) - 1. 

Doing this makes it clearer what the intent is. As it stands, it's hard to understand what is happening here.

jasmine-standalone-2.3.4 (4)/src/Game.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+  if (this._valueOfPair(18) + this._valueOfPair(19) === 10) {
+    spares += this._valueOfPair(20)
+  }
+  return spares;
+}
+
+Game.prototype.strikes = function() {
+  var strikes = 0;
+  for (i = 0; i <= 14; i += 2) {
```
mdsomerfield: I don't really understand what is going on here. I think that we're trying to get the bonuses from the next frame in the case of a strike. When performing calculations, it's worth putting them in strongly named methods so it is easy for another developer to read what is happening. Eg:

```
if (!nextFrameIsAStrike()) {
  strikes += getRollsFromNextFrame()
} else {
  strikes += getFirstRollFromNextFrameAndFirstRollFromFrameAfter()
}
```

Obviously that last method name is really long and feels kind of unnecessary... which is another reason to do this. Because when you end up with method names like that, it's another hint that you should probably refactor your code (see comment above). If I can't express the logic in a simple, well named method, then I probably need to interrogate it.

For example, here I would be thinking 'actually, really what I want is just the next two rolls... it's just that if the next frame is a strike, then it has only a single roll. How could I refactor this so that didn't matter?'.

jasmine-standalone-2.3.4 (4)/spec/GameSpec.js
```diff
@@ -0,0 +1,243 @@
...
+
+  describe("many-rolls game - counting score", function() {
+
+    beforeEach(function() {
+      player.roll(4);
+      player.roll(4); //
+      player.roll(5);
+      player.roll(3); //
+      player.roll(4);
+      player.roll(5); //
```
mdsomerfield: Do we need to do all of these rolls? What does rolling 10 times test that rolling twice does not? 

This is another sign that we should be refactoring. Our tests should be focussed on specific, isolated actions - I roll and it records the score. I roll twice and it records a frame score. I score two frames and the game adds them together correctly. I roll a strike and it records the bonuses correctly.

If we're having to repeat a lot of actions in test setup then we're either introducing a lot of points of failure (brittle tests) or we are not really adding anything with many of the actions (redundant tests).

PR #706, opened 2018-03-04
Submit Bowling Challenge

MatthewBurstein:


spec/frameSpec.js
```diff
@@ -0,0 +1,169 @@
...
+      strikeRoll = 10;
+      frame.multipliers = [1];
+    });
+
+    it("when final frame, leaves first multiplier untouched, sets second to 1", function() {
+      finalFrame.multipliers = [1];
+      finalFrame.rolls = [strikeRoll];
+      finalFrame._calculateMultipliers();
+
+      expect(finalFrame.multipliers).toEqual([1,1]);
```
mdsomerfield: I think I've worked out what 'multipliers' is... (I haven't looked at the implementation yet, just the tests). I think this is related to bonuses. When it comes to naming and implementation, it's useful to think about whether the purposes of a method or property would be obvious to somebody who was coming into the code from the cold. I might be tempted to call this 'bonusMultiplier' or something along those lines, to make it clear what the context of the multiplier is.

spec/frameSpec.js
```diff
@@ -0,0 +1,169 @@
...
+    describe("when final frame", function() {
+      it("calls this._finalFrameIsComplete()", function() {
+        spyOn(finalFrame, "_finalFrameIsComplete");
+        finalFrame.isComplete();
+
+        expect(finalFrame._finalFrameIsComplete).toHaveBeenCalledWith();
+      });
+    });
+
+    describe("._notFinalFrameIsComplete()", function() {
```
mdsomerfield: We don't need to test private methods!

src/frame.js
```diff
@@ -0,0 +1,54 @@
+function Frame(multipliers, isFinalFrame) {
```
mdsomerfield: I like that you're encapsulating the fact that this is the final frame within the frame (rather than treating them as separate classes).

I think it's possible not to even need a special case though :). For me, the frame's responsibility is to allow up to two rolls to knock over up to 10 pins, and then to accept bonus scores if applicable. This doesn't change if it's the last frame - what does change is that further rolls only count towards the bonus and not towards another frame. This feels like the game's responsible to manage.

src/frame.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+};
+
+Frame.prototype._updateScore = function(pinsKnocked) {
+  let currentMultiplier = this.multipliers.shift();
+  this.score += (pinsKnocked * currentMultiplier);
+};
+
+Frame.prototype._calculateMultipliers = function() {
+  let multipliers = this.multipliers;
```
mdsomerfield: I'm afraid I'm struggling to understand this implementation of multipliers for bonuses. Might be worth discussing.

src/frontEnd.js
```diff
@@ -0,0 +1,96 @@
...
+  $(".roll-button").click(function() {
+    if (game === null) {
+      alert("Please fill out your name and click start game");
+    }
+    let pinsKnocked = parseInt($(this).attr("name"));
+    game.roll(pinsKnocked);
+    let frameNumber = game.frameNumber;
+    let thisFrame = game.frames[frameNumber - 1];
+    let score = game.score;
+    if(thisFrame.rolls.length === 1) {
```
mdsomerfield: It feels like these if statements could be simplified. What we essentially want to do, once we have handled the game mechanics stuff above and extracted the score, is:

(1) find the relevant element
(2) set the text of this element to the roll
(3) update the total frame score (if the frame is complete)
(4) hide any buttons that are invalid for the next roll (if the frame is not complete)

If we rethink our HTML slightly and uses classes to denote element types, we can do (1) e.g by:

```
var rollElement = $('.frames')[frameNumber - 1].find('rolls')[thisFrame.rolls.length - 1);
```

Then the rest of this can be done in sequence without branching.

PR #713, opened 2018-03-05
Lewis bowling challenge.

l3w15:


l3w15:
Submitted for individual code review!

src/game.js
```diff
@@ -0,0 +1,98 @@
...
+Game.prototype = {
+
+  addBonuses() {
+    var [frames, l] = [this.frames, this.frames.length];
+    var [antepenult, penult, ult] = [frames[l - 3], frames[l - 2], frames[l -1]];
+    if (l > 2 && this.isStrike(antepenult) && !this.hasBonus(antepenult)) {
+      antepenult.push(ult[roll1]);
+      ult[roll2] ? penult.push(ult[roll1], ult[roll2]) : penult.push(ult[roll1]);
+    } else if (l > 1 && this.isStrike(penult) && !this.hasBonus(penult)) {
+      ult[roll2] && ult[roll2] !== 10 ?
```
mdsomerfield: Whenever we see multi-pronged if statements, we should ask ourselves whether we can refactor/simplify. If we see several of these in the same class (as we do here), then that is an even stronger indication, and it usually means that we should push some logic out into another class. I mentioned your commented out frame class above - was that the original intention here?

A technique I use here is something like the 5 whys - I keep asking myself why I need the conditions. This can help me to work out whether there's a more fundamental and potentially universal truth. E.g.

Why do I have 3 conditions? 
Because the last 3 frames behave differently with regards to bonuses

Why do they behave differently with regards to bonuses?
Because strikes and half strikes need bonuses but their may not be enough frames to play

Why may there not be enough frames to play?
Because we've reached the end of the game

Why does reached the end of the game mean we don't have enough frames?
Because we want to continue rolling but we need a frame to record the score against

Why do we need a frame to record the score against?
... erm, we don't. 

At this point, I might start thinking that actually, all the frames as the same, and they all behave the same with regards to bonuses, and the only difference is that the game allows me to take a couple of extra rolls if a bonus is still pending near the end. That ends up being a much simpler case - in simplified pseudocode, it's basically something like:

```
function isGameOver() {
  return frames.find((frame) => frame.needsRolls() || frame.needsBonuses()) === null;
}

while (!isGameOver()) {
  nextRoll();
}

```

src/game.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+  },
+
+  addRollToFrame() {
+    if (this.frames.length < MAX_FRAMES && !this.frames[this.frameCount]) {
+      this.frames[this.frameCount] = [this.roll];
+    } else if (this.frames.length < MAX_FRAMES) {
+      this.frames[this.frameCount].push(this.roll);
+    } else {
+      this.frames[this.frameCount] ?
```
mdsomerfield: When we're using something like this a lot, it's a good idea to give it a meaningful name e.g.

```
function getActiveFrame() {
  return this.frames[this.frameCount]; 
}
```

It reveals more intent about your code and makes any potential refactor easier.

src/interface.js
```diff
@@ -0,0 +1,89 @@
...
+
+  function createScoreCardAndButtonsHTML() {
+    for (var i = 0; i < 11; i++) {
+      $('.roll-buttons').append(
+        `<button class="roll-score" id="roll${i}" value="${i}">${i}</button>`
+      )}
+
+    for (var i = 9; i > 0; i--) {
+    $('.scorecard').prepend(`<div class="scorecard-box" id="f${i}">
+      <div class="header">Frame ${i}</div>
```
mdsomerfield: I like that you've kept this DRY. For next level OO, you could try combining this with your displayFrames method below and have a frameView class. It might look something like:

```
function FrameView(parentElement) { 
  this.parentElement;
}

FrameView.prototype.render = function() {
   .. a variation of your code from createScoreCardAndButtonsHTML, appended to parentElement ...
}

FrameView.prototype.update = function() {
  ... a variation of your code from displayFrames...
}
```

src/game.js
```diff
@@ -0,0 +1,98 @@
...
+        || this.frames[this.frameCount].length === 2) {
+      this.addBonuses();
+      this.frameCount ++;
+    }
+  },
+
+  checkFrameTenComplete() {
+    this.addBonuses();
+    if (this.frames[this.frameCount].length === MAX_ROLLS
+      || this.frames[this.frameCount][roll1]
```
mdsomerfield: Whenever I see a multidimensional array, I wonder whether I'm missing a class. This feels similar to reaching into the state of another object, i.e. manipulating the roll property of a frame class from the game class. Also, we're doing a lot of calculations with this.frames[this.frameCount] - that kind of repetition is another indication that this logic should be encapsulated elsewhere. 

I would expect instead to be seeing something like:
```
this.activeFrame().isFrameComplete() 
```
And move the logic into a frame class.

PR #766, opened 2018-07-16
Bowling challenge

neobay991:
See README for details


jasmine/jasmine-standalone-3.1.0/src/Score.js
```diff
@@ -0,0 +1,125 @@
...
+  } else if (this.isFourthRollStrike(index)) {
+    this.addBonus(index);
+  } else if (this._rollScore[index - 4] +
+    this._rollScore[index - 3] === 10) {
+      this.addBonus(index);
+  }
+};
+
+Score.prototype.addBonus = function(index) {
+  var bonus = 0;
```
mdsomerfield: This method feels unnecessarily complex. Wherever you have magic numbers in your code, you should be thinking 'can I simplify this?'. By magic numbers, I mean that you have different logic depending on whether a value is a 1,2,3,4 etc. 

Applying a strike bonus should be the same in every case - why have different methods depending on whether it is a second roll, fourth roll or sixth roll?

PR #767, opened 2018-07-16
Let's go bowling Dude

heuperman:


heuperman:
I put in less work into the weekend challenge than usual this weekend as I attended PLIBMTTBHGATY on Saturday were I co-created [a (not very functional) React app](https://github.com/heuperman/pomodoro-react) instead.

PR #780, opened 2018-09-10
Weekend challenge

andrewwood2:


spec/ScoreSpec.js
```diff
@@ -0,0 +1,108 @@
+describe("Game", function() {
+  var score;
+  var frame;
+  var display;
+
+  beforeEach(function() {
+    game = new Game();
+    frame = new Frame();
```
mdsomerfield: If you're testing Game class, then I would expect frame, display and other dependencies to be mock objects

spec/ScoreSpec.js
```diff
@@ -0,0 +1,108 @@
...
+    display = new Display();
+  });
+
+  it("should have an initial score of 0", function() {
+    expect(game.score).toEqual(0);
+    expect(game.cumulatives).toEqual([]);
+  });
+
+  describe('Retrieving score from Frame', function() {
+    it('uses frame to get the score', function() {
```
mdsomerfield: First, I would expect game.score probably to be a method, game.getScore(), since I think it's dynamic state dependent on other things. 

I would then expect the test here to look something more like:

```
game.getScore();
expect(frame.getScore).toHaveBeenCalledOnce();
```

Since that actually tests that it calls frame - what you are testing below would pass even if frame wasn't used and it calculate the score inside the game class.

src/Frame.js
```diff
@@ -0,0 +1,60 @@
...
+  this.isOver = false;
+
+  this.bonusBalls = 0;
+  this.bonusscore = 0;
+  this.bonusAdded = false;
+};
+
+Frame.prototype.add_score = function(score) {
+  this.scores[this.ball] = score;
+  this.flatscore += score;
```
mdsomerfield: Do you need this? Could you not instead have a getScore method which added the scores of the this.scores?

src/Frame.js
```diff
@@ -0,0 +1,60 @@
+function Frame() {
+  // This tracks the rolls for a frame
+  this.scores = [null, null];
```
mdsomerfield: I think most of these should be private state, indicated as such with an underscore

src/Score.js
```diff
@@ -0,0 +1,142 @@
...
+  if (previousFrame.bonusAdded === false) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+Game.prototype.allBonusBallsRolled = function () {
+    lenFrames = this.frames.length
+    lenCumulatives = this.cumulatives.length
```
mdsomerfield: This logic confuses me. I'm not sure why you are maintaining cumulatives as independent state - you seem to be tending towards pushing as much stuff into state variables as possible when a lot of what you are holding there could be calculated dynamically. The cumulative score is just the sum of all the frame scores (with bonuses), so why not just have a method getOverallScore() which gets the sum of all the frame scores?

src/ScorecardController.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  function populate_with_new_rows_to_9(new_body, frames, cumulatives) {
+    counter = 0
+    frames.forEach( function(frame) {
+      if (counter < 9) {
+        var [row, framenum, roll1, roll2, roll3, frameTotal] = insert_row(new_body)
+
+        framenum.innerHTML = counter + 1
+        add_roll_1(roll1, frame.scores[0])
```
mdsomerfield: Try and avoid using magic numbers. If you have methods with numbers in them there is almost certainly an opportunity to refactor - add_roll should probably just be a single method, and the index of the roll to add could be a parameter.

PR #789, opened 2018-09-16
Malachy Bowling Challenge

Mallig:
The bonuses work, just trust me and don't look at the code.


src/Card.js
```diff
@@ -0,0 +1,126 @@
...
+// this function removes frames rolled for final frame bonus
+// it's necessary because the final frame logic doesn't exist
+Card.prototype._endGame = function() {
+  while (this.frames.length > 10) {
+    this.frames.pop()
+  };
+};
+
+Card.prototype.score = function() {
+  var score = 0
```
mdsomerfield: This method really confuses me. What is the purpose of having a local array called frames when you already have a property this.frames? At the very least, the naming needs to be less ambiguous so that the purpose of each is clearer.

PR #794, opened 2018-10-07
Completed javascript backend, with build passing on Travis, and HTML interface.

camjw:
I've linted most things and have good test coverage for all files except the interface which is untested. I thought that the way I would test it was through Capybara feature tests but I have not (and don't really plan to) integrated this into a Ruby app.


src/finalFrame.js
```diff
@@ -0,0 +1,69 @@
...
+  } else if (score > this.remaining) {
+    throw new Error('You cannot knock down more pins than there are standing.');
+  }
+  this.scoreRoll(score);
+};
+
+FinalFrame.prototype.scoreRoll = function scoreRoll(score) {
+  this.rolls.push(score);
+  if (this.rolls.length === 3) {
+    this.over = true;
```
mdsomerfield: If you have state, you should declare it in your constructure. Does this need to be state though? Could it not be a method, e.g. 

```
function isOver() { 
  return this.rolls.length >= 3 || this.rolls.length >= 2 && this.score() < 10;
}
```

Generally unless we really care about performance, we want to keen this stuff dynamic, as then we know that it's properly reflective of the state. Otherwise we could introduce more complexity, in that we have to check that the this.over state is maintained alongside the score.

src/finalFrame.js
```diff
@@ -0,0 +1,69 @@
...
+  } else if (this.rolls.length === 2) {
+    this.secondRoll();
+  } else if (this.rolls.length === 1) {
+    this.firstRoll();
+  } else {
+    throw new Error('Incorrect rolls length.');
+  }
+};
+
+FinalFrame.prototype.firstRoll = function firstRoll() {
```
mdsomerfield: Anywhere where I see a number in the method I am suspicious - could this be refactored? I think you could simplify this but just having a 'roll' function and using array indices. You may feel like there is extra logic on the second roll, but I think that could be generalised.

src/finalFrame.js
```diff
@@ -0,0 +1,69 @@
...
+    this.remaining = 10;
+  } else {
+    this.remaining -= this.rolls[1];
+  }
+};
+
+FinalFrame.prototype.calculateScore = function calculateScore() {
+  if (this.strikes[0] === true) {
+    if (this.strikes[1] === true) {
+      return 20 + (this.rolls[2] || 0);
```
mdsomerfield: Similar to numbers in method names, whenever I see a lot of random numbers I am suspicious. Could you not just do `this.rolls.sum((roll) => roll)`? The fact that you seem to have the possible for this.rolls[x] to be null seems to have introduced complexity - the score is really just the sum of the items in the array though.

src/interface.js
```diff
@@ -0,0 +1,93 @@
+$('document').ready(function runDocument() {
+  let scorecard = new Scorecard(Frame, FinalFrame);
+
+  refreshPoints = function refreshPoints() {
+    for (let i = 0; i < 10; i += 1) {
+      const frame = scorecard.frames[i];
+      for (let j = 0; j < frame.rolls.length; j += 1) {
+        if (frame.rolls[j] === 10) {
+          $('#frame' + i + 'roll' + j).text('X');
```
mdsomerfield: See comment near the top about making presentation code DRY - if you use view classes then a lot of this becomes simpler and you have fewer nested if statements.

PR #800, opened 2018-10-14
Bowling challenge: backend + interface

m-rcd:


src/interface.js
```diff
@@ -0,0 +1,88 @@
...
+
+  $('#namesub').on('click', function(event) {
+    event.preventDefault()
+    let name = $('#nameInput').val()
+    $('#nameTable').text(name)
+    $('#playername').hide()
+  })
+
+  $('.scoreButton').on('click', function(data) {
+    var input = Number(data.currentTarget.innerHTML)
```
mdsomerfield: This method is pretty hard to read and understand - you could probably refactor it into some named methods. I think that's actually true to a greater or lesser extent of some of the rest of the code in this file - the drive to write clean, readable code should still apply when building the interface - a developer should be able to read this file and understand what each bit of code is doing without having to think too hard.

PR #802, opened 2018-10-14
Bowling Challenge

JonathanAndrews:
Bowling Challenge with jQuery interface
Linted with ESLint
Successful Travis CI build


spec/frame.test.js
```diff
@@ -0,0 +1,41 @@
...
+    it('inserts score into scores array', function () {
+      frame.addScore(3)
+      expect(frame.scores[0]).toEqual(3)
+    })
+
+    it('only accepts three scores', function () {
+      frame.addScore(7)
+      frame.addScore(3)
+      frame.addScore(4)
+      frame.addScore(1)
```
mdsomerfield: I would expect the result of trying to add more than 3 scores would be an exception, rather than silent failure. When we do invalid or unexpected things in code, we want to make it obvious to the code that is calling that this is invalid. I think this applies to a few other bits of your code as well - when you validate, through an exception or make it obvious in some other way that the thing that has been attempted has failed, and let the code that is calling the method deal with it (usually by catching the exception and outputting some sort of error message to the user to a log file).

spec/game.test.js
```diff
@@ -0,0 +1,110 @@
...
+      game = new Game()
+    })
+    it('sums the score of all frames', function () {
+      game.roll(3)
+      game.roll(3)
+      game.roll(3)
+      expect(game.calculateTotalScore()).toEqual(9)
+    })
+
+    it('works for spares', function () {
```
mdsomerfield: Try and be a bit more descriptive with your tests. Remember that part of the purpose of tests is to serve as documentation for the intent of the code. Saying something 'works' isn't as descriptive as describing what you actually want it to achieve, e.g. in this case 'applies a bonus on the next roll when the frame is a spare'.

public/src/game.js
```diff
@@ -0,0 +1,62 @@
...
+
+Game.prototype._isValidRoll = function (score) {
+  if (isNaN(score)) {
+    throw new Error('Not a Number')
+  }
+  this._pins().isImpossibleScore(score)
+}
+
+Game.prototype._pins = function () {
+  return this.pins
```
mdsomerfield: I'm confused by this - you're wrapping a public variable in a private accessor?

PR #815, opened 2018-11-10
Working full game w/ final frame

jebax:


spec/final-frame-spec.js
```diff
@@ -0,0 +1,66 @@
...
+
+  beforeEach(function () {
+    finalFrame = new FinalFrame()
+  })
+
+  describe('A final frame', function () {
+    it('should inherit from a regular frame', function () {
+      spyOn(Frame.prototype, 'getBowls')
+      finalFrame.getBowls()
+      expect(Frame.prototype.getBowls).toHaveBeenCalled()
```
sjmog: Admirable that you're so thorough with your testing but remember that we should test a program's behaviour, _not_ its structure. Testing inheritance is 100% testing structure and makes it harder to change the structure of your app later.

A good question to ask is: "what sorts of changes would break this test? Do I want them to?" – if yes, it's a good test.

spec/final-frame-spec.js
```diff
@@ -0,0 +1,66 @@
...
+      expect(finalFrame.getScore()).toEqual(15)
+    })
+
+    it('should be 30 for 3x strikes', function () {
+      finalFrame.addBowl(10)
+      finalFrame.addBowl(10)
+      finalFrame.addBowl(10)
+      expect(finalFrame.getScore()).toEqual(30)
+    })
+  })
```
sjmog: These tests are beautiful – succinct, clear, itemised, and so so easy to read. Nice job.

spec/frame-spec.js
```diff
@@ -0,0 +1,101 @@
...
+      secondFrame.addBowl(7)
+      expect(frame.getScore(secondFrame)).toEqual(17)
+    })
+
+    it('strike bonus will be 20 when next 2 frames are strikes', function () {
+      frame.addBowl(10)
+      secondFrame.addBowl(10)
+      thirdFrame.addBowl(10)
+      expect(frame._strikeBonus(secondFrame, thirdFrame)).toEqual(20)
+    })
```
sjmog: Again – lovely. There's strong evidence here that you've built up the concept of a "frame" from basic, basic behaviour to gradually something more complex worthy of the name "frame".

It's strong evidence that you can Test-Drive things, in my opinion. Great 👍

spec/game-spec.js
```diff
@@ -0,0 +1,146 @@
+describe('Game', function () {
+  var game
+  var frame
+  var currentFrame
+
+  beforeEach(function () {
+    game = new Game()
+    game.startNextFrame()
+    currentFrame = game.getCurrentFrame()
+    frame = jasmine.createSpyObj('frame', ['getScore', 'addBowl', 'isFinished'])
```
sjmog: Oh I just love this setup. Each test is thoroughly independent, meaning you shouldn't ever leak state between your tests. Great great great

src/game.js
```diff
@@ -0,0 +1,42 @@
...
+  if (this.isFinished()) { throw 'Cannot add bowl - game over!' }
+  if (this.getCurrentFrame().isFinished()) { this.startNextFrame() }
+  this.getCurrentFrame().addBowl(pins)
+  this.getCurrentScore()
+}
+
+Game.prototype.isFinished = function () {
+  if (this._frames.length >= 10 && this.getCurrentFrame().isFinished()) {
+    this._finished = true
+    return this._finished
```
sjmog: I don't know if this is intentional, but lines 39 and 40 are a great pattern. It's tempting to set something and then return the value we set, rather than looking up the value we just changed wherever we just changed it – but you didn't do that, and that's a good thing.

It's especially important when there might be a lag between setting something and doing the next step – and when you hit Promises (in week 7) and `async`/`await`, you'll see how important this is.

styles.css
```diff
@@ -0,0 +1,116 @@
...
+  bottom: 0;
+  left: 0;
+  height: 15%;
+  width: 100%;
+  background-color: #F1BE47;
+  border-style: solid;
+  border-width: 4px 0px 0px 0px;
+  border-color: #BC9437;
+  display: block;
+}
```
sjmog: There's an opportunity here to think about styles that are shared across similar HTML components. Can you reduce the duplicate styling across these two different `section` entities?

sjmog: (Think of it like inheritance)

PR #835, opened 2018-11-12
Bowling Challenge

BillMux:


bowling.html
```diff
@@ -0,0 +1,102 @@
+<!DOCTYPE html>
+<html>
+
+  <head>
+    <script src="js/jquery-3.1.0.js"></script>
```
jonurry: Usually, you would put script tags after the <body> so that they don't block the page loading.

src/bowlingApp.js
```diff
@@ -0,0 +1,132 @@
+$(document).ready(function() {
+
+  var bowlingScorer = new BowlingScorer();
+
+  $("#score").text(bowlingScorer.returnScore());
+  // $("#frame_" + )
+
+  $("#add_score_0").click(function() {
```
jonurry: You have a massive amount of duplicated code in this file. It wouldn't take much effort to extract the duplication into a helper function and call it from each click function, passing in the number of the frame.

src/bowlingScorer.js
```diff
@@ -0,0 +1,68 @@
...
+
+BowlingScorer.prototype.returnFrame = function() {
+  return this.frame
+};
+
+BowlingScorer.prototype.returnFrameScore = function(frame) {
+  return this.scoreChart[frame-1]
+};
+
+BowlingScorer.prototype.returnBowlScore = function(score) {
```
jonurry: Not sure what the point of this function is? Why return the value that you pass in?

src/bowlingScorer.js
```diff
@@ -0,0 +1,68 @@
...
+  return score
+};
+
+BowlingScorer.prototype.isSpare = function(frame) {
```
jonurry: This is actually the logic for a strike.

src/bowlingScorer.js
```diff
@@ -0,0 +1,68 @@
...
+  return(this.scoreChart[frame-1] === 10 && this.roll === 1)
+};
+
+BowlingScorer.prototype.isStrike = function(frame) {
```
jonurry: This is actually the logic for a spare.

PR #837, opened 2018-11-12
Bowling challenge

Edy1988:
Score logic only.


camjw:
You asked about TDDing your code so I’m going to talk about that here and I'll leave some other comments on your code.

I looked at your commit history and I definitely like that you wrote the minimum amount of code necessary to pass the tests, but I think that while you’ve stuck to the ‘Red/Green’ loop you haven’t included the ‘Refactor’ bit. Your score method is quite long and does a lot of stuff. Having a nested 'if' statement is a sign that your method is doing too much, and you have a nested if nested in a for loop. Think about how you would refactor this into a series of methods doing stuff like: checking for a spare, checking for a strike, recalculating frame scores etc. 

In terms of your tests, it’s great that you’ve written a lot of tests to cover your code base but some of them are redundant - try using istanbul or another code coverage too keep track of how covered your code is and to check if you could refactor. This isn’t always obvious (to me) when running tests in a SpecRunner so I prefer using node.js testing frameworks (this is probably personal preference, but there are good tools for node.js). Also, try to test behaviour rather than state. This means testing how your methods behave rather than how your instances are initialised (i.e. you don’t need to tests that the score starts at 0, you test this through your methods). It might be a useful exercise to see how many of your tests you need to keep but maintain the same coverage.

Overall I would say this repo shows good evidence of a strong TDD process. I think this repo would benefit from more refactoring and more of an emphasis on DRY and SOLID principles (that score function has too many jobs!).

Please let me know if there’s anything you want further clarity on, or if you think anything I’ve said is unfair.

PR #846, opened 2018-11-23
Chris Cooksley Bowling Challenge for Review

Learner5200:


lib/js/frame.js
```diff
@@ -0,0 +1,72 @@
...
+
+Frame.prototype._expendBonuses = function(value) {
+  this.game.frames.forEach(function(frame) {
+    frame.expendBonus(value);
+  });
+};
+
+Frame.prototype._setBonus = function() {
+  if (this.rolls == 1) this.bonusRolls += 2; // strike
+  else this.bonusRolls += 1; // spare
```
sjmog: These two comments suggest that there's information you're neglecting to put directly into your code. Why not make them more explicit, using a variable, or a constant (a variable that never changes)?

lib/js/frame.js
```diff
@@ -0,0 +1,72 @@
...
+  return this.rolls >= 2 || this.total >= 10;
+};
+
+Frame.prototype._tenthFrameComplete = function() {
+  if (this.rolls >= 3) return true;
+  if (this.rolls == 2 && this.bonusRolls == 0) return true;
+  return false;
+};
+
+module.exports = Frame;
```
sjmog: + The use of private methods for readability here – it makes the Frame much easier to understand. 
+ The way your conditionals read. You put the simple case first, making it easy to understand, and make good use of JavaScript's sugary syntax to omit unnecessary brackets. It's easy to read.

- I wonder if some of your conditionals could be extracted to statements. Something like

```javascript
if (this.rolls >=3) return true;
```

Could have the conditional (the bit in the brackets) extracted like this:

```javascript
var finishedRolling = this.rolls >=3 || (this.rolls == 2 && this.bonusRolls == 0)

if(finishedRolling) return true;
```

And finally, don't forget that whenever you find yourself writing `if(condition) true else false`, you can just `return condition`:

```javascript
var finishedRolling = this.rolls >=3 || (this.rolls == 2 && this.bonusRolls == 0)

return finishedRolling;
```

So, for an even simpler method overall, line 66 could be:

```javascript
Frame.prototype._tenthFrameComplete = function() {
 return this.rolls >=3 || (this.rolls == 2 && this.bonusRolls == 0)
};
```

lib/js/script.js
```diff
@@ -0,0 +1,49 @@
...
+  var value = Number($("#score-input").val());
+  if (game.currentFrame().accepts(value)) {
+    game.roll(value);
+    $("#value-error").hide();
+  } else {
+    $("#value-error").show();
+  }
+  updateDisplay(game);
+};
+
```
sjmog: Great structure. Good to have all your scripts up-front, then the bit that uses those scripts lower down.

spec/frameSpec.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+    it("returns false if value greater than 10", function() {
+      expect(frame.accepts(11)).toEqual(false);
+    });
+    it("returns false if value less than 0", function() {
+      expect(frame.accepts(-1)).toEqual(false);
+    });
+    it("returns false if value is not an integer", function() {
+      expect(frame.accepts("string")).toEqual(false);
+    });
```
sjmog: This is a nice form of defensive type-checking. It might interest you to know that some languages introduce the idea of Types, which you can use to do this sort of thing more formally. For instance, Elm would allow you to write a `Score` type with valid values, that you could then use in the `Frame` instead of trying to exhaustively check across all sorts of things. Look into it!

sjmog (commented):
All in all, this is a good solution. There's loads of attention to code cleanliness and changeability, and a solid approach to test-driven development. It's clear that you've thought about how to build this system up from its basic units test-first, even where that's sometimes gone a bit awry (for instance, I'm still tickled by the Frame and Game dependency).

It's also quite clear that you're getting there with refactoring. I'd really recommend reading [What to Tidy](https://medium.com/@kentbeck_7670/what-to-tidy-28cb46e55009) by Kent Beck for a really simple once-over approach to refactoring easy things. It'll get you closer to being able to refactor anything.

Thanks for the submission! It was fun to read :)

PR #856, opened 2018-12-09
Bowling challenge. PR 1

MichaelSlevin:
Fully functional. I'm happy with the logic, but code could do with some refactoring to make it read more easily.

I need to separate out the concerns, perhaps create a class that handles the formatting of the scores when they are displayed.

I have struggled to get the background video to show in full screen.


src/frame.js
```diff
@@ -2,30 +2,50 @@ function Frame(frameNumber) {
...
+    this.bowlsToDouble = 3 - this.numberOfBowls();
   }
-  if (this.remainingPins == 0) {
-    this.bowlsToDouble = 3 - this.numberOfBowls()
-    console.log("bowls to double = "+this.bowlsToDouble)
+  if(this.frameComplete()) { this.complete = true; }
+};
+
+Frame.prototype.frameComplete = function() {
+  var finalFrameOver = this.frameNumber == 10 && this.bowlsToDouble == 0  && this.numberOfBowls() > 0;
```
AgaDufrat: Nice. I like the way of avoiding lengthy if statement

index.html
```diff
@@ -0,0 +1,125 @@
...
+    updateScoreCard()
+  });
+
+
+
+</script>
+<script async src="https://www.youtube.com/iframe_api"></script>
+<script>
+
+ // Written by @labnol
```
alicelieutier: ? 🤔

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
alicelieutier: No need for this file.

spec/features/bowlingSpec.js
```diff
@@ -0,0 +1,115 @@
...
+    game.bowl(10)
+    game.bowl(10)
+    expect(game.frames[0].score).toEqual(30)
+    expect(game.frames[1].score).toEqual(20)
+    expect(game.frames[2].score).toEqual(10)
+    expect(game.score[2]).toEqual(60)
+  });
+
+  it("12 strikes", function() {
+    game.bowl(10)
```
alicelieutier: You could do this in a loop. Or, maybe, have a helper function that does something like:
```js
function bowlSeveralTimes(rolls) {
  for (var i = 0; i < rolls.length; i++) {
    game.bowl(rolls[i]);
  }
}
```
and then
```js
bowlSeveralTimes([10, 10, 10, 10, ...]);
```

src/frame.js
alicelieutier: So do I, but I'd prefer splitting that long line to see all the conditions better
```js
var finalFrameOver = (
  this.frameNumber == 10 &&
  this.bowlsToDouble == 0  &&
  this.numberOfBowls() > 0
);
```

src/game.js
```diff
@@ -0,0 +1,65 @@
...
+Game.prototype.framesToDouble = function() {
+  return this.frames.filter(function(x) {
+    return (x.bowlsToDouble != 0);
+  });
+}
+
+Game.prototype.compileScores = function() {
+  for (var i = 0; i < this.frames.length; i++) {
+    this.individualScores.push(this.frames[i].bowls);
+    //change strikes and spares to symbols. Output an array of arrays
```
alicelieutier: symbols? I'd love to hear the idea behind this comment.

PR #860, opened 2018-12-10
Logic complete, not transformed into working html just yet

AsadK47:


MHUS25:
Code review feedback:

- Clear and readable code with good methods names. I agree that the code is easy to change and functions perform single tasks
- Good refactoring of the score function with helper methods. Commits show good evidence for “I can refactor anything”
- Good test coverage and well written tests

PR #861, opened 2018-12-10
Bowling challenge

AgaDufrat:
**Features**
* calculates scores without bonuses
* adds bonus for strike (10 pins knocked in first roll)
* adds bonus for a spare (10 pins knocked with the two rolls of a frame)
* calculates total score
* remove drop down list at the end of the Game
* hide select options based on number of pins knocked
* show select options at the beginning of new frame

**To do**
* fix last frame
* refactoring


MichaelSlevin:
6/7 tests seem to be failing
    - [x]  Gutter game
    - [x]  One frame
    - [x]  Multiple frames
    - [x]  Spares
    - [ ]  Strikes
    - [ ] Final Frame

Removal of impossible scores is good!
Spares are working.
Strikes are not working if multiple strikes are bowled
Final frame is not working

PR #865, opened 2018-12-10
Total Game Score added

RobertsK284:
Still need to test some edge cases. Having trouble with try throw catch.


RobertsK284:
Hi Alice. Would you please give feedback?

spec/gameSpec.js
```diff
@@ -0,0 +1,54 @@
...
+    game = new Game();
+  });
+
+  it('can add a frame to a game', function() {
+    game.addFrame(3, 5);
+    expect(game.frames.length).toEqual(1);
+  });
+
+  it('sets the maximum number of normal frames to 9', function () {
+    var i = 0;
```
alicelieutier: unused variable i

src/frame.js
```diff
@@ -0,0 +1,73 @@
...
+Frame.prototype.getAllRolls = function () {
+  return this.rolls;
+};
+
+Frame.prototype._rollSum = function () {
+  if (this.rolls[2] == undefined) {
+    return this.rolls[0] + this.rolls[1];
+
+  } else {
+      return this.rolls.reduce(function (total, roll) {
```
alicelieutier: Nice use of reduce.
You can also define a helper function called sum
```js
function sum(array){
    return array.reduce((a, b) => a + b)
}
```
and then call `sum(this.rolls)`

PR #868, opened 2018-12-10
Bowling challenge

Obsiye:
Created a bowling score card/game. I came up with the wrong solution I believe. It should have just kept a score and not actually scored.

I have Unit tests with Jasmine. No eslint yet. No 100% test coverage just yet.


.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    "extends": "airbnb-base"
```
alicelieutier: ? 🤔

PR #869, opened 2018-12-10
Bowling challenge submission

willmcb:
My submission - see readme on how to get it running.

There are a few bits I didn't get to:
- validation of numbers you enter - it will calculate scores correctly as long as you enter scores that are compatible with bowling i.e. 1-10. However, there is no validation to stop you entering something like 234.
- haven't run ESlint.
- need to refactor code to make it dryer especially in the game class and gameSpec.

I think if I was to do the challenge again I would most definitely try to think of a simpler architecture, I think I have over engineered parts of my solution and it has made it difficult to add to. Though it has fairly comprehensive tests so this has given me some safety net for refactoring.


main.js
```diff
@@ -0,0 +1,77 @@
...
+     return 'X';
+   } else if (frame.isSpare()) {
+     return '/';
+   } else {
+     return deNullify(frame.getSecondScore());
+   }
+}
+
+function deNullify(score) {
+   if (score == 0) {
```
alicelieutier: Should you be using `===` here?

spec/gameSpec.js
```diff
@@ -0,0 +1,122 @@
...
+    it("its last frame should be of type 'tenthFrame'", function() {
+      expect(game.frames.pop()).toEqual(jasmine.any(TenthFrame));
+    });
+
+    it("can return a given frame", function(){
+      expect(game.getFrame(9)).toEqual(jasmine.any(TenthFrame));
+    });
+
+    it("can correctly update frames with an input score method", function(){
+      game.inputScore(10); // 1
```
alicelieutier: You should probably have a helper function here that does something like:
```js
function addMultipleRolls(rolls) {
    for (var i=0; i < rolls.length; i++) {
        game.inputScore(rolls[i]);
    }
}
```

spec/updaterSpec.js
```diff
@@ -0,0 +1,167 @@
+var Updater = require('../src/updater.js');
+var Frame = require('../src/frame.js');
+var TenthFrame = require('../src/tenthFrame.js');
+
+describe("Updater", function() {
+    var perfectRoundUpToRoundNine = [[10, 0],[10, 0], [10, 0],
```
alicelieutier: Nice!

spec/updaterSpec.js
```diff
@@ -0,0 +1,167 @@
...
+
+    var regularRoundUpToRoundNineTwo = [[10, 0],[1, 3], [3, 7],
+                                        [5, 0], [10, 0], [6, 3],
+                                        [0, 6], [1, 8], [3, 3]];
+    var partialGame = regularRoundUpToRoundNineTwo.slice(0,6);
+
+    beforeEach(function() {
+      updater = new Updater();
+      frames = [];
+      [1, 2, 3, 4, 5, 6, 7, 8, 9].forEach(function(){
```
alicelieutier: This is a very "ruby" way to do this. Javascript would probably use a for loop here instead:
```js
for (var i=0; i < 9; i++) {
    frames.push(new Frame());
}
```

src/frame.js
```diff
@@ -0,0 +1,67 @@
...
+    if (score == this.STRIKE_SCORE) {
+        this.strike = true;
+    }
+    this.firstScore = score;
+}
+
+Frame.prototype.setSecondScore = function(score){
+    if (this.isStrike()) {
+        return null;
+    } else {
```
alicelieutier: you don't need an else after a return.
```js
if (this.isStrike()) {
        return null;
if((score + this.getFirstScore()) === this.STRIKE_SCORE) {
    this.spare = true;
}
this.finalised = true;
this.secondScore = score;
this.setFinalFrameScore(this.getFirstScore() + this.getSecondScore());
```

src/frame.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+Frame.prototype.getSecondScore = function(score){
+    return this.secondScore;
+}
+
+Frame.prototype.getFinalFrameScore = function() {
+    if (!this.isFinalised()) {
+        return null;
+    } else {
```
alicelieutier: same

src/game.js
```diff
@@ -0,0 +1,97 @@
...
+
+Game.prototype.addFrame = function(frame){
+  this.frames.push(frame);
+}
+
+Game.prototype.update = function() {
+  this.updater.update(this.frames);
+}
+
+Game.prototype.inputScore = function(score){
```
alicelieutier: I'm sure you know that this is too long for a function. You may want to extract part of it in an other function.
Maybe everything inside the condition `if (this.increment == 9)`

src/game.js
```diff
@@ -0,0 +1,97 @@
...
+
+   if (currentFrame.isStrike()) {
+       this.increment += 1;
+       currentFrame = this.getFrame(this.increment);
+       currentFrame.setFirstScore(score);
+       this.update();
+       return;
+   }
+
+   if (!currentFrame.isStrike()) {
```
alicelieutier: Is this test necessary? What is the default case?

src/updater.js
```diff
@@ -0,0 +1,65 @@
...
+    var next = frames[index + 1];
+    var nextNext = frames[index + 2];
+    if (frame.isStrike()) {
+       that.updateStrike(frame, next, nextNext, index);
+    } else if (frame.isSpare()) {
+       that.updateSpare(frame, next);
+    }
+}
+
+Updater.prototype.updateStrike = function(frame, next, nextNext, index) {
```
alicelieutier: it's really hard for me to guess what this function does from the name and argument names alone.

main.js
willmcb: Yes probably! I think it makes sense to use `===` by default.

spec/gameSpec.js
willmcb: Yep, I had used a few helper functions similar to this elsewhere - I just didn't get around to refactoring it. I think sometimes it is good to 'get it working' and then refactor (which backfires when you run out of time :-(.

spec/updaterSpec.js
willmcb: I actually googled something like `ruby .each in javascript` and it came back with this. I'll have to swap it out for the JS way, as this way is pretty ugly.

src/frame.js
willmcb: hmm interesting, I never knew that. thanks!

src/game.js
willmcb: Yep, I need to refactor this out - I had trouble getting the logic right so I concentrated on that to begin with and didn't get around to refactoring.

src/game.js
willmcb: Probably not - I'll remove it and rerun my tests.

PR #871, opened 2018-12-10
Pull request

DGajewska:


firstAttempt/spec/MockGame-NoBonusStreak.js
```diff
@@ -0,0 +1,93 @@
+describe('Mock game - no bonus streaks, 2 bowls in frame 10', function(){
+  //Play full game of bolwing
+  bowlingCard = new BowlingScorecard();
+  bowlingCard.enterBowl(10);
```
alicelieutier: You may want to write a helper function like:
```js
function addMultipleRolls(rolls) {
    for (var i=0; i < rolls.length; i++) {
        bowlingCard.enterBowl(rolls[i]);
    }
}
```

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
alicelieutier: Unsure what this file is supposed to be 🤔

PR #876, opened 2018-12-10
Bowling Challenge

Amac01:
I would like a code review please. 
Thanks


spec/PlayerSpec.js
```diff
@@ -0,0 +1,58 @@
+describe("Player", function() {
```
alicelieutier: This does not feel like it comes from this project. If you used to have an example, you should delete before sending a PR.

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
alicelieutier: Same

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,28 @@
...
+    game = new BowlingGame();
+  });
+
+  //player never hits a pin (20 zero scores)
+  it('can roll a gutter game', function() {
+    rollMore(0, 10);
+    expect(game.score()).toBe(0);
+  });
+
+  it('can roll ones', function() {
```
alicelieutier: If you roll a one ten times, you are probably still on frame 5

src/Player.js
```diff
@@ -0,0 +1,22 @@
+function Player() {
```
alicelieutier: Same as above - unrelated file

PR #883, opened 2019-01-20
Weekend challenge partially completed

scass91:
Hello! Jasmine tests pass, for some reason when I try to open this in an index.html file it doesn't function as desired within this console, any feedback on that, as well as on a methodical approach to problem solving would be appreciated. Thank you.


jasmine/src/BowlingGame.js
```diff
@@ -0,0 +1,52 @@
+"use strict";
```
AlinaGoaga: Very easy to read code :) Thanks for making my life easier! I like you're using strict as well, shows commitment to writing good syntax

AlinaGoaga:
Nice history, shows you're committing frequently and are using TDD to build your code!

PR #887, opened 2019-01-20
Bowling Weekend Challenge - Alina

AlinaGoaga:


spec/GameSpec.js
```diff
@@ -7,25 +7,43 @@ describe('Game', function() {
     expect(game.overallScore).toEqual(0);
   });
 
-  it('generates all the frames', function() {
+  it('generates all the frames when there are no strikes or spares at the end', function() {
     expect(game.loadFrames([3, 5, 10, 5, 5, 3, 5, 6, 2, 2, 3, 5, 4, 7, 1, 7, 1, 6, 2]).length).toEqual(10);
   });
```
scass91: Consider putting these in a describe block with a before element before the 3 tests which could represent the first 9 frames of each scenario

scass91:
There's a lot of evidence of red-green-refactor here so you can say you've followed a process well!

spec/GameSpec.js
scass91: To clarify, you could create a function:  

	var rollX = function(pins, rolls) {
		for (var i = 0; i < rolls; i++)
			game.roll(pins);
	};
Then before each scenario do rollX(1,18) to roll 18 individual 1's, leaving your final frame clean for manipulation with a strike or spare

PR #889, opened 2019-01-21
Celine Kaslin Bowling-challenge November 2018

CelineKaslin:


README.md
```diff
@@ -1,77 +1,97 @@
+# Bowling Challenge #
 
-Bowling Challenge
-=================
+![bowling ASCII](https://user-images.githubusercontent.com/43742795/51447391-9ef6bd80-1d15-11e9-9878-96abb0d39038.png)
```
joaoag: Very nice, good content in the readme and visually pleasing too! 💯

spec/scorecardSpec.js
```diff
@@ -0,0 +1,64 @@
+describe ('ScoreCard', function(){
+  let scoreCard;
+
+  beforeEach(function() {
+    scoreCard = new ScoreCard();
+  });
+
+  describe ("Gutter Game", function(){
```
joaoag: Nice having really explicit descriptions here

spec/scorecardSpec.js
```diff
@@ -0,0 +1,64 @@
...
+    it('the player should have a score of 0 at the end of a game of 20 frame where they did not knocked any pins', function(){
+      for (let s = 0; s < 20; s++){
```
joaoag: neat loop

src/scorecard.js
```diff
@@ -0,0 +1,28 @@
...
+        r += 1;
+      } else if (this.rolls[r] + this.rolls[r+1] === 10){
+        s += this.rolls[r] + this.rolls[r+1] + this.rolls[r+2]
+        r += 2
+      } else {
+        s += this.rolls[r] + this.rolls[r+1]
+        r += 2
+      };
+    };
+    return s;
```
joaoag: Great work implementing the logic for strikes and spares, I didn't get anywhere close. As discussed given more time this would be reformatted into multiple methods to adhere to SRP. I guess first step would be to ship the code executed in each of the if and else blocks into separate methods, so it would read like 'if (condition) { return methodName }'

src/scorecard.js
```diff
@@ -0,0 +1,28 @@
+class ScoreCard {
```
joaoag: With more time this could be split up into multiple classes. For example maybe you could create a Score class that's responsible for receiving and passing on user scores to ScoreCard.

spec/scorecardSpec.js
```diff
@@ -0,0 +1,64 @@
...
+    it('player play a perfect game with a score of 300', function(){
+      for (let s = 0; s < 10; s++){
+        scoreCard.roll(10)};
+      scoreCard.roll(10);
+      scoreCard.roll(10);
+      expect(scoreCard.score()).toEqual(300);
+    });
+  });
+
+});
```
joaoag: Great coverage on these, can see you've got the key situations covered in these Feature tests. Would you consider creating unit tests for Class methods / properties as well?

PR #894, opened 2019-01-21
Bowling Challenge - Jethro Whaley

jaywayawyaj:


jaywayawyaj:
UPDATE - All tests are passing but functionality isn't there fully, the scoring gets confused with a truly mixed game. Can still roll a perfect game though so you just gonna have to up them skills.

PR #943, opened 2019-03-11
Almost finished back end

kate102:
Had it all working but my naming was horrible and so I refactored it. I wasn't happy with the algorithm  for the perfect score so I got rid of that and now it doesn't work. Will get back to that and the front end next week.


src/Bowling.js
```diff
@@ -0,0 +1,120 @@
...
+  this.gameFrames.push(this.currentFrame);
+  if (!this.gameOver)
+  {
+    this.checkGameOverStatus()}
+    this.currentFrame = new Frame()
+    this.firstTryOfFrame = true
+    this.updateScore()
+};
+
+Bowling.prototype.updateScore = function() {
```
Brendao1: This could be refactored to call strikeScore and spareScore.

src/Bowling.js
```diff
@@ -0,0 +1,120 @@
+function Frame(){
+  this.firstScore = 0
+  this.secondScore = 0
+  this.frameStatus = 'Open'
+}
+Frame.prototype.setStatus = function()
```
Brendao1: This is a method called on the Frame class.

PR #945, opened 2019-03-18
Bowling

SurekaStoeckigt:


spec/FrameSpec.js
```diff
@@ -0,0 +1,161 @@
+describe('Frame', function(){
+
+  beforeEach(function(){
+    frame = new Frame(false);
+  });
+
+  it('starts with a total score of 0 for first bowl of frame', function(){
```
alicelieutier: These two first tests, and most of the other ones do not need to exist. They are testing *state* (instance variables).
As a rule, what you want to test are *behaviours*, which usually mean testing externally accessible methods (public methods) of the class.

Can you guess why that is?

spec/FrameSpec.js
```diff
@@ -0,0 +1,161 @@
...
+  it('does not give bonus if no strike or spare is rolled', function(){
+    frame.roll(2);
+    frame.roll(2);
+    expect(frame._done).toEqual(true);
+    expect(frame._spare).toEqual(false);
+    expect(frame._strike).toEqual(false);
+    expect(frame._bonusScore).toEqual(0);
+  });
+
+  it('does not record a bonus if strike is rolled and only added in the game class', function(){
```
alicelieutier: I don't understand what this test is testing for.
If a test has a "code" oriented title and not a "user" oriented title, then it is probably not a needed test.

spec/GameSpec.js
```diff
@@ -0,0 +1,236 @@
...
+  });
+
+  it('knows which frame it is in after multiple rolls', function(){
+    game.roll(4);
+    game.roll(1);
+    game.roll(4);
+    game.roll(1);
+    game.roll(4);
+    game.roll(1);
+    game.getTotalScore();
```
alicelieutier: Why do you need to call this method?

spec/GameSpec.js
```diff
@@ -0,0 +1,236 @@
...
+
+  it('does not add the bonus score for a strike if the following frame has not ended', function(){
+    game.roll(10);
+    game.roll(3);
+    game.roll(1);
+    game.roll(1);
+    expect(game.getTotalScore()).not.toEqual(21);
+  });
+
+<<<<<<< HEAD
```
alicelieutier: It looks like you marked a merge conflict as resolved when it wasn't actually resolved. This type of line should never appear in your committed files.

spec/GameSpec.js
```diff
@@ -0,0 +1,236 @@
...
+=======
+it('adds next bowl to bonus score if spare is bowled', function(){
+  game.frames[0].bowl(3);
+  game.frames[0].bowl(7);
+  game.frames[1].bowl(1);
+  game.frames[1].bowl(0);
+  game.bonusScores();
+  expect(game.bonusScore).toEqual(1);
+});
+>>>>>>> 5924be6834240ba6f4558b7950c8028c2bbff579
```
alicelieutier: Same as above

src/Frame.js
```diff
@@ -0,0 +1,49 @@
+function Frame(finalFrame) {
+this._frameRoll1Score = 0;
+this._frameRoll2Score = 0;
+this._frameRoll3Score = 0;
+this._bonusScore = 0;
+this._strike = false;
+this._done = false;
+this._spare = false;
+this._finalFrame = finalFrame;
+this._rollNumber = 0;
```
alicelieutier: You are not closing the constructor here, but instead defining all methods inside the constructor. This is not how the constructor/prototype pattern is supposed to work, so I'm surprised it actually does work... 🤔

src/Frame.js
```diff
@@ -0,0 +1,49 @@
+function Frame(finalFrame) {
```
alicelieutier: The name for this parameter should make it obvious it expects a boolean. Something like `isFinalFrame` would be nice.

src/Frame.js
```diff
@@ -0,0 +1,49 @@
...
+this._rollNumber = 0;
+
+Frame.prototype.totalScoreForFrame = function(){
+  if (this._done) {
+    return this._frameRoll1Score + this._frameRoll2Score + this._frameRoll3Score + this._bonusScore;
+  }
+  return 0;
+};
+
+Frame.prototype.roll = function(pins_hit) {
```
alicelieutier: This method is quite long, with many if/else's. I'm sure you can simplify it further.

alicelieutier: One way to do that may be private methods (actually, JS does not have private methods, but by convention, an method name starting with `_` will be considered private).

For example, this method could be:
```
Frame.prototype.roll = function(pins_hit) {
    if (this._done) return;

    if (this._rollNumber === 0) {
    	this._addFirstRoll(pins_hit);
    } else if (this._rollNumber === 1) {
    	this._addSecondRoll(pins_hit);
    } else {
      this._frameRoll3Score = pins_hit;
      this._done = true;
    };
}
```

And then the rest could be separated into the private methods `_addFirstRoll` and `_addSecondRoll`

src/Game.js
```diff
@@ -0,0 +1,93 @@
+function Game() {
+  this._frames = [];
+  this._currentFrameNumber = 1;
+  this._numberOfFrames = 10;
+  this._frames.push(new Frame(false))
+
+<<<<<<< HEAD
```
alicelieutier: Similar comment to above. This line indicates that a merge conflict was not properly resolved.

src/Game.js
```diff
@@ -0,0 +1,93 @@
...
+    perfect_game = (perfect_game) && (this._frames[i]._strike);
+  };
+
+  perfect_game = perfect_game && (this._frames[this._frames.length-1].totalScoreForFrame() === 30)
+
+  if (perfect_game && this._frames.length >= 10) {
+   totalScoreForGame = 300;
+  };
+ return totalScoreForGame
+=======
```
alicelieutier: The function above is not closed.

PR #948, opened 2019-03-18
partial solution

Brendao1:


spec/BowlingSpec.js
```diff
@@ -0,0 +1,55 @@
...
+      });
+
+      it('can score a whole game of ones', function () {
+        rollMany(1, 20);
+        expect(bowling.score()).toEqual(20);
+      });
+
+      // eslint-disable-next-line no-var
+      // eslint-disable-next-line vars-on-top
+      var rollMany = function (pinsDown, rolls) {
```
kate102: I like the inline function but perhaps it could have been used in the previous test too.

PR #949, opened 2019-03-18
Some bowling stuff

Schofield88:
Funny old game, isn't it?


spec/frameSpec.js
```diff
@@ -0,0 +1,56 @@
+describe('Frame', function(){
+
+  var frame
+
+  describe('Initial setup state', function(){
```
acodeguy: I like your use of nested describe blocks

PR #957, opened 2019-04-08
Bart bowling challenge v1

fo0s:


fo0s:
Update to v1. Added frontend, tried to tweak algorithm. Sort of WIP.

PR #958, opened 2019-04-08
Incomplete

JRRS1982:
As per commit messages. Please see last commit for passing tests and code with a few features that work nicely incrementing up frames if not a strike etc, i then started refactoring and extracted roll and frame, and it started to fail tests... not sure how to link the files or combine the functions currently. Requires more work. Feel like it is neat and am enjoying javascript, but have only studied it for a week. Satisfied with progress, but want to do more!


WardahTarar:
- Good work with having different functions for different purposes.
- Well defined names for properties and functions
- Making a rolls generator wasn't a requirement for the challenge, but good to see that you have done it, would be good to add a test for it as well.
- Having a more descriptive name for the 10th frame function would be great (instead of just Frame)
- Would be a good practice to use ternary conditional boolean or to make another function, instead of using a long if/else statement

PR #966, opened 2019-04-15
Bowling Challenge - unfinished - Arthur

arthurashman:


src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
+function Scorecard() {
+  this.frames = [];
+  this.score = 0;
+};
+
+Scorecard.prototype.enter = function(roll1, roll2) {
+
```
amyj0rdan: From user point of view, user might want to enter one roll at a time (eg, if you had a click button of the number of pins). Maybe change to enter one roll at a time?

src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
+function Scorecard() {
+  this.frames = [];
+  this.score = 0;
+};
+
+Scorecard.prototype.enter = function(roll1, roll2) {
+  
+    if(roll1 + roll2 > 10){
+      throw Error()
```
amyj0rdan: Nice use of error-ing

src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
...
+Scorecard.prototype.calculate = function() {
+  var calculator = []
+  if ((this.frames).every(isStrike)) {
+    return 300
+  }
+  else {
+    this.frames.forEach(function(frame) {
+     calculator.push(frame.firstRoll +  frame.secondRoll)
+    })
+    const reducer = (accumulator, currentValue) => accumulator + currentValue;
```
amyj0rdan: Lovely bit of syntax here.

src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+ else{
+  this.frames.push({firstRoll: roll1, secondRoll: roll2})
+
+ }
+};
+
+Scorecard.prototype.calculate = function() {
+  var calculator = []
+  if ((this.frames).every(isStrike)) {
```
amyj0rdan: Again, like this 'every' syntax.

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,55 @@
+describe('Scorecard', function() {
+  var scorecard;
+  var frame;
+
+  beforeEach(function() {
+    scorecard = new Scorecard();
+    frame = new Frame();
```
amyj0rdan: You don't use this yet so maybe delete until you do?

src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
...
+    if(roll1 + roll2 > 10){
+      throw Error()
+    }
+ else{
+  this.frames.push({firstRoll: roll1, secondRoll: roll2})
+
+ }
+};
+
+Scorecard.prototype.calculate = function() {
```
amyj0rdan: According to rubric, good practice to have one function that calculates and one method that returns the pre-calculated score (eg, calculateScore - does all the calculating - and score - that returns just the score)

src/Frame.js
```diff
@@ -0,0 +1,3 @@
+function Frame() {
+
+};
```
hybridbad: Whats this going to be? Seems like you already defined what a frame is in the scorecard js

hybridbad:
Commits have clear messages but could you have done more commits, with references to tests etc

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,55 @@
...
+
+    it('ensures rolls do not sum to over 10', function() {
+      expect( function(){scorecard.enter(3,8); } ).toThrow( new Error, ('Check entry'));
+    }); 
+  });
+
+
+  describe('#calculate', function() {
+    it('scores 0 for a gutter game', function() {
+      for (var i = 0; i < 10; i++) {
```
hybridbad: LOOPS whey

src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
...
+    this.frames.forEach(function(frame) {
+     calculator.push(frame.firstRoll +  frame.secondRoll)
+    })
+    const reducer = (accumulator, currentValue) => accumulator + currentValue;
+    this.score = calculator.reduce(reducer)
+    return this.score;
+  };
+};
+
+function isStrike(frame) {
```
hybridbad: nice predicate function to delegate responsibilities

src/Scorecard.js
```diff
@@ -0,0 +1,36 @@
+function Scorecard() {
```
hybridbad: Did you try installing eslint? It helps a lot with them semi colons, whether you needed it or not

amyj0rdan:
- Step 0: All green tests. Lovely.
- Step 1: If you're going to revisit, maybe build up with the functionality in the order on the code rubric (think that perfect score in the project description is a bit of a ruse in terms of ease of implementation, but that might just be cos I got stuck on it):
* Gutter game
* One frame
* Multiple frames
* Spares
* Strikes
* Final Frame
* UI
- Step 2: README has been updated with project specific instructions 👍 You could add a bit more about your approach and future plans for the project if you had more time.
- Step 3: your tests are very clean and easy to follow and you have tested some edge cases 👍 You have a frame spec and src file even though those aren't set up yet, which is a little confusing so maybe get rid of those until you start implementing them. 
- Step 4: 
* I added a comment in the file itself, but (code review advised) good practice is to have one method that calculates the score and one that returns it.
* Good SRP for isStrike function. Might want to extract that out to the frame class when you implement that. Could you extract out the reducer to elsewhere if you think you might use it again?

PR #967, opened 2019-04-15
Paul's bowling challenge

Hives:


index.html
```diff
@@ -0,0 +1,294 @@
...
+          $(".error").stop().fadeIn(0, function() {
+            $(this).fadeOut(4000)
+            return false
+          })
+        }
+
+        return false
+      })
+    })
+  </script>
```
Hives: this is carnage, you should be ashamed of yourself

src/Scorer.js
```diff
@@ -0,0 +1,93 @@
...
+  total: function () {
+    return this._lastRunningTotal()
+  },
+
+  _tryUpdatingScores: function () {
+    var i = this._scoringIndex
+    if (this._scoredFrames() === 10) {
+      return false
+    }
+    if (this._rolls[i] === 10) {
```
Hives: would it be better to have a frame object here which could store whether it's a strike or spare? so you would do `if (this.isStrike()) {}` or something? problem i had with that was with wanting to be able to store the rolls in an array so i could iterate over them easily, and not wanting to store them in the frame as well... so you could have them stored in an array in `scorer`, and then... store an index reference to the appropriate rolls in the `frame`? thoughts...

src/Scorer.js
```diff
@@ -0,0 +1,93 @@
...
+
+  _scoreFrame: function (score) {
+    this.runningTotals.push(this._lastRunningTotal() + score)
+  },
+
+  _scoredFrames: function () {
+    return this.runningTotals.length
+  },
+
+  _lastRunningTotal: function () {
```
Hives: consider adding a `.last()` method to `Array.prototype`

src/Game.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  this._scorer = scorer || new Scorer()
+}
+
+Game.prototype = {
+  constructor: Game,
+
+  roll: function (roll) {
+    this._scorer.addRoll(roll)
```
riyadattani: .addRoll is from the Score object - circular dependancy. From the workshop I think each object is not supposed to depend on other objects as much as possible (I think this was actually your question).

PR #968, opened 2019-04-15
Really bad bowling challenge :( 

LaurenQurashi:


spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+
+  var bowl;
+  // similar to an instance variable initialization.
+
+  beforeEach(function() {
+    bowl = new Bowl(0,0,0,0);
+    // creates a new object before each test.
+  });
+
+  it('The new bowl has a Score which starts at 0', function() {
```
m-budryte: good for setting up, however, this test is checking the **state** (attribute) rather than **behaviour**.

testing behaviour: setting up a function called **'currentScore()'** which calls the attribute _score_ . Hope it makes sense!

spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(bowl.score).toEqual(0);
+  });
+
+  it('The new bowl has a Roll1 which starts at 0', function() {
```
m-budryte: see above

spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(bowl.roll1).toEqual(0);
+  });
+
+  it('The new bowl has a Roll2 which starts at 0', function() {
```
m-budryte: see above

spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(bowl.roll2).toEqual(0);
+  });
+
+  it('The new bowl has a Frame which starts at 0', function() {
```
m-budryte: see above

spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(bowl.frame).toEqual(0);
+  });
+
+  it('Adds 1 to the frame counter', function() {
```
m-budryte: consider renaming 'add' into something more self-explanatory?

spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+    expect(bowl.frame).toEqual(1);
+  });
+
+  it('Adds the rolls to the score counter', function() {
+    bowl.add(5,5)
+    expect(bowl.score).toEqual(10);
+  });
+});
+
+describe('Bowl Functions', function() {
```
m-budryte: I like that you separated tests according to the functionalities. 
Consider renaming the first group 'Bowl setup" or something similar to make it clearer

spec/BowlSpec.js
```diff
@@ -0,0 +1,61 @@
...
+  var bowl;
+
+  it('Gives a gutterball score if no pins are hit', function() {
```
m-budryte: sensible use of a first test to make sure that other functions work properly - great job

src/Bowl.js
```diff
@@ -0,0 +1,27 @@
...
+  this.score = score;
+  this.bonus = []
+  this.frameTotal = 0
+};
+
+
+Bowl.prototype.add = function(roll1, roll2) {
+  this.score += (roll1 + roll2)
+  this.frame += 1
+  this.frameTotal += (roll1 + roll2)
```
m-budryte: consider extracting frame as a separate object to adhere to the Single Responsibility Principle

src/Bowl.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+Bowl.prototype.bonusCalc = function(roll1, roll2, frame, score) {
+  if (this.frame == 10 && this.score == 0) {
+    this.bonus.push("Gutter")
+  } else if (this.frameTotal == 10) {
+    this.bonus.push("Strike")
+    this.frameTotal = 0
+  } else if (this.bonus[-1] == "Strike") {
```
m-budryte: javascript doesn't know '-1', instead, use this.bonus[this.bonus.length - 1] or this.bonus.last

pls refer to http://www.codeblocq.com/2016/05/Get-the-last-element-of-an-Array-in-JavaScript/

m-budryte: in Javascript (because it's erm a quite peculiar language) it's always better to use a triple equal sign (===) rather than (==) 

https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a
https://bytearcher.com/articles/equality-comparison-operator-javascript/

PR #969, opened 2019-04-15
Partial Bowling Challenge - some business logic - no readme...

hybridbad:
Enjoy...


spec/scorcardSpec.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable */
+describe('Scorecard', function(){
```
arthurashman: All tests passing, and clearly described

spec/scorcardSpec.js
```diff
@@ -0,0 +1,76 @@
...
+    })
+
+    it('returns true on complete game', function(){
+      for(var i = 1; i < 11; i++){
+        scorecard.recordFrame(i, 0, 1)
+      }
+      expect(scorecard.isComplete()).toBe(true);
+    })
+
+    it('wont record more than 10 frames', function(){
```
arthurashman: Could be worth throwing an error too, so user knows the game has ended

spec/scorcardSpec.js
```diff
@@ -0,0 +1,76 @@
...
+      for(var i = 1; i < 11; i++){
+        scorecard.recordFrame(i, 0, 0)
+      }
+      expect(scorecard.isComplete()).toBe(true);
+      expect(scorecard.total()).toBe(0);
+    })
+
+  });
+
+  describe('new game', function(){
```
arthurashman: I like the implementation of this, so that the eventual interface would then allow for the next game to be set up

arthurashman:
Commit messages could be a little clearer
https://medium.com/@steveamaza/how-to-write-a-proper-git-commit-message-e028865e5791
The more defined each commit is, the more obvious it will be that development follows red, green, refactor to prove it was TDD'd

PR #986, opened 2019-05-31
Bowling Challenge

breycarr:
Task is to create a bowling scorecard programme in JavaScript


src/scorecard.js
```diff
@@ -0,0 +1,42 @@
...
+  this._frame = 1;
+  this._gameScore = [[]]
+}
+
+Scorecard.prototype.showScore = function() {
+  return this.sumFrameScore();
+}
+
+Scorecard.prototype.enterScore = function(pins) {
+  if (this.sumFrameScore() + pins > 10) throw "maximum input per frame is 10";
```
lbarber729: I like that you added an error if over 10, good thinking!

src/scorecard.js
```diff
@@ -0,0 +1,42 @@
...
+  if ((this._frameScore.length === 2) || (pins === 10)) {
+    this.frameReset();
+  }
+}
+
+Scorecard.prototype.isFrame = function() {
+  return this._frame;
+}
+
+Scorecard.prototype.frameReset = function() {
```
lbarber729: I like that you moved this out as a separate method.

src/scorecard.js
```diff
@@ -0,0 +1,42 @@
...
+  this._frameScore.push(pins);
+  if (this._gameScore[this._frame -1].includes(10)) {
+    this._gameScore[this._frame -1].push(pins);
+  }
+  if ((this._frameScore.length === 2) || (pins === 10)) {
+    this.frameReset();
+  }
+}
+
+Scorecard.prototype.isFrame = function() {
```
lbarber729: Do you need this as a separate method? is it used anywhere?

src/scorecard.js
breycarr: Thanks!

src/scorecard.js
breycarr: Good point! It was made to keep this._frame as a private method, but it probably doesn't need to be private in that instance.

src/scorecard.js
breycarr: Thanks!

PR #987, opened 2019-06-02
Bowling Challenge

EdinaBMakers:


js/frame.js
```diff
@@ -0,0 +1,83 @@
...
+      return this._firstRollScore === 10;
+   },
+
+   isSpare: function() {
+      return this._firstRollScore + this._secondRollScore === 10
+             && !this.isStrike();
+   },
+
+   isEnded: function() {
+      return this._isFinalFrame
```
alicelieutier: This method is difficult to understand. Can you make it more readable by extracting part of the logic, or using variables to give meaningful names?

spec/scoreCardSpec.js
```diff
@@ -0,0 +1,131 @@
...
+      function() {
+         scoreCard.recordScore(1);
+         scoreCard.recordScore(2);
+
+         expect(scoreCard.frames.length).toEqual(1);
+
+         scoreCard.recordScore(4);
+
+         expect(scoreCard.frames.length).toEqual(2);
+         expect(scoreCard.frames[1].getTotalScore()).toEqual(4);
```
alicelieutier: Is this tested in isolation? You may want to use doubles (or spies) here to not reply on Frame when testing ScoreCard.

alicelieutier: Or, maybe this test is trying to test a behaviour that belongs to Frame and not ScoreCard? If this is the case, then maybe both classes are too coupled. It would be interesting to see how the responsibilities are separated - maybe in a diagram.

spec/scoreCardSpec.js
```diff
@@ -0,0 +1,131 @@
...
+
+      it('adds score to final frame after the ninth frame ended', function() {
+         for(var i = 0; i < 9; i++) {
+            scoreCard.recordScore(10);
+         }
+
+         scoreCard.recordScore(1);
+
+         expect(scoreCard.frames.length).toEqual(10);
+         expect(scoreCard.frames[9]._isFinalFrame).toEqual(true);
```
alicelieutier: Same as above

PR #988, opened 2019-06-02
Bowling Challenge - not complete

lbarber729:


src/Bowling.js
```diff
@@ -0,0 +1,54 @@
+function Scorecard() {
+  this.score = 0;
```
breycarr: Is it worth adding underscores (eg. this._score) to mark these as private characteristics?

src/Bowling.js
```diff
@@ -0,0 +1,54 @@
...
+  this.frames = [];
+  this.currentFrame = [];
+  this.frameCounter = 0;
+};
+
+Scorecard.prototype.rolls = function (roll1, roll2) {
+  this.currentFrame = []
+  this.currentFrame.push(roll1, roll2)
+  this.score = this.score + (this.currentFrame.reduce((a, b) => a + b, 0));
```
breycarr: could you extract the reduce method to maintain single responsibility principle?

src/Bowling.js
```diff
@@ -0,0 +1,54 @@
...
+  this.frames.push(this.currentFrame);
+  this.frameCounter += 1;
+};
+
+Scorecard.prototype.isStrike = function () {
```
breycarr: Line 16 to  line 47 don't seem to be called anywhere, where did you intend to add these?

src/Bowling.js
```diff
@@ -0,0 +1,54 @@
+function Scorecard() {
+  this.score = 0;
+  this.frames = [];
+  this.currentFrame = [];
+  this.frameCounter = 0;
+};
+
+Scorecard.prototype.rolls = function (roll1, roll2) {
+  this.currentFrame = []
```
breycarr: Do you need to add if statements to account for different scores and outcomes (strikes, etc)?

src/Bowling.js
```diff
@@ -0,0 +1,54 @@
...
+  this.currentFrame.push(roll1, roll2)
```
breycarr: what happens if a player adds rolls that add up to more than 10?

src/Bowling.js
```diff
@@ -0,0 +1,54 @@
...
+Scorecard.prototype.isSpare = function () {
+  return this.currentFrame.reduce((a, b) => a + b, 10);
+};
+
+// Scorecard.prototype.spareBonus = function () {
+//
+// };
+
+Scorecard.prototype.isGutter = function () {
+  return this.getScore === 0;
```
breycarr: Why do you need to call a method to mark something as a gutter or a perfect?

src/Bowling.js
```diff
@@ -0,0 +1,54 @@
...
+};
+
+Scorecard.prototype.isPerfect = function () {
+  return this.getScore === 300;
+};
+
+Scorecard.prototype.endGame = function () {
+    return this.frameCounter === 10;
```
breycarr: Did you intend to add this to the rolls function logic using `if (endGame) {}`?

PR #991, opened 2019-06-03
Bowling Challenge - vastly incomplete!

olliebrownlow:


spec/BowlingScorekeeperSpec.js
```diff
@@ -0,0 +1,145 @@
...
+describe("BowlingScorekeeper", function() {
+  var bowlingScorekeeper;
+
+  beforeEach(function(){
+    bowlingScorekeeper = new BowlingScorekeeper();
+  });
+
+  it("allows a player to enter their name", function() {
+    bowlingScorekeeper.addName("Ollie")
+    expect(bowlingScorekeeper._playerName).toEqual("Ollie")
```
alicelieutier: This test should not be here - don't test the value of instance variables, that is part of the implementation. The test below is good instead, as you are testing a public method of the class.

spec/BowlingScorekeeperSpec.js
```diff
@@ -0,0 +1,145 @@
...
+  it("has a method to get the entered name", function() {
+    bowlingScorekeeper.addName("Ollie")
+    expect(bowlingScorekeeper.name()).toEqual("Ollie")
+  });
+
+  it("starts with a current score of zero", function() {
+    expect(bowlingScorekeeper._currentScore).toEqual(0)
+  });
+
+  it("starts with an empty current frame array to store the rolls in a given frame", function() {
```
alicelieutier: Same as above

spec/BowlingScorekeeperSpec.js
```diff
@@ -0,0 +1,145 @@
...
+      bowlingScorekeeper.addRollScore(5)
+      expect(bowlingScorekeeper._currentFrame).toEqual([5])
+    });
+
+    it("pushes each frame into an array of frames", function(){
+      bowlingScorekeeper.addRollScore(5)
+      bowlingScorekeeper.addRollScore(4)
+      bowlingScorekeeper.addRollScore(5)
+      bowlingScorekeeper.addRollScore(4)
+      expect(bowlingScorekeeper._arrayOfFrames).toEqual([[5, 4], [5, 4]])
```
alicelieutier: Basically the rule is no `._` in tests.

PR #994, opened 2019-06-10
Incomplete Bowling Challenge

fetc90:
Unfinished due to restarting it.


spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
alicelieutier: You may want to remove this from your project - clean up unused files.

spec/frameSpec.js
```diff
@@ -0,0 +1,71 @@
+//
```
alicelieutier: Remove? Or do not commit.
If it was committed at some point, you can always go back through the commit history to find it again.

spec/scoreCardSpec.js
```diff
@@ -0,0 +1,88 @@
+describe("Scorecard", function() {
+  var scorecard;
+
+  beforeEach(function() {
+    scorecard = new Scorecard();
+  });
+
+  it('player starts on frame 1', function() {
```
alicelieutier: This and the next few tests are testing the initial values of instance variables. This is not needed, and it focuses on implementation rather than behaviour.

spec/scoreCardSpec.js
```diff
@@ -0,0 +1,88 @@
...
+
+  it('sets the bonus points to zero', function() {
+    expect(scorecard.bonusScorePoints).toEqual(0);
+  });
+
+  describe('#endFrame', function() {
+
+    it('switches isFrameActive to false', function() {
+      scorecard.endFrame();
+      expect(scorecard.isFrameActive).toBeFalsy();
```
alicelieutier: Same comment. You want to test the behaviour induced by this.

alicelieutier: Or, maybe this should be a private method and not be tested.

PR #995, opened 2019-06-10
Cosmin's bowling

Cosmin-Croitoriu:


spec/BowlingSpec.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable no-undef */
+'use strict'
```
hughbric: Nice, using strict mode.

spec/BowlingSpec.js
```diff
@@ -0,0 +1,46 @@
...
+describe('Bowling', function () {
+  var bowling
```
hughbric: I don't see any semicolons at the end of your declarations. In some places you may need them.

src/Bowling.js
```diff
@@ -0,0 +1,43 @@
...
+      rollIndex++
+    } else if (isSpare()) {
+      result += getSpareScore()
+      rollIndex += 2
+    } else {
+      result += getNormalScore()
+      rollIndex += 2
+    }
+  }
+  return result
```
hughbric: Is this return a part of a function? Is it part of the .score function? Not too sure on why it is here.

src/Bowling.js
```diff
@@ -0,0 +1,43 @@
+function Bowling () {
+  this.rolls = []
+}
+Bowling.prototype.roll = function (pins) {
+  this.rolls.push(pins)
+}
+Bowling.prototype.score = function () {
+  var result = 0
+  var rollIndex = 0
+  var bowling = this
```
hughbric: You declare this variable in the function but don't use it. Maybe move it into the constructor if you need it elsewhere? Is it necessary to have a var = this?

src/Bowling.js
```diff
@@ -0,0 +1,43 @@
...
+  for (var frameIndex = 0; frameIndex < 10; frameIndex++) {
```
hughbric: There's a lot going on here. It might be worth separating out your concerns. Maybe have a class that creates frames and calculates that frame's score and then a class that manages the game.

MIT.LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2008-2017 Pivotal Labs
```
alicelieutier: Why is your code copyright of Pivotal Labs? 🤔

Cosmin-Croitoriu:
Oh, is from Jasmine, I got it as a zip, and it seems that I was unzipping
it into my projects. I haven't noticed it until now. Is that a problem?

PR #996, opened 2019-06-29
James H Bowling Challenge Solution

zepherine2006DEV:


src/Scorecard.js
```diff
@@ -0,0 +1,170 @@
...
+      bonus = nextFrame.scoreRoll1 + nextFrame.scoreRoll2;
+    }
+  }
+  else if (this.isStrike(thisFrame)) {
+    bonus = nextFrame.scoreRoll1 + nextFrame.scoreRoll2;
+  }
+  else if (this.isSpare(thisFrame)) {
+    bonus = nextFrame.scoreRoll1;
+  }
+  return bonus;
```
LucyMHall: this is a tad complicated, but that's because the bowling rules are complicated

PR #1007, opened 2019-07-08
Celia Bowling Partial Solution 

celfro:


jasmine-standalone-3.4.0/spec/scoreSpec.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it("can roll a strike", function() {
+      game.roll(10);
+      game.roll(4);
+      game.roll(3);
+      rollMany(0,16);
+      expect(game.score()).toBe(24);
+  });
+
+  var rollMany = function(pins, rolls) {
```
ajosephides: I like the addition of this function. Neat idea.

jasmine-standalone-3.4.0/spec/scoreSpec.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it("can roll a spare", function() {
+      game.roll(5);
+      game.roll(5);
+      game.roll(3);
+      rollMany(0,17);
+      expect(game.score()).toBe(16);
+  });
+
+  it("can roll a strike", function() {
```
ajosephides: I would change the it here to be - can roll a single strike.
There is another use case where you can roll multiple strikes in a row. However - I may being premature here as you function might cover all these.

jasmine-standalone-3.4.0/src/game.js
```diff
@@ -0,0 +1,24 @@
...
+
+BowlingGame.prototype.roll = function(pins) {
+  this.rolls.push(pins);
+};
+
+BowlingGame.prototype.score = function () {
+  var result = 0;
+  var rollIndex = 0;
+
+  for (var frameIndex = 0; frameIndex < 10; frameIndex++) {
```
ajosephides: I'm intrigued by frameIndex here. How does frameIndex know that there are 2 rolls per frame? As it stands this loop reads (I may be wrong) but that it only looks at the first 10 rolls. In your tests because the majority of rolls are padded with 0s this might be hiding something under the surface. I'd suggest writing out a full game test to explore this.

ajosephides: Ah - is the 10 frames coming in because rollIndex is incremented by 2 each time?

PR #1008, opened 2019-07-08
Rhys partial solution

rhysco8:


README.md
```diff
@@ -1,77 +1,167 @@
...
+open SpecRunner.html
+```
+
+## Approach
+
+I laid out the requirements of the app as User Stories, detailed below. To keep the build simple, I implemented the requirements in stages.
+
+#### Business logic
+
+* Version 1 does not recognise spares, strikes or frames, so the player rolls 20 times and their score is added together.
```
rachjgriff: This is really helpful to understand your approach and link to the user stories

README.md
```diff
@@ -1,77 +1,167 @@
...
+git clone git@github.com:rhysco8/bowling-challenge.git
+cd bowling-challenge
+```
+
+### Run tests
+
+```sh
+open SpecRunner.html
+```
+
```
rachjgriff: Would be great to get some instructions on how to play the game in the console

src/bowlingGame.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+function BowlingGame () {
+  this.rolls = [];
+  this.finished = false;
+  this.numberOfRolls = 0;
+};
+
```
rachjgriff: This is clever! I hadn't though about feeding in the 'game' as an argument!  Might be me, but took me a little while to work it out so instructions would be helpful.

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,41 @@
...
+        game.roll(0);
+      };
+
+      expect(game.score(game)).toEqual(0)
+      expect(game.finished).toBe(true)
+    });
+  });
+
+  describe('average game', function () {
+    it('consists of 20 rolls that each score 4', function () {
```
rachjgriff: Great use of for loops to keep tests DRY

PR #1018, opened 2019-08-04
Richies bowling challenge

richieganney:


README.md
```diff
@@ -1,3 +1,28 @@
+Bowling Challenge - Richie Ganney
+=================
+
+# User stories
```
willhowes: Like the user stories. Nice and clear what the program needs to do.

public/src/Frame.js
```diff
@@ -0,0 +1,24 @@
+function Frame(){
```
willhowes: Good use of Javascript's constructor function to have separate objects for Frame and Scorecard. Means you are complying with SRP and code is nice and readable.

public/src/Frame.js
```diff
@@ -0,0 +1,24 @@
...
+  this._allFrames = []
```
willhowes: like the use of the private variable convention for ```_.allFrames```

public/src/Frame.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+Frame.prototype.rollOneToNine = function(rollOne, rollTwo) {
```
willhowes: The functions for this class are nice and clear

public/src/ScoreCard.js
```diff
@@ -0,0 +1,73 @@
...
+// change this so that the array isn't merged
+ScoreCard.prototype.strikes = function(arr) {
+  return arr.map(function(item, index){
+    if(item[0] === (10) && item.length < 3){
+     return item.concat(arr[index+1])
+  } else {
+     return "no"
+  }
+}).map(function(item, index) {
+  return item !== "no" ? item.reduce(function(a,b){return a+b}) : item;
```
willhowes: learned some new JS methods reading this ```reduce``` and ```concat```. Nice.

PR #1022, opened 2019-08-05
Genny Bowling Challenge

gennyallcroft:


src/Game.js
```diff
@@ -0,0 +1,48 @@
...
+  this.is_a_spare = false;
+  this.frame_count = 0;
+};
+
+Game.prototype.total_score = function(roll1, roll2) {
+  this.add_to_score(roll1, roll2);
+  return this._total_score;
+};
+
+Game.prototype.frame_score = function(roll1, roll2) {
```
ckettell: When rolling a strike, I get a NaN when leaving the second roll empty. Could add a default parameter of 0 for roll 2?

interface.js
```diff
@@ -0,0 +1,204 @@
...
+    frame2_roll1 = Number(roll1_input);
+  })
+
+  $('#frame2_roll2').change(function() {
+    roll2_input = $('#frame2_roll2').val();
+    frame2_roll2 = Number(roll2_input);
+  })
+
+  $('#frame2').click(function() {
+    game.frame_score(frame2_roll1, frame2_roll2);
```
ckettell: Could may add a condition in here which stops the frame's score exceeding 10. Butttttt, I do think there would be better ways to do this!

src/Game.js
```diff
@@ -0,0 +1,48 @@
+function Game() {
+  this._total_score = 0;
+  this._frame_score = 0;
```
ckettell: Because you have frame score, you might be able to write a function which checks if the frame score is 10 and if it is a strike. Means you would be able to increase the number of goes you get from the bonus

src/Game.js
```diff
@@ -0,0 +1,48 @@
...
+  this.is_a_strike = false;
+  this.is_a_spare = false;
+  this.frame_count = 0;
+};
+
+Game.prototype.total_score = function(roll1, roll2) {
```
ckettell: Small thing which will seem incredibly pedantic! In JS, they use camelCase instead of underscores_for_functions

src/Game.js
gennyallcroft: completely agree! I also wanted to have the second roll to adapt to the first one so you couldn't roll more than 10 in one frame.

interface.js
gennyallcroft: agree! Was thinking of just restricting it in the front end

src/Game.js
gennyallcroft: good point!

src/Game.js
gennyallcroft: yes! have slightly ignored the strike on the 10th frame - was thinking I could just have something clever on the front end but didn't get that far - agree a function would have worked better!

PR #1060, opened 2019-10-20
Incomplete

melvinlau:
Bonuses due to 'strike' and 'spare' occasions have not been dealt with.
HTML display needs to be fixed to show the result table rows correctly.
I was hoping to store the game results in a JS object called this.frames, which seems to log the number of knocked down pins at each roll.
Better luck next weekend.


melvinlau:
Outstanding tasks:
- Deal with how to process 10th frame bonus rounds
- Refine the function to update results dynamically on HTML page
- Validate the user input; limit no. of pins that can be knocked down
- A better user interface

PR #1076, opened 2019-11-15
rob baile - pull request

robbaile:
This code is not 100% working but this is where I got up to.


README.md
```diff
@@ -2,76 +2,27 @@
...
 
+## SETUP 
 
-* Challenge time: rest of the day and weekend.
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday week
+* Clone the project into a folder of your choice
+* Run 'npm install' to install relevant npm packages
```
david-mears: In fact, I found I had to run npm install in two places: the root directory, and then also in client. When I had only run npm install in the root directory, I got:

```
$ cd client
$ npm start

> client@0.1.0 start /Users/david/makers_apprenticeships/assessments/bowling-challenge/client
> react-scripts start

sh: react-scripts: command not found
```

I notice you have package.jsons at two levels: in the 'client' folder and in the root directory. So I now have two node_modules folders. Ideally these dependencies would be all managed in one place, as I'm sure you know. And I see there are two copies of 'bowling.js', one for the client and one for the tests; again, it would be good to only need to change one file to change one thing.

client/src/lib/bowling.js
```diff
@@ -0,0 +1,71 @@
+class Bowling {
```
david-mears: I want to query this name as being a little vague. A class called Bowling might represent the game, the bowling alley, or something else.

client/src/lib/bowling.js
```diff
@@ -0,0 +1,71 @@
...
+    isAStrike() {
+        return this.rolls[this.rollsIndexPosition] === 10;
+    }
+
+    isASpare() {
+        return (this.rolls[this.rollsIndexPosition] + this.rolls[this.rollsIndexPosition + 1]) === 10;
+    }
+
+    isLastFrame() {
+        return this.frame === 10;
```
david-mears: I'm pretty sure that you are zero-indexing, so the last frame of 10 frames will be index 9. This broken logic would prompt another test to be written.

client/src/lib/bowling.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    calculateScore() {
+        while (this.rollsIndexPosition < this.rolls.length - 1 && this.frame < 10) {
+            this.calculateRollScore();      
+            // console.log(`frame: ${this.frame} index: ${this.rollsIndexPosition} score: ${this.score}`) 
+        }
+    };
+
+    calculateRollScore() {
```
david-mears: The logic in this method is hard to understand, probably due to the nesting of conditionals. I couldn't work out what rollsIndexPosition was tracking. If you extracted a class `Frame` which stored its rolls, I think this logic would be simpler.

spec/gameSpec.js
```diff
@@ -0,0 +1,133 @@
...
+        it('should have a score added to the rolls array when a roll is made', () => {
+            bowling.roll(5)
+            expect(bowling.rolls[0]).toEqual(5)
+        });
+
+        it("should not be able to roll a negative number", () => {
+            expect(() => bowling.roll(-2)).toThrow("roll must be between 0 and 10");
+        });
+
+        it("should not be able to roll a number bigger than 10", () => {
```
david-mears: Good testing for edge cases.

spec/gameSpec.js
```diff
@@ -0,0 +1,133 @@
...
+        })
+        
+        it("test a strike in the last frame gets a two roll bonus that is counted once", () => {
+            let rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 7, 1];
+            rolls.map(roll => bowling.roll(roll));
+
+            expect(bowling.getScore()).toEqual(18); 
+        });
+
+        it("test rolling a spare with the two roll bonus does not get a bonus roll", () => {
```
david-mears: I'm not sure this test is sufficient to prove what it's meant to. I think we'd need to see that adding another roll to `rolls` on line 106 either doesn't change the score calculation or throws an error.

david-mears:
Generally a good implementation, and a clean React UI :)

I think the complexity of the logic demands that you extract out another class. You can check some example solutions to see how that could work: https://github.com/makersacademy/bowling-challenge/blob/master/docs/review.md#example-solutions

PR #1081, opened 2019-11-18
David's Bowling-Challenge

DMartinBGL:


david-mears:
Decent MVP, good tests and TDD.

david-mears:
Good idea to separate the ScoreCard and the Frame functionality into those classes.

src/frame.js
```diff
@@ -0,0 +1,16 @@
+class Frame {
+  constructor() {
+    this.bowlsLeft = 2;
+  }
+
+  bowl() {
+   if (this.isFrameOver) {
+     return;
+   }
+   return this.bowlsLeft -= 1;
```
david-mears: Why should the `bowl` method be returning anything?

PR #1099, opened 2019-11-25
first commit - should have done in earlier

edcourage:
Had to restart!!! this was wasn't going anywhere!


edcourage:
Bit late but logic is now complete!

PR #1146, opened 2020-03-23
Bowling challenge

Kittaru87:


rafahg:
This is review comment on Sophia's code.

- Appreciation 
    I love your code, is simple and do the job really good.
    I particularly love how you solve the challenge just using arrays and not getting into any more complicated system of hashes (not necessary at all).
   Your refactored methods are lovely and the process is quite clear and nice.
    Your test are clear and i specially like the one which throws and error if there is more than 10 pins knockdown in a roll (quite impossible, but some people managed to dry cats in a microwave so, you never know!).
 - Suggestion : You followed a logic quite similar to me doing the challenge so, i don't have any real suggestion to make, maybe just to write a few more comments to explain your logic process for someone who reads the code without previous knowledge of the challenge or even the lenguaje.
  it would be nice too,  to run a passing test for a complete game, to check if the code calculates the total score correctly. 
 
    Final comment : Nice code, awesome refactoring and clear idea behind the whole thing. Congratulations!.

PR #1148, opened 2020-03-27
Rita: added .travis.yml

lxierita:


app.rb
```diff
@@ -0,0 +1,10 @@
+require 'sinatra'
+
+class Scorecard < Sinatra::Base
+
+	get "/" do 
+		File.read(File.join('./', 'index.html'))
```
lindseydew: Nice use of a server here!

PR #1150, opened 2020-03-29
Bowling Challenge

jarasmar:
Basic specifications implemented in backend. The front end is not fully connected to the back end, it fills a chart generating random rolls for every frame, keeps the score of the whole game but doesn't add any bonus points.


KG700:
I really like how your's looks and your Game.js code is really tidy and DRY. In your interface.js I you could try modifying your frame methods to use classes rather than ids and see if you can combine them into one method with string interpolation.

PR #1158, opened 2020-03-30
Karla's bowling challenge

KG700:


jarasmar:
Hola Karla! Im really impressed by how neat your code looks. I like that you split all the functionality in so many different classes, that way it looks super readable and its very easy to understand what it is doing. Also having implemented a multiplayer option is really cool. The interface, as we were talking looks a bit less readable, but it works pretty good as well, so I guess it just needs some refactoring that it is still unreachable for us, we'll get there ;)
As a suggestion, I could only say that maybe having some basic CSS layout would make you feel a bit more confortable with showing your work. As I said, I was very impressed when you showed me how it worked, but I had the feeling that you thought it was just too ugly. Not that is needed at this point, but it might help you. Anyways, if you ask me, its awesome as it is! Have a lovely week

PR #1165, opened 2020-04-26
Hey cousin, let's go bowling!..... but don't tell me the score til the end.

Edward-Phillips:
had a lot of trouble making the page be kind of responsive to different devices, code is ugly AF, especially in interface.js sorry about that. I wrote about areas to improve on in the readme, if you want to see how it looks without cloning the repo you can go to hey-cousin-lets-go-bowling.surge.sh.


interface.js
```diff
@@ -0,0 +1,142 @@
...
+    cell.innerHTML= 'Foul'
+  }
+;}
+function inputChecker(number) {
+  if (number >= 0 && number <= 10) {
+    return true;
+  }
+  return false;
+};
+
```
grahamfalconer, line 141: ^ Could some refactoring be done here?

spec/BowlingSpec.js
```diff
@@ -0,0 +1,114 @@
...
+      playerOne,
+      playerTwo;
+
+  beforeEach(function() {
+    bowling = new Bowling();
+    playerOne = "Tom";
+    playerTwo = "Alex";
+  });
+
+  it(" should hae a list of players", function() {
```
grahamfalconer, line 12: Is this necessarily what your test does?

spec/BowlingSpec.js
```diff
@@ -0,0 +1,114 @@
...
+    expect(bowling.getCurrentFrame()).toEqual(1);
+  });
+
+  it(" should be able to increment the frame once every player has had their turn", function() {
+    bowling.addPlayer(playerOne);
+    bowling.addPlayer(playerTwo);
+    bowling.changeTurn();
+    bowling.changeTurn();
+    expect(bowling.getCurrentFrame()).toEqual(2);
+  });
```
grahamfalconer, line 48: I like the '.getCurrentFrame()' method, very nifty.

spec/BowlingSpec.js
```diff
@@ -0,0 +1,114 @@
...
+      bowling.saveCurrentPlayerFrame();
+    }
+    bowling.storeFirst(10);
+    bowling.saveCurrentPlayerFrame();
+    bowling.storeFirst(4);
+    bowling.storeFirst(2);
+    bowling.saveCurrentPlayerFrame();
+    expect(bowling.gameEnded()).toEqual(false);
+  });
+});
```
grahamfalconer, line 114: LGTM

spec/PlayerSpec.js
```diff
@@ -0,0 +1,67 @@
+describe("Player", function() {
+  var playerOne;
+
+  beforeEach(function() {
+     playerOne = new Player("steve");
+  })
+  it(" should be able to store the player name", function() {
```
grahamfalconer, line 7: again, is this test doing necessarily doing what it says?

src/Player.js
```diff
@@ -0,0 +1,79 @@
...
+    this.gameOver = true;
+    this.calcScore();
+  } else if(this.tenthIsSpare) {
+    this.bowls.push(one);
+    this.gameOver = true;
+    this.calcScore();
+  }
+}
+  this.frameCount += 1;
+};
```
grahamfalconer, line 51: ^could this be shorter>

style.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+#names-pls {
+  display: inline-block;
+  background-color: black;
+  opacity: 90%;
+  color: white;
+  width: 58%;
+  text-align: center;
+}
```
grahamfalconer, line 54: ^ NICE STYLING 😎

interface.js
Edward-Phillips: yeah, kind of rushed at the end to get it out, definitely could refactor to clearer methods

spec/BowlingSpec.js
Edward-Phillips: no I guess you've got me here, I guess this test is that it should be able to hold players in a list before any players are added

spec/PlayerSpec.js
Edward-Phillips: I guess I'm actually testing it can tell me the name of the player more than being able to just store the player

src/Player.js
Edward-Phillips: Yeah, I probably need to extract the two ways that the scores are pushed, either just bowl one pushed or both bowl one and bowl two, then maybe I could remove the tenth is strike method and just have an is strike method and use that with an if statement to set the value of the tenthIsStrike/Spare boolean attribute

style.css
Edward-Phillips: haha cheers, definitely trying to up my styling game, got a way to go but this challenge definitely helped me on the way there

interface.js
```diff
@@ -0,0 +1,142 @@
...
+    x[index].style.display = "block"
+  };
+};
+function makeTable(players) {
+  var perrow = 1, // 1 cells per row
+      count = 0, // Flag for current cell
+      table = document.createElement("table"),
+      row = table.insertRow(),
+      cell = row.insertCell();
+      cell.innerHTML = 'Names\\Frames';
```
alicelieutier, line 93: This line should be indented at the same level as `var` above.

interface.js
```diff
@@ -0,0 +1,142 @@
...
+       if (count%perrow==0) {
+         row = table.insertRow();
+        }
+      };
+      
+      table.id = "score-board"
+      $("#game-board").append(table);
+};
+function addBallsToTable(ballOne, ballTwo) {
+  var cell = document.getElementById(`${bowling.getCurrentPlayer().name}${bowling.currentFrame}`)
```
alicelieutier, line 124: Too much happening on one line here.

spec/PlayerSpec.js
```diff
@@ -0,0 +1,67 @@
...
+    expect(playerOne.name).toEqual("steve");
+  });
+
+  it(" should be able to store the player's bowls", function() {
```
alicelieutier, line 11: It feels like you are testing the constructor here... Is reading `bowl` a behaviour of your class or just part of the implementation?

interface.js
```diff
@@ -0,0 +1,142 @@
...
+    cell.innerHTML= `${ballOne},${ballTwo}`
+  } else if(ballOne < 10 && ballOne + ballTwo === 10) {
+    cell.innerHTML= '/'
+  } else if(ballOne === 10) {
+    cell.innerHTML='X'
+  } else {
+    cell.innerHTML= 'Foul'
+  }
+;}
+function inputChecker(number) {
```
alicelieutier, line 135: For things like this that should not happen, you could throw an error instead.

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
alicelieutier, line 1: Remove Jasmine examples.

src/Player.js
```diff
@@ -0,0 +1,79 @@
...
+if (this.frameCount < 10) {
+  if(one <= 9) {
+    this.bowls.push(one);
+    this.bowls.push(two);
+  } else if( one == 10) {
+    this.bowls.push(one);
+  } else {
+    this.bowls.push(0);
+    this.bowls.push(0);
+  }
```
alicelieutier, line 21: What is this for? Should this throw an error?

src/Player.js
alicelieutier: Yes. But even as a small refactor you could have a local variable to store whether the tenth frame is a strike, since you only use it in one function.

PR #1168, opened 2020-04-27
bowling challenge 

Gina-Frankel:


src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+Bowling.prototype.isSkipped = function (rollCount) {
+  if (rollCount === 2 && this.strike === true) {
+    return true;
+  }
+};
+
+Bowling.prototype.makeCardTemplate = function () {
+  /// changed obj but perhaps was better before as means I have to make conditions for runCardmaking
+  var frameCount,
+    key,
```
FayeCarter, line 63: Would separating methods linked to individual frames into a frame object help here?

src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+    obj[frameCount] = {};
+    // obj[key] = { r1PinsDown: 0, r1Score: 0, r2PinsDown: 0, r2Score: 0 };
+    this.card = obj;
+  }
+  return this.card;
+};
+
+Bowling.prototype.getInput = function (input) {
+  input = parseInt(input,10)
+  this.input = input
```
FayeCarter, line 75: These two lines could be one line of code
this.input = ParseInt(input,10)

src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+};
+
+Bowling.prototype.scoreStrike = function (rollCount, r1PinsDown) {
+  if (this.strike === true && rollCount === 2) {
+    var previousFrameKey = this.frameKey - 1;
+    var r2PinsDown = this.card[this.frameKey]["r1PinsDown"];
+    this.card[previousFrameKey]["r2Score"] =
+      this.card[previousFrameKey]["r2Score"] + r1PinsDown + r2PinsDown;
+    this.score = this.score + r1PinsDown + r2PinsDown;
+    // if  previous frame [r2 pinsDown] == x then add 10
```
FayeCarter, line 94: If you separated this into multiple methods would it be easier to reuse bits of the code?

FayeCarter:
Overall, the projects was great. Readme was clear and it was great you hi-lighted your focus and methods that you had tried.
I think the code is not yet DRY but specification is covered well

spec/BowlingSpec.js
```diff
@@ -0,0 +1,229 @@
...
+      });
+
+      it("If a spare is not thrown and spare is true, spare will change to  false", function () {
+        // the way that this is testing it is not quite how the program works as a first roll would change it to 'spare' state but this would need to add additional parts to the test which I am not sure are needed to test what I need to know 
+        var alreadyCalled = false;
+        spyOn(bowling, "switchStrike").and.callFake(function () {
+          if (alreadyCalled)  return this.strike = false 
+            alreadyCalled = true;
+            return (this.strike = true);
+        });
```
alicelieutier, line 130: You should not be mocking methods of the class under test.

src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+  this.makeCardTemplate();
+  for (frameCount = 1; frameCount < 11; frameCount++) {
+    this.frameKey = frameCount;
+    for (rollCount = 1; rollCount < 3; rollCount++) {
+      if (this.isSkipped(rollCount)) {
+        pins = "x";
+        score = this.calculateScore(0);
+      } else {
+       
+        pins = this.input;
```
alicelieutier, line 34: What is `this.input` here?

src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+    return (this.spare = false);
+  }
+};
+
+ Bowling.prototype.switchStrike = function (rollCount, pins) {
+  if (
+    (rollCount === 1 && pins === ALL_PINS) ||
+    (this.frameKey > 1 &&
+      rollCount === 2 &&
+      this.card[this.frameKey - 1]["r2PinsDown"] === "x") //too much functionality in one method?
```
alicelieutier, line 137: Too much logic in one conditional, yes.

src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+  this.strike = false;
+  this.input 
+  //this.rollCount = 1   maybe this should be instead of passing rollCount;
+}
+
+function getRndInteger(min, max) {
+  return Math.floor(Math.random() * (max - min)) + min;
+}
+
+Bowling.prototype.runCardMaking = function (input) {
```
alicelieutier, line 20: This function is too long.

src/Bowling.js
```diff
@@ -0,0 +1,144 @@
...
+        pins = this.input;
+
+        this.switchStrike(rollCount, pins)
+         
+        this.scoreStrike(rollCount, pins);
+        score = this.calculateScore(pins);
+      }
+      this.fillCard(pins, score, rollCount);
+
+      if (this.strike === false && this.spare === true) {
```
alicelieutier, line 43: Throughout this class, you rely a lot on state (or instance variables). This makes the understanding more difficult.
What you have now is something like:
- Method A calculates something, and stores it in X
- Later in the function, you read the value at X.

This is not very good design, because you have to always make sure to update all values before you read from them. Instead, you should be doing:
- call method A that calculates a value, and returns it. 

More generally, instance variables should be kept to a minimum, and to things that describe the current state of the object - not the current state of your computation! Local variable are here for that.

alicelieutier: This is actually a thing I see a lot with Makers devs.

PR #1169, opened 2020-04-27
5/10

grahamfalconer:


index.html
```diff
@@ -0,0 +1,62 @@
...
+        <label for="secondBowl"> Second Bowl:</label><input type="text" id="usersSecondBowl"><br>
+      </section>
+      <input type="submit"> -- </input>
+      <h2 id="result_round_one"></h2>
+    </form>
+
+    <form id="round_two"> ROUND TWO
+      <section>
+        <label for="firstBowl2"> First Bowl:</label><input type="text" id="usersFirstBowl"><br>
+        <label for="secondBowl2"> Second Bowl:</label><input type="text" id="usersSecondBowl"><br>
```
Edward-Phillips, line 22: repeating ids for form inputs, should work if you change them to the eexpected value in interface.js

spec/scorecardSpec.js
```diff
@@ -0,0 +1,35 @@
+describe('Scorecard', function() {
+  var scorecard;
+  beforeEach(function() {
+    scorecard = new Scorecard();
+  });
+  // The first test - Does my frame contain 3 objects [Score1, Score2, Score3[sum]]
+
+  it('It allows firstbowl value to be added to be added to frame', function() {
+    expect(scorecard.firstBowl(5)).toEqual(scorecard.bowlOne)
```
Edward-Phillips, line 9: nice test, I like it!

spec/scorecardSpec.js
```diff
@@ -0,0 +1,35 @@
...
+    scorecard.secondBowl(3)
+    scorecard.roundSummary()
+    expect(scorecard.frameTotal).toEqual([8])
+  });
+
+  xit('If first bowl is a strike, second bowl is skipped', function() {
+    //if bowl.One
+
+  });
+
```
Edward-Phillips, line 33: I like the idea of this test, definitely going in the right direction!

src/scorecard.js
```diff
@@ -0,0 +1,27 @@
...
+  this.bowlTwo.push(num);
+}
+
+Scorecard.prototype.roundSummary = function() {
+  this.frameTotal.push(this.bowlOne[0] + this.bowlTwo[0]);
+}
+
+Scorecard.prototype.printFrameScore = function() {
+  return this.frameTotal
+}
```
Edward-Phillips, line 22: nice and short methods, excellent!

index.html
```diff
@@ -0,0 +1,62 @@
...
+      <input type="submit"> -- </input>
+      <h2 id="result_round_five"></h2>
+    </form>
+    <section>
+      <h1 id="allScores"> Hello </h1>
+
+    </section>
+
+    <script src="src/scorecard.js"></script>
+    <script src="interface.js"></script>
```
Edward-Phillips, line 60: I've seen this in a bunch of places but not sure why these're put at the bottom, is it a style choice or is there a technical motivation?

interface.js
```diff
@@ -0,0 +1,36 @@
...
+    $('#round_two').submit(function(event) {
+      event.preventDefault();
+      var ball1 = parseInt($('#usersFirstBowl2').val()),
+          ball2 = parseInt($('#usersSecondBowl2').val());
+        $('#usersFirstBowl2').val('');
+        $('#usersSecondBowl2').val('');
+        scorecard2.firstBowl(ball1);
+        scorecard2.secondBowl(ball2);
+        scorecard2.roundSummary();
+        round2Total();
```
Edward-Phillips, line 26: could this be dried out?

src/scorecard.js
```diff
@@ -0,0 +1,27 @@
+// the following code is for one frame in a scorecard...
```
alicelieutier, line 1: Surely the name should be Frame, then, not scorecard? 🤔

interface.js
```diff
@@ -0,0 +1,88 @@
+
```
alicelieutier, line 1: Your interface is doing a lot of logic. You would probably benefit from having one more logic class that handles the game, and then the interface talking to that class, without doing any logic.

interface.js
alicelieutier: Agreed.

spec/scorecardSpec.js
alicelieutier: I'm not sure what this is testing. What is the "behaviour" of firstBowl? From a user of your class point of view, what is it supposed to produce?
The confusion here for me is because you have code on both sides of the matcher (`scorecard.firstBowl(5)` and `scorecard.bowlOne`) and it's not obvious to test readers what the values for each should be.
The expected output should be fixed, instead.
Let me know if that's not clear!

spec/scorecardSpec.js
```diff
@@ -0,0 +1,35 @@
...
+  it('Stores users firstBowl in bowlOne ', function() {
+    scorecard.firstBowl(5)
+    expect(scorecard.bowlOne).toEqual([5]);
+  });
+
+  it('Stores users secondBowl in bowlTwo', function () {
+    scorecard.secondBowl(3)
+    expect(scorecard.bowlTwo).toEqual([3]);
+  });
```
alicelieutier, line 20: These two test are testing state (or implementation) rather than behaviour.
The main behaviour here is actually described by the next test, that calculates the score, so you can remove these!

src/scorecard.js
```diff
@@ -0,0 +1,27 @@
...
+  this.bowlOne.push(num);
+}
+
+Scorecard.prototype.secondBowl = function(num) {
+  this.bowlTwo.push(num);
+}
+
+Scorecard.prototype.roundSummary = function() {
+  this.frameTotal.push(this.bowlOne[0] + this.bowlTwo[0]);
+}
```
alicelieutier, line 18: It's always better design to have methods that return values, rather than just change the class state.

alicelieutier: This way, this method and the next can be all rolled into one.

PR #1180, opened 2020-05-25
Catriona's bowling challenge

cmb84scd:


src/bowling.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+    return this.score[0] + this.score[1];
+  };
+
+  frame2() {
+    if(this.score[2] + this.score[3] === 10) {
+            return this.score[2] + this.score[3] + this.score[4] + this.frame1();
+        }
+    return this.score[2] + this.score[3] + this.frame1();
+  };
```
Patrick-Sawyer, line 24: Consider refactoring to put the separate functions into one function.

spec/bowlingSpec.js
```diff
@@ -0,0 +1,195 @@
...
+    it('checks and scores for a spare', function() {
+      for (let i = 0; i < 19; i++) {
+        bowl.scoring(4);
+      }
+      bowl.scoring(6);
+      bowl.scoring(4);
+      expect(bowl.frame10()).toEqual(86);
+    });
+  });
+});
```
Patrick-Sawyer, line 195: Your tests look very thorough.

src/bowling.js
Patrick-Sawyer: Maybe send the frame number as an argument to your frame method to avoid repetition.

PR #1186, opened 2020-05-26
Logic completed - Jquery in process - No travis

collier-jo:


spec/scoreSpec.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+console.log("I AM IN THE SPEC!")
```
tanilserbes, line 3: :)

spec/scoreSpec.js
```diff
@@ -0,0 +1,88 @@
...
+
+describe('Score', () => {
+
+  var score 
+  var game 
+
+  beforeEach(function(){
+    score = new Score
+    game = new Game // How to spy this?
```
tanilserbes, line 12: i usually spy functions instead of classes. I think at the first it's good start creating objects and then spy on them

src/Score.js
```diff
@@ -0,0 +1,45 @@
...
+    this.frames ++
+    return score 
+  };
+  
+  isStrike(roll) {
+    return roll === 10;
+  };
+
+  isSpare(roll, roll2) {
+    return roll + roll2 === 10
```
tanilserbes, line 42: let's say first roll is 10 which is strike.
and second roll is 0. 
it is strike right ? 
However isSpare function gives true as well .. 
maybe its better to refactor : if roll +roll2 ===10 and roll2!=null.. ?

PR #1189, opened 2020-06-20
Just a Scorecard...

RaeRachael:
Super basic atm, absolutely no css. Only a browser app, JQuery used to update the logic in the interface.js file. (will probably add some css and improve the readme, but wanted to make the pull request early and its currently functional)


interface.js
```diff
@@ -0,0 +1,282 @@
...
+  var score7 = new Score(frame7,frame8,frame9)
+  var score8 = new Score(frame8,frame9,frame10)
+  var score9 = new Score(frame9,frame10,frame11)
+  var score10 = new Score(frame10,frame11,frame12)
+
+  var scorecard = new Scorecard([score1, score2, score3, score4, score5,
+     score6, score7, score8, score9, score10])
+
+  $('#bonus1').attr('class', 'hidden')
+  $('#bonus2').attr('class', 'hidden')
```
tatsiana-makers, line 31: What is the reason for setting this attribute here inside of `interface.js`?

interface.js
```diff
@@ -0,0 +1,282 @@
...
+      askForRoll('#roll2frame2', roll)
+    }
+  })
+  $('#roll2frame2').submit(function() {
+    var roll = $('#roll').val();
+    frame2.roll2(roll);
+    $('#roll2frame2cell').html(roll);
+    updateScores()
+    askForRoll('#roll1frame3', 0)
+  })
```
tatsiana-makers, line 71: The code in this block (and the following blocks) is very similar to the code in https://github.com/makersacademy/bowling-challenge/pull/1189/files#diff-c680b607552f1865398553e1ea8ab421R35-R52

Is there a way to refactor these similar repeated blocks to follow the DRY principle of code?

interface.js
RaeRachael: as opposed to starting the bonus 1 and bonus 2 rows of the table in index.html as hidden at the start? (I guess I wanted to have a cleaner html file...)

interface.js
RaeRachael: yes of course, I think I my brain was looking for ways to create these functions as a means to DRY rather than extract out for them.

PR #1192, opened 2020-06-21
Bowling is hard!

minsiyang:


spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+describe('BowlingGame', function() {
+
```
ChrisCooney05, line 4: Overall, A good start to the testing. Working with baby steps is key to getting a full understanding of the challenge especially when working on your own

src/bowlingGame.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+class BowlingGame {
+  constructor() {
+    this.rolls = [];
+  }
+
```
ChrisCooney05, line 7: may be worth thinking about storing the frames as there own class, this will help separate things so that you don't just have a long string of numbers to deal with.

src/bowlingGame.js
```diff
@@ -0,0 +1,20 @@
...
+   }
+
+   score() {
+     var array = this.rolls
+     var total = array.reduce(function(a, b) {
+       return a + b;
+       }, 0);
+     return total; 
+   }
+
```
ChrisCooney05, line 19: Overall a good start to the bowling challenge, I think you were on the right lines with this but maybe just not 100% sure what direction to move forward with. This Challenge benefits from real heavy planing - what do I need to keep track of? frames, turn, score, spares, strikes... By knowing this all in advance it helps moving forward.

PR #1197, opened 2020-06-22
As far as I got in the bowling challenge

Sumner1185:


spec/frameSpec.js
```diff
@@ -0,0 +1,75 @@
...
+      expect(frame.bowlTwo).toEqual(2);
+    });
+
+    it('calculates the total of two bowls', function () {
+      expect(frame.calculateTotal()).toEqual(9);
+    });
+    
+  });
+
+  describe('when a bowler throws a 10', function () {
```
tatsiana-makers, line 29: Does 'when a bowler throws a 10' applies to all of the test cases within this `describe` block?

spec/scorecardSpec.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+describe('Scorecard', function () {
+  var scorecard;
+  var testFrame = { score: 8 };
```
tatsiana-makers, line 5: Nice use of fake frame objects within these unit tests!

src/frame.js
```diff
@@ -0,0 +1,37 @@
...
+      this.isStrike = true;
+    }
+    this.bowlOne += pins;
+    this.calculateTotal();
+  };
+
+  addBowlTwoScore(pins) {
+    this.bowlTwo += pins;
+    this.calculateTotal();
+    if(this.calculateTotal() === 10) {
```
tatsiana-makers, line 24: Could we refactor this to avoid having [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))? Are there other places in the code where we refactor magic numbers?

src/scorecard.js
```diff
@@ -0,0 +1,50 @@
...
+
+  isComplete() {
+    if(this.frames.length === 10) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  storeFrame(frame = new Frame) {
```
tatsiana-makers, line 16: What is the reason for having a default value of the input parameter here?

src/scorecard.js
```diff
@@ -0,0 +1,50 @@
...
+      var isAStrike = this.frames[i - 1].isStrike;
+      return isAStrike;
+    } else {
+      return false;
+    }
+  }
+
+  isPreviousFrameASpare(frame) {
+    if (this.frames.indexOf(frame) > 0) {
+      var i = this.frames.indexOf(frame);
```
tatsiana-makers, line 32: The logic for finding the previous frame is repeated in `isPreviousFrameAStrike` and this method. Could we extract this out to follow the DRY principle?

PR #1199, opened 2020-06-22
First bowling ch attempt

sofyloafy:


interface.js
```diff
@@ -0,0 +1,37 @@
+$(document).ready(function () {
```
victorvallet, line 1: Nice interface, easy to follow

public/style.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+button, input{
+  font-family: 'Pacifico', cursive;
+  font-size: 20px;
+
+}
+
+h1,h2{
+  margin-top: 0px;
```
victorvallet, line 22: Just a css shortcut when you need to put margins 0 for top/bottom/left/right you can use `margin:0;`

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,50 @@
+describe('BowlingGame', function() {
+
+  beforeEach(function () {
```
victorvallet, line 3: Nice, forgot about that one 👍

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,50 @@
...
+    expect(game.score()).toEqual(24)
+
+  });
+
+  it('can roll a perfect game', function(){
+    rollMany(10,12)
+    expect(game.score()).toEqual(300)
+  })
+
+  function rollMany(pins, rolls) {
```
victorvallet, line 43: Didn't know we could define a function in the test file, nice one!

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,50 @@
+describe('BowlingGame', function() {
+
+  beforeEach(function () {
+    game = new BowlingGame();
+  });
+
+  it('should begin with no score', function () {
+      expect(game.rolls).toEqual([]);
```
victorvallet, line 8: An alternative could be to test the length `expect(game.rolls.length).toEqual(0)`

src/bowlingGame.js
```diff
@@ -0,0 +1,32 @@
...
+  constructor() {
+  this.rolls = [];
+  this.total =[];
+  }
+
+  roll(pins){
+    this.rolls.push(pins)
+  }
+
+  score() {
```
victorvallet, line 11: I had quite a similar approach here but wondering if this one could be split into different functions

PR #1208, opened 2020-07-20
Bowling Challenge - Josh

JoshuaNg2332:
Bowling is hard.
I didn't start the front end but completed the backend logic.
Wasn't happy with the logic so started again with v2.


src/frame.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+    return this.score[24] + this.score[25] + this.frame8();
+  };
+
+  frame10() {
+    if(this.score[27] + this.score[28] === this.MAXSCORE) {
+            return this.score[27] + this.score[28] + this.score[29] + this.frame9();
+        }
+    return this.score[27] + this.score[28] + this.frame9();
+  };
```
alicelieutier, line 81: These functions are a great case for practising DRY-in your code. If we look at them, they are all doing the same thing. The only differences are:
* the frame number
* the indexes used in the `score` array.

Whenever that happens, you should think of how you can "parametrise" these, or turn them into variables.
For example:
```
this.score[24] + this.score[25] === this.MAXSCORE
```
can initially be rewritten as 
```js
start_index = 24
this.score[start_index] + this.score[start_index + 1] === this.MAXSCORE
```

Next, you can try to calculate `start_index` based on the frame number. So if we look at these numbers here:
Frame number  |  start_index
1 | 0
2 | 3
3 | 6
4 | 9
5 | 12

etc, you can start to see a pattern. So you can probably refactor your 10 functions into one, that takes the frame number as an argument, and then calculates the `start_index` based on it, for example:
```js
frame(frame_number) {
    start_index = // something based on frame_number
    if(this.score[start_index] + this.score[start_index + 1] === this.MAXSCORE) {
 // etc..
```

PR #1213, opened 2020-08-16
Harry LB - Bowling Challenge 

harrylb14:
All functionality working as expected. Used require and module.export to implement jasmine CLI and travis CI.


src/bowlingGame.js
```diff
@@ -0,0 +1,64 @@
...
+    this._rollScores = [];
+    this._frameNumber = 0;
+  }
+
+  addRoll(pins) {
+    this._rollScores.push(pins);
+  }
+
+  scoreGame() {
+    let score = 0;
```
rjkviegas, line 12: you can use parallel assignment eg: let score = 0, index = 0;

PR #1225, opened 2020-09-13
Bowling Challenge

Djura22:


spec/bowling_SC_spec.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('returns a total score', function() {
+    for (let i = 0; i < 20; i++) {
+      game.firstRoll(2);
+    }
+    expect(game.runningScore).toEqual(40);  
+  });
+  
+  describe('roll combinations', function() {
```
tatsiana-makers, line 27: Is this testing the same scenario as the test above?

spec/bowling_SC_spec.js
```diff
@@ -0,0 +1,68 @@
...
+      game.firstRoll(2);
+      game.secondRoll(2);
+      game.strikeCalc(); 
+      for (let i = 0; i < 17; i++) {
+        game.firstRoll(0);
+      } 
+      expect(game.runningScore).toBe(18);
+    });
+  });
+
```
tatsiana-makers, line 66: There are tests missing for a couple of special scenarios outlined in the [specification](https://github.com/makersacademy/bowling-challenge#bowling--how-does-it-work):
- 10th frame
- perfect game

spec/bowling_SC_spec.js
```diff
@@ -0,0 +1,68 @@
...
+  it('can create new game', function() {
+    let game = new bowlingGame();
+    expect(game).toBeInstanceOf(bowlingGame);
+  });
+
+  it('can roll a gutter game', function() {
+    for (let i = 0; i < 20; i++) {
+      game.firstRoll(0);
```
tatsiana-makers, line 15: How come we don't also use `secondRoll` here? Unless I'm misunderstanding what the functions are meant to be used for, as a user I would expect to alternate between calling `firstRoll` and `secondRoll` so we should mimic how a user would use our application in the tests

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
...
+    };
+  };
+
+  secondRoll(points) {
+    this.rolls[this.frameIndex -1].push(points);
+    this.runningScore += points;
+    this.frameIndex++
+  };
+
+  spareCalc() {
```
tatsiana-makers, line 26: It looks like this function is only called within a test. Is this a function that a user would be expected to remember to call themselves or could we add a call to this function somewhere in the code?

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
...
+    let bonus = 0;
+    const frameTotal = this.rolls[this.frameIndex - 2][this.rollIndex] + this.rolls[this.frameIndex - 2][this.rollIndex + 1];
+    if (this.isSpare(frameTotal)) {
+      bonus += this.spareBonus();
+    };
+    this.runningScore += bonus;
+  };
+
+  strikeCalc() {
```
tatsiana-makers, line 35: Same comment as for the `spareCalc` function above

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+class bowlingGame {
+
+  constructor() {
+    this.rolls = [[], [], [], [], [], [], [], [], [], []];
+    this.runningScore = 0;
+    this.frameIndex = 1;
```
tatsiana-makers, line 8: It's not obvious why the `frameIndex` begins at 1 - perhaps this is a good place to add a comment?

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
...
+    this.rollIndex = 0;
```
tatsiana-makers, line 9: It doesn't look like we ever reassign a new value to this variable - should this be a constant?

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+class bowlingGame {
+
+  constructor() {
+    this.rolls = [[], [], [], [], [], [], [], [], [], []];
```
tatsiana-makers, line 6: Currently each frame is represented by a Javascript Array and we're doing some calculations on each frame. Is there another way we could represent a frame which would allow us to achieve more encapsulation?

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
...
+
+  secondRoll(points) {
+    this.rolls[this.frameIndex -1].push(points);
+    this.runningScore += points;
+    this.frameIndex++
+  };
+
+  spareCalc() {
+    let bonus = 0;
+    const frameTotal = this.rolls[this.frameIndex - 2][this.rollIndex] + this.rolls[this.frameIndex - 2][this.rollIndex + 1];
```
tatsiana-makers, line 28: Is this the same calculation done in `strikeBonus`? Might there be another name for `strikeBonus` that would allow us to use that function here?

src/bowling_SC.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
```
tatsiana-makers, line 1: Quite readable code - nice!

tatsiana-makers (commented):
This is a really nice go at the bowling challenge! I've left some comments and suggestions - let me know if you want to chat through any of them (:

PR #1226, opened 2020-09-13
Something of a bowling challenge

Charlotte-91:
Front end works well with backend, But some logic is wrong somewhere as the scores do not add up correctly. 
I learnt more about the rules of bowling, but still not enough to get the logic to work properly...


src/interface.js
```diff
@@ -0,0 +1,36 @@
...
+      console.log(total)
+      var element = document.getElementById(`fs${num}`);
+      element.innerHTML = `${total}`;
+      num += 1
+      sn1 += 1
+      sn2 += 2
+    }
+  });
+
+
```
just-tam, line 35: Nice tidy interface 👍

PR #1229, opened 2020-09-14
Yasmin Bowling Challenge

yu2491:


Djura22:
great use of the single responsibility principle. Extracting the frame responsibilities into its own class was a good call.

Djura22:
README instructions could be more in depth.

Djura22:
Test coverage is very high. Although I'm not sure extracting the final frame into its own class was necessary. I would like to know your reasons for doing so.

src/finalFrame.js
```diff
@@ -0,0 +1,50 @@
...
+  getScore()  {
+    return this.score
+  }
+
+  isThirdBowlAllowed() {
+    return this.firstBowl() === 10 || this.firstBowl() + this.secondBowl() === 10;
+  }
+
+  isComplete() {
+    return this.isThirdBowlAllowed() === true && this.frame.length === 3 ||
```
tatsiana-makers, line 46: A comparison like `this.isThirdBowlAllowed() === true` can be simplified to `this.isThirdBowlAllowed()` 
and `this.isThirdBowlAllowed() === false` can be simplified to `!this.isThirdBowlAllowed()` 🙂

src/finalFrame.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+
+class FinalFrame {
```
tatsiana-makers, line 3: It seems like a `FinalFrame` is mostly the same as a `Frame`. Is there a way to remove this class and instead have the `Frame` class capture the necessary functionality?

spec/scorecardSpec.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+describe('Scorecard', function(){
+
+  var scorecard;
```
tatsiana-makers, line 5: Our Scorecard tests currently rely on the actual implementation of the Frame class. Can you think of a way to use [mocking](https://github.com/makersacademy/course/blob/master/pills/mocking_in_javascript.md) to isolate the unit tests?

src/scorecard.js
```diff
@@ -0,0 +1,62 @@
...
+
+  constructor() {
+    this.frames = [new Frame()];
+    this.currentFrameIndex = 0;
+  }
+
+  play(score) {
+    if(this.gameOver()) {
+      return;
+    } else {
```
tatsiana-makers, line 13: We can actually remove the `else` block in this case to make the code look cleaner:
```
if(this.gameOver()) {
      return;
}
this.gameSetup();
this.currentFrame().bowl(score);
this.addBonus(score);
```

tatsiana-makers (commented):
This is a really nice go at the bowling challenge! Your code is really readable 🎉  I've left a few comments and suggestions. Let me know if you want to chat through any of them 🙂

PR #1230, opened 2020-10-03
The Big Lebowlski

TheDancingClown:


TheDancingClown:
Script runner not picking up full roll_display - also need to add final frame logic.
Refactoring of Score and Game class to be done.
Tests need to be refactored for edge cases and remove tests that would now be repeated.
Need to complete README.

PR #1234, opened 2020-10-04
Bowling Challenge Part 1 (in Ruby)

veronicavlee89:
First weekend of the bowling challenge.

Part 1 of the bowling challenge covers the logic, developed in Ruby.
Part 2 (next weekend) will involve converting it to Javascript and introducing a UI.


spec/units/frame_spec.rb
```diff
@@ -0,0 +1,221 @@
+require 'frame'
+
+describe Frame do
+
+  subject(:frame) { Frame.new(1) }
+
+  let(:roll_2) { double('roll', pins: 2) }
+  let(:roll_3) { double('roll', pins: 3) }
+  let(:roll_8) { double('roll', pins: 8) }
+  let(:roll_10) { double('roll', pins: 10) }
```
alicelieutier, line 10: I'd put these in the tests themselves - tests don't need to be DRY, they mostly need to be readable.

https://exceptionalcode.wordpress.com/2010/08/23/do-repeat-yourself-in-unit-tests/

alicelieutier: What I'd like is to be able to read one test and understand everything that is happening in it, without having to read the full file.

spec/units/frame_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+
+
+  it 'has a frame number between 1-10' do
+    frame_1 = Frame.new(1)
+    expect(frame_1.number).to eq 1
+
+    frame_10 = Frame.new(10)
+    expect(frame_10.number).to eq 10
+  end
```
alicelieutier, line 19: No need for these tests - this is state not behaviour.

spec/units/frame_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+
+  it 'throws error if frame_num < 1' do
+    expect{ Frame.new(0) }.to raise_error(RuntimeError)
+  end
+
+  it 'throws error if frame_num > 10' do
+    expect{ Frame.new(11) }.to raise_error(RuntimeError)
+  end
```
alicelieutier, line 27: These are behaviour - Good

spec/units/frame_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+
+  describe '#add_roll(roll)' do
+    it 'adds the roll to rolls if roll count < 2 AND new frame score would be <= 10' do
+      frame.add_roll(roll_2)
+      frame.add_roll(roll_8)
+      expect(frame.rolls).to include(roll_2)
+      expect(frame.rolls).to include(roll_8)
+    end
```
alicelieutier, line 35: For this test and the others below, it's useful to ask yourself - how does the user of this class use this class. I would bet that they would use `add_roll` to input things, and then use `score` to get the result. Is that is indeed the case, then these are the methods you should be using for testing. Not testing the values of actual instance variables, which would fix your implementation into something specific.

So something like:
```rb
    it 'adds the roll to rolls if roll count < 2 AND new frame score would be <= 10' do
      frame.add_roll(roll_2)
      frame.add_roll(roll_8)
      expect(frame.score).to eq 10
    end
```

alicelieutier: BTW, the test description seems odd.

spec/units/frame_spec.rb
```diff
@@ -0,0 +1,221 @@
+require 'frame'
+
+describe Frame do
+
+  subject(:frame) { Frame.new(1) }
```
alicelieutier, line 5: This is not best practice. You do not want tests to rely on each other, or in the order they are written in. Otherwise, if you remove one tests, the others would fail. Instead, each test should be independent, and work whatever the order (some testing frameworks actually randomize test order).

spec/units/frame_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+        expect(frame.score).to eq false
+      end
+
+      it 'returns false if 1 roll and total pins < 10' do
+        frame.add_roll(roll_8)
+        expect(frame.score).to eq false
+      end
+    end
+  end
+end
```
alicelieutier, line 221: I did not read the whole file. I think this file is longer than it needs to be, and that is probably because you are testing more internal things than you need to.

spec/units/roll_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'roll'
+
+describe Roll do
+  it 'can have a value of pins between 0-10' do
+    roll_0 = Roll.new(0)
+    expect(roll_0.pins).to eq 0
+
+    roll_10 = Roll.new(10)
+    expect(roll_10.pins).to eq 10
+  end
```
alicelieutier, line 10: No need for these

spec/units/roll_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  it 'throws error if pins < 0' do
+    expect{ Roll.new(-1) }.to raise_error(RuntimeError)
+  end
+
+  it 'throws error if pins > 10' do
+    expect{ Roll.new(11) }.to raise_error(RuntimeError)
+  end
```
alicelieutier, line 18: Good - these are describing the one and only behaviour of this class. I would add two more for the "normal" cases.
```rb
  it 'does not throw an error if pins = 10' do
    expect{ Roll.new(10) }.not_to raise_error
  end
```

spec/units/scorecard_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'scorecard'
+
+describe Scorecard do
+  subject(:scorecard) { Scorecard.new(frame) }
+  let(:frame) { double('frame', :add_roll => [roll], :complete? => false, number: 1) }
+  let(:roll) { double('roll') }
+
+  it 'is initialised with the first frame' do
+    scorecard = Scorecard.new
+    expect(scorecard.frames[0]).to be_an_instance_of(Frame)
```
alicelieutier, line 10: This line tells me this test is not isolated. Unit tests should not depend on any class that is not the class under test.

alicelieutier (commented):
Overall:
* well organised (good SRP)
* DRY
* quite readable

* you have a good command of rspec, however your tests should focus on behaviour more and be isolated. You probably have too many tests ;)

More comments below.

lib/scorecard.rb
```diff
@@ -0,0 +1,30 @@
+require_relative './frame'
+
+class Scorecard
+  attr_reader :frames, :current_frame
+
+  def initialize(frame_1 = Frame.new(1))
+    @frames = [ frame_1 ]
```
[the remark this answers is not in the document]
veronicavlee89: My current_frame logic just takes the last item from the frames array, so wouldn't work if it is initialised with more than one. But I'm going to think how I can change this in my next redo version 👍

PR #1237, opened 2020-10-04
ollie smith bowling challenge ruby

olliesmith3:


veronicavlee89 (commented):
General review comments:
- Code is DRY and follows SRP well. You've split out functions nicely.
- As we talked about, the changes I'd recommend you'd already thought about and highlighted in your readme as future improvements:
-- splitting out classes
--  handling the edge cases e.g. 2 bowls that total > 10 pins
-- consider the scoring logic to apply the bonus rolls to the frame that had the spare or strike, rather than to the following frames
-- could some of these functions be private? Maybe more relevant after the classes are split

PR #1239, opened 2020-10-05
some logic finished, fully tested

LinusMjorn:


LinusMjorn:
logic translated into java, interface and html page still to be built

PR #1254, opened 2020-11-08
Dan Holmes bowling challenge

dan-holmes:


index.html
```diff
@@ -0,0 +1,24 @@
...
+  <script src="src/interface.js"></script>
+  <script src="src/Frame.js"></script>
+  <script src="src/Scorecard.js"></script>
+</head>
+
+<body>
+  <h1>Bowling Scorecard</h1>
+  <div id="scores"></div><br><br>
+  <form id="new-roll">
+    <input type="number" min="0" max="10" id="roll">
```
emilyalice2708, line 19: Such a good idea to use min and max here! Wish I'd thought of it!

src/Frame.js
```diff
@@ -0,0 +1,103 @@
...
+
+  nextRoll2(nextFrame1 = null, nextFrame2 = null) {
+    nextFrame1 = this.handleNulls(nextFrame1)
+    nextFrame2 = this.handleNulls(nextFrame2)
+    if (nextFrame1.isStrike()) {
+      return nextFrame1.nextRoll1(nextFrame2)
+    } else if (this.number == 10) {
+      return this.getBonusRoll(1)
+    } else {
+      return nextFrame1.getRoll(1)
```
emilyalice2708, line 92: I'm still not sure of the best syntax choices, but you could use single line if statements here instead of a large if/else block, because if the first conditional is carried out, the function will return and not move on to the next conditional anyway :)

src/Frame.js
```diff
@@ -0,0 +1,103 @@
...
+    if (this.isStrike()) {
+      return this.nextRoll1(nextFrame1) + this.nextRoll2(nextFrame1, nextFrame2)
+    } else if (this.isSpare()) {
+      return this.nextRoll1(nextFrame1)
+    } else {
+      return 0
+    }
+  }
+
+  nextRoll1(nextFrame1 = null) {
```
emilyalice2708, line 74: Good use of a default parameter value!

src/Frame.js
```diff
@@ -0,0 +1,103 @@
...
+      return 0
+    } else {
+      return this.rolls[n]
+    }
+  }
+
+  getBonusRoll(n) {
+    if (this.bonusRolls[n] == null) {
+      return 0
+    } else {
```
emilyalice2708, line 19: You could use a ternary operator here :)

index.html
benlynch1931: hate to jump on this. With the number, is there a way to store their added number then use THAT to set the next min/max value? so you can't select a total over 10 pins?

src/Frame.js
```diff
@@ -0,0 +1,103 @@
...
+      return this.bonusRolls[n]
+    }
+  }
+
+  getNumber() {
+    return this.number
+  }
+
```
emilyalice2708, line 27: Great Single Responsibility Principle following

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,78 @@
...
+    f1 = new Frame
+    f2 = new Frame
+    f3 = new Frame
+    spyOn(f1, 'setNumber')
+    spyOn(f2, 'setNumber')
+    spyOn(f3, 'setNumber')
+    spyOn(f1, 'score')
+    spyOn(f2, 'score')
+    spyOn(f3, 'score')
+  })
```
emilyalice2708, line 17: Perfect unit testing with spies! Definitely something that I need to work on :')

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,78 @@
...
+    describe("First roll", function () {
+      it("creates a frame, calls roll on it and adds it to the scorecard", function () {
+        scorecard.roll(8)
+        var frame = scorecard.getFrame(0)
+        expect(frame).toBeInstanceOf(Frame)
+        expect(frame.getRoll(0)).toEqual(8)
+      })
+    })
+    describe("Current frame incomplete", function () {
+      it("calls roll on that frame", function () {
```
emilyalice2708, line 51: Love the organisation of your tests, super easy to read and really clear describe/it blocks

src/Scorecard.js
```diff
@@ -0,0 +1,46 @@
+class Scorecard {
+  constructor() {
+    this.frames = []
+  }
+
+  getFrame(n) {
+    return this.frames[n]
+  }
+
+  getLastFrame() {
```
emilyalice2708, line 10: Super clear method names :) and good SRP again!

src/interface.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  function getStandardRollDisplays(frame) {
+    return [frame.getRoll(0), frame.getRoll(1), '']
+  }
+
+  function getTenthRollDisplays(frame) {
+    if (frame.isStrike()) {
+      return [frame.getRoll(0), frame.getBonusRoll(0), frame.getBonusRoll(1)]
```
emilyalice2708, line 35: You could refactor this with a ternary operator or two single line if statements

src/Frame.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  score(nextFrame1 = null, nextFrame2 = null) {
+    return this.pins() + this.bonus(nextFrame1, nextFrame2)
+  }
+
+  bonus(nextFrame1 = null, nextFrame2 = null) {
+    if (this.isStrike()) {
+      return this.nextRoll1(nextFrame1) + this.nextRoll2(nextFrame1, nextFrame2)
```
emilyalice2708, line 66: I'm slightly obsessed with how well you've used the frames class and given it all the frame responsibilities, I honestly feel like you extracted this class so perfectly.

src/Frame.js
```diff
@@ -0,0 +1,103 @@
+class Frame {
+  constructor(rolls = [], number = null) {
+    this.rolls = rolls
+    this.bonusRolls = []
```
emilyalice2708, line 4: Genius idea to use a bonusRolls array

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,78 @@
...
+        scorecard.roll(2)
+        expect(f1.roll).toHaveBeenCalledWith(2)
+      })
+    })
+    describe("Current frame is complete", function () {
+      it("creates a new frame, calls rolls on it and adds it to the scorecard", function () {
+        spyOn(f1, 'isComplete').and.returnValue(true)
+        scorecard.add(f1)
+        scorecard.roll(8)
+        expect(scorecard.getLastFrame()).not.toEqual(f1)
```
emilyalice2708, line 64: Awesome extensive testing

index.html
dan-holmes: Yeah Ben that is a good idea. I didn't get time to do any input validation but this would have been a nice extra flair if I had!

src/Scorecard.js
```diff
@@ -0,0 +1,46 @@
...
+    return this.frames[this.frames.length - 1]
+  }
+
+  getNumFrames() {
+    return this.frames.length
+  }
+
+  roll(n) {
+    this._checkGameOver()
```
emilyalice2708, line 19: You could change this into a return statement to make sure you're not throwing errors for regular behaviour, I'm not sure if there's a way to extract that to another method though. When I tried to do it, it returned 'game over' to the roll method but didn't cause the roll method itself to return. Defo something to look into :)

src/Scorecard.js
```diff
@@ -0,0 +1,46 @@
...
+    var currentFrame = this.frames[this.frames.length - 1]
+    if (this.frames.length == 0 || currentFrame.isComplete()) {
+      currentFrame = new Frame
+      this.add(currentFrame)
+    }
+    currentFrame.roll(n)
+  }
+
+  add(frame) {
+    frame.setNumber(this.frames.length + 1)
```
emilyalice2708, line 29: Again, such good extraction, the frames know what number they are. Beaut.

src/Frame.js
dan-holmes: Yeah I found this bit very tricky. I actually didn't realise you could use a single line if in JS, but I saw you use the ternary operator in your so perhaps I should have googled it haha.

src/Scorecard.js
dan-holmes: I really liked your approach on this, you're right the game being over is not an error

PR #1263, opened 2020-11-09
bowling-challenge-javascript

Oli-Le-Maire:


spec/scoreCountersSpec.js
```diff
@@ -0,0 +1,18 @@
+// 'use strict';
+//
+// describe('ScoreCounters', function(){
+//   var scorecounters;
+//   var frame;
+//
+//   beforeEach(function(){
+//     frame = new Frame();
```
tatsiana-makers, line 8: I wonder if you've seen [this pill](https://github.com/makersacademy/course/blob/master/pills/mocking_in_javascript.md#use-a-fake-version-of-an-object-in-a-test)? Particularly the section "Use a fake version of an object in a test". Can you see how you could use that technique here to create a fake version of a Frame?

src/frame.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  islastFrame(){
+    if (this.frameNumberCounter() !== 10) {
+    return true;
+  } else return false;
+  }
+
+  currentTotalRegularPoints(){
+    return this.rollOne + this.rollTwo;
```
tatsiana-makers, line 31: Where are `this.rollOne` and `this.rollTwo` set?

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
tatsiana-makers, line 1: You can probably remove this code and replace it with other code you might need to define in order to mock the Frame class 🙂

spec/scoreCountersSpec.js
```diff
@@ -0,0 +1,18 @@
...
+//     scorecounters = new ScoreCounters();
+//   });
+//
+//   it('adds the total regular points up', function(){
+//     frame.rollOne(2);
+//     frame.rollTwo(2);
+//     expect(scorecounters.currentTotalRegularPoints()).toEqual(4);
```
tatsiana-makers, line 15: Nice! This code is testing the end user behaviour of the class 🙌  Testing all of the scenarios outlined in the readme would allow you to TDD code to cover the edge cases

tatsiana-makers (commented):
I've left a few comments and a link to a pill about mocking in JS, which I hope you'll find useful. Let me know if you have any further questions, happy to chat more about anything 🙂

PR #1270, opened 2020-12-05
Code with problems!

joshuaretallick:
So versus my Ruby one, this feels very messy.
Couple of questions:

1) Is it ok that all of my methods are declared in score() ?  Is there a way I could have made a bunch of these private too?

2) You will notice on line 17 of the main code I declared self.  Apparently this isn't good convention though, why is that? The reason I did it was I wasn't able to access this.rolls inside the score methods, this was the only way I could get it to work.  Im sure I have made things globally available before just using 'this'?


.idea/bowling-challenge.iml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
TobyRet, line 1: Possibly some of the folders / files in you PR were committed by  accident? e.g '.idea/*`, 'lib/jasmine-3.6.0/*` etc

To ignore stuff you don't intend to commit you can create a `.gitignore` file and add folders or files that you do not want too commit. 

In the root of your folder ...
```
$ touch .gitignore 
$ echo .idea >> .gitignore 
```

src/bowlingGame.js
```diff
@@ -0,0 +1,56 @@
...
+
+class BowlingGame{
+
+    constructor() {
+        this.rolls = [];
+        this.currentRoll = 0;
+    }
+
+    roll(pins) {
+        this.rolls[this.currentRoll++] = pins;
```
TobyRet, line 11: If you do this then the first item in your array will be undefined as you are incrementing the value of currentRoll. An alternative could be: 

```
// populate index 0 in your array on the first roll.
this.rolls[this.currentRoll] = pins; 
// then increment currentRoll
this.currentRoll++;
```

alicelieutier: @TobyRet Actually, that's not true. `a++` returns the value before incrementation and then increments. `++a` increments first then returns.
Try this:
```
let a = 9
let b = a++
console.log(a)
console.log(b)
let c = ++a
console.log(a)
console.log(c)
```

alicelieutier: However, I'd probably put this in two lines to avoid confusion.

src/bowlingGame.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    score() {
+        let result = 0;
+        let rollIndex = 0;
+        let self = this;
```
alicelieutier, line 17: This is actually fine, and a trick that is used in JS regularly (often under the form `let that = this`).
Since some time this year, you can also have actual private methods in JS. See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

alicelieutier: I don't mind the way it's done here, however, In some ways it's even cleaner as you can be sure these methods are only ever used by score.
Another thing you could do instead of `let self = this;` is to have `let rolls = this.rolls` and then refer to `rolls` in the inner functions.

alicelieutier (commented):
This is clean code, and to me, well-written. It would be interesting to see how you'd have to adapt it to build an interface.

alicelieutier:
Also - Your tests are good, all super behavioury.

src/bowlingGame.js
joshuaretallick: Ah great! I couldn't think for the life of me why I couldn't just refer to this.rolls in my score method, as it kept saying undefined.  But I thought it was defined via the method it was in above.

src/bowlingGame.js
TobyRet: Huh, that is surprising. I stand corrected! I had wrongly assumed the behaviour would be the same as Ruby ... I should have known!

```
irb(main):001:0> a = 0
=> 0
irb(main):002:0> array = []
=> []
irb(main):003:0> array[a += 1] = 'foo'
=> "foo"
irb(main):004:0> array
=> [nil, "foo"]

```

Reading up on it further, when you use the increment/decrement operator after the operand, the value will be returned before the operand is increased/decreased (as @alicelieutier mentioned). Incrementing / decrementing the operator before the operand will make the variable increment/decrement before returning.

Fun times indeed.

PR #1272, opened 2020-12-06
Bowling Challenge the 2nd DONE

horthbynorthwest:
Also can be found in the readme

Due to limited time and the fact that I knew I *could* figure out the logic I decided I wanted to focus more on translating into JavaScript to make sure I felt solid with where I was going into further JavaScript.

My solution doesn't cover much in the way of edge cases nor does it stop a game at 10 frames I feel pretty happy with where I got to with my time limitations.

I ended up adding in ESLint at the end and was very happy to find only one warning, which given the fact that it was simply my class was defined and not called again was not really something I could fix.


spec/BowlingGameSpec.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+describe('bowling', () => {
+  let game
+  beforeEach( () => {
+    game = new BowlingGame();
+  });
+
+  it('can roll a gutter game', () => {
```
sedwards93, line 9: Really nice clear and concise testing here

src/BowlingGame.js
```diff
@@ -0,0 +1,48 @@
+class BowlingGame{
+
```
sedwards93, line 2: Such clean code! I love it.

src/BowlingGame.js
```diff
@@ -0,0 +1,48 @@
...
+    function isStrike() {
+      return game.rolls[rollIndex] === 10;
+    }
+
+    function isSpare() {
+      return game.rolls[rollIndex] + game.rolls[rollIndex+1] === 10;
+    }
+
+    function getStrikeScore () {
```
sedwards93, line 36: Could make some of these _private, but honestly what does _this even do anyway?

.eslintrc.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+    "env": {
```
sedwards93, line 2: Love a bit of linting

PR #1275, opened 2020-12-06
stringiest Bowling Challenge - JavaScript

stringiest:


interface.js
```diff
@@ -0,0 +1,29 @@
...
+
+  function updateFrameNumberInterface() {
+    var frameNumber = game.frameNumber;
+    if (frameNumber < 11) {
+      $('#frame').text(Math.floor(game.frameNumber));
+    } else {
+      $('#frame').text(10);
+    }
+
+  }
```
alicelieutier, line 25: Nice display update function

interface.js
```diff
@@ -0,0 +1,29 @@
...
+
+  updateFrameNumberInterface(1); // sets default frame number as 1
```
alicelieutier, line 27: You are never reading this argument.

spec/bowlingGameSpec.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+    expect(game.score()).toEqual (300);
+  });
+
+  it("can roll all spares", function() {
+    for (var i = 0; i < 21; i++) {
+      game.roll(5);
+    }
+    expect(game.score()).toEqual (150);
+  });
```
alicelieutier, line 53: Good tests overall. You could have a few more tests around the endgame. strike + spare in the last frame, mixed rolls - a typical game.

src/BowlingGame.js
```diff
@@ -0,0 +1,73 @@
...
+      } else {
+        result += this._frameScore(rollIndex);
+        rollIndex += 2;
+      }
+    }
+    console.log(result);
+    return result;
+  }
+
+  updateFrameNumber() {
```
alicelieutier, line 34: So this is for the interface only? Given that this calculates the frame number from scratch every time it's called, maybe it could not use an instance variable, and instead just return the frameNumber.  So in the interface, you would call `game.frameNumber()` and it would be this function, that returns the variable. Otherwise, you need to be sure this function here has been called by the `roll` method before you read the value of the variable in the interface.

alicelieutier (commented):
Clean code, good test. I have a possible improvement that I added as a comment.

src/BowlingGame.js
stringiest: Thanks!  Have amended.

PR #1278, opened 2020-12-07
Bowling Challenge in JS - first attempt

CodeZeus-dev:
Game Logic - Partially Implemented (10th frame still pending)
UI - Implemented
jQuery - Pending (for changing the table values based on user input)


js/interface.js
```diff
@@ -0,0 +1,10 @@
+$(document).ready(function() {
+
+  let game = new Game();
+  let frameIndex;
+
+  $('#frame-choice').change(function() {
+    $('#scoreCard #scoreCardBody #kdp11').html("Hey there1");
```
AdamusBG, line 7: hey1 👋🏼

public/style.css
```diff
@@ -0,0 +1,110 @@
...
+.nav-masthead .nav-link + .nav-link {
+    margin-left: 1rem;
+}
+
+.nav-masthead .active {
+    color: #fff;
+    border-bottom-color: #fff;
+}
+
+@media (min-width: 48em) {
```
AdamusBG, line 83: loving the media query!

src/Frame.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+function Frame() {
+  this.strike = false;
+  this.spare = false;
+  this.frameScore = {roll1: 0, roll2: 0};
+  this.skipSecondRoll = false;
+}
+
+Frame.prototype = {
```
AdamusBG, line 10: Nice use of prototype here!

src/Frame.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+function Frame() {
+  this.strike = false;
+  this.spare = false;
+  this.frameScore = {roll1: 0, roll2: 0};
```
AdamusBG, line 6: Suggest considering changing this to an array, it's more optimized for performance and more adaptable (e.g. when you need to account for three rolls in the final frame)

src/Frame.js
```diff
@@ -0,0 +1,66 @@
...
+  this.skipSecondRoll = false;
```
AdamusBG, line 7: this.strike and this.skipSecondRoll are always gonna be the same, could suggest having only one of them or removing beth and just having a method which returns whether its a strike or not

spec/unit/FrameSpec.js
```diff
@@ -0,0 +1,130 @@
+'use strict';
+
+describe('Frame', function() {
```
AdamusBG, line 3: Nice testing here, seems to cover every condition/program state.

AdamusBG (commented):
Nice work constantine, you are not far from having the final version implemented!

PR #1288, opened 2021-01-17
Bowling game in JS

AmanTank187:
Converted my ruby bowling game into JS. I'm going improve on this further and add html.


jasmine-standalone-3.6.0 2/src/bowling.js
```diff
@@ -0,0 +1,58 @@
...
+        score += frameScore;
+      }
+
+
+      rollIndex += 2;
+    }
+    return score;
+  }
+
+isSpare(frameScore){
```
PiperS52, line 41: I think indentation for lines 41-55 might need adjusting (to the right)

AmanTank187: Thanks , I'll do it

jasmine-standalone-3.6.0 2/src/bowling.js
```diff
@@ -0,0 +1,58 @@
...
+
+  roll(pins){
+    this._rolls.push(pins)
+  }
+
+  reset(){
+    this._rolls = [];
+  }
+
+    get score(){
```
PiperS52, line 16: As it is, this method doesn't quite work (needs to be one word). Adjusting to 'getScore' works ok :)

PiperS52:
Nice use of different methods and tests in JS

PR #1290, opened 2021-01-17
Natasha's attempt at JS bowling

natashamcintyre:


index.html
```diff
@@ -0,0 +1,164 @@
...
+    <script src="jquery.js"></script>
+    <script src="interface.js"></script>
+    <header>
+      Picture of 10 pin bowling
+      <button id="reset">Start Again</button>
+    </header>
+
+    <div class="frame_results">
+
+      <div class="frame">
```
tatsiana-makers, line 21: Nice grouping of the frames. Makes the html file easy to read

index.html
```diff
@@ -0,0 +1,164 @@
...
+          <option value="9">9 pins</option>
+          <option value="10">10 pins</option>
+      </select>
+    </div>
+
+    <div class="gameover">
+      <h2 id="gameover"></h2>
+    </div>
+  </body>
+</html>
```
tatsiana-makers, line 164: This looks good to me (:

index.css
```diff
@@ -0,0 +1,60 @@
...
+
+.frame_results {
+  padding: 20px;
+  height: 200px;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-around;
+}
+
+.frame {
```
tatsiana-makers, line 13: Nice use of class selectors to style all the frames!

spec/featureSpec.js
```diff
@@ -0,0 +1,108 @@
...
+      game.roll(4);
+      expect(game.gameover()).toEqual(7);
+    });
+  });
+
+  describe('multiple frames, no spares/strikes', function() {
+    it('resets frame after every 2 rolls', function() {
+      game.roll(3);
+      game.roll(4);
+      expect(game.currentFrame.pins).toEqual([]);
```
tatsiana-makers, line 35: This assertion is testing the state of the program rather than the behaviour for the end user: https://github.com/makersacademy/skills-workshops/blob/master/practicals/testing/behaviour_not_state.md

src/Frame.js
```diff
@@ -0,0 +1,35 @@
...
+
+  result() {
+    if (this._isaStrike()) {
+      return 10
+    } else {
+      return this.pins[0] + this.pins[1]
+    }
+  }
+
+  _isaStrike() {
```
tatsiana-makers, line 23: Nice use of private methods!

From [this resource on OOP](http://roboprog010.github.io/posts/technical/week5.html) :
>Encapsulation just means that all of the implementation details of the object being modeled are hidden and only the methods and properties required to use the object are exposed

By making any methods that aren't required to use the object private, you are increasing encapsulation and reducing the risk that another object will mess up the state of this object

src/Game.js
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor() {
+    this.currentFrame = new Frame();
+    this.scorer = new Scorer();
+  }
+
+  // User adds a roll to their scorecard
+  roll(pins) {
+    if (pins < 0 || pins > 10) {
+      throw "Please enter number between 0 and 10";
```
tatsiana-makers, line 11: 0 and 10 appear multiple times across the code base. It would be cool to get rid of the [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) and replace them with clearly-named constants

src/Game.js
```diff
@@ -0,0 +1,59 @@
...
+
+  _isAtStart() {
+    return this.frameCount() === 1
+  }
+
+  _isOver() {
+    return this.scoreCount() >= 10
+  }
+
+}
```
tatsiana-makers, line 59: Really clearly named methods, nice!

interface.js
```diff
@@ -0,0 +1,44 @@
+$(document).ready(function() {
+  var game = new Game();
+
+  $('.User_Roll').change(function() {
+    let pins = parseInt($('.User_Roll').val());
+    $(`#f${game.frameCount()}r${game.rollCount()}`).text(pins);
+    if (game.rollCount() === 1 && pins === 10) {
+      $(`#f${game.frameCount()}r2`).text("X");
+    }
```
tatsiana-makers, line 9: I think extracting these lines of code into a separate clearly-named function would make this method a bit more readable

tatsiana-makers (commented):
Thanks for asking for a review! Nice work on this challenge, your code is overall readable with clearly defined method names! I've left a few other comments, do reach out if you want to chat through any of them 🙂

src/Frame.js
natashamcintyre: I'm not sure what you mean here. Do you mean only methods that need to be called directly by the user should be public (I think in this challenge that would be only game.roll(n)) or should this also apply to methods that the interface needs to call on (eg game.isOver() or game.framecount() which returns the number of the frame currently being played)?

index.html
natashamcintyre: That's great, thank you. I was worried about it not being DRY because of the repeated frames I guess this syntax is necessary :)

spec/featureSpec.js
natashamcintyre: Thanks for the link to the practical, this is something I've not really focused on yet so I will make it my focus for upcoming projects.

src/Game.js
natashamcintyre: Yes I should have seen this! Thanks for pointing it out.

src/Frame.js
natashamcintyre: Having had a look at the OOP I think it should just be game.roll...

tatsiana-makers: Private methods should not be called from outside of a class. Does this guide you to the answer to the question?

natashamcintyre: So private methods should not be called from anything outside of the model? (I'm pretty sure you don't mean they should not be called from outside the class in which they themselves are defined...)
So this has led me to another question, is the interface.js file considered to be a part of the model, or is it the controller? I think it's the latter?

tatsiana-makers: [This page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) is some documentation about private class fields in JS. Just like in Ruby, you shouldn't call private class methods from outside of the class. 


For example, in the Thermostat example, we don't call the private method [`_isMinimumTemperature`](https://github.com/makersacademy/thermostat-example/blob/bfd88a9e50687aa462b7fabb0c777ee6877bafcc/src/Thermostat.js#L54) from the [thermostat-example interface](https://github.com/makersacademy/thermostat-example/blob/master/src/interface.js), we only call the public methods in the Thermostat class

PR #1292, opened 2021-01-17
Weekend Challenge W6: Bowling-JavaScript

chrismabdo:


src/Game.js
```diff
@@ -0,0 +1,51 @@
...
+    constructor() {
+        this.scorecard = [];
+    }
+
+    addSpareBonus() {
+        this.scorecard.length < 4 ? this.scorecard.push(this.first + this.first + this.second) : this.scorecard.push(this.scorecard[this.scorecard.length-3] + this.first + this.first + this.second);
+    }
+
+    addStrikeBonus() {
+        this.scorecard.length < 4 ? this.scorecard.push(this.first + this.first + this.second + this.second) : this.scorecard.push(this.scorecard[this.scorecard.length-3] + this.first + this.first + this.second + this.second);
```
WillDixon93, line 14: could this be refactored a bit more?

spec/GameSpec.js
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    it('will return a message if both rolls go in the gutter', function() {
+        game.roll(5, 4)
+        expect(function() {game.roll(0, 0)}).toThrow("Double Gutter! Bad Luck");
+    });
+
+
+    it('will end when the tenth round ends', function() {
+        game.roll(6, 4)
```
WillDixon93, line 62: could you do a loop here so you don't have to write out game.roll 10 times? (not sure if it would actually work but worth a thought)

WillDixon93: i.e. (i = 0; i < 20; i++)

WillDixon93 (commented):
Good stuff man! Just a couple of things might be worth a look at.

PR #1293, opened 2021-01-17
Ollie Norman - Bowling Challenge

ollienorman:
* Code doesn't have 10th frame functionality (handling extra rolls correctly/allowing the correct number of maximum rolls). Had to change the logic as was using input from .gets in Ruby, and absolutely coded myself in knots trying to implement the new logic so had to roll it back.
* Generally not very happy with the code here, needs a full pass at refactoring, lots of obvious things to tidy up. Did feel more comfortable writing JS during this challenge though.


Ben-glitch-cloud:
Nice code for the weekend challenge. Over all I don't have any issues with the code.

src/scorecard.js
```diff
@@ -0,0 +1,90 @@
...
+
+  applyBonus() {
+    if(this._strikesSpares[this._frame -1] === "strike" && this._strikesSpares[this._frame -2] === "strike"){
+      this._frameScores[this._frame -1] += this.sum(this._pinsKnocked[this._frame])
+      this._frameScores[this._frame -2] += this._pinsKnocked[this._frame][0]
+    }
+    else if(this._strikesSpares[this._frame -1] === "strike") {
+      this._frameScores[this._frame -1] += this.sum(this._pinsKnocked[this._frame])
+    }
+    else if(this._strikesSpares[this._frame -1] === "spare") {
```
Ben-glitch-cloud, line 37: this last else if, could have just been else

src/scorecard.js
```diff
@@ -0,0 +1,90 @@
...
+  constructor() {
+    this._pinsKnocked = []
+    this._frame = 0
+    this._roll = 0
+    this._frameScores = []
+    this._strikesSpares = {}
+    this._maxFrames = 9
+  };
+
+  inputRoll(number) {
```
Ben-glitch-cloud, line 11: Could have a return statement to the user/player, about the number pins they have knocked down.

spec/scorecardSpec.js
```diff
@@ -0,0 +1,95 @@
+describe('Scorecard', function(){
```
alicelieutier, line 1: Nice tests!

src/scorecard.js
```diff
@@ -0,0 +1,90 @@
...
+    if(this._frameNotExist) {
```
alicelieutier, line 12: You are not calling the method `this._frameNotExist` here, so this condition is in fact testing whether that property is defined, which it is, so it's always true, which means you are always pushing []... Weird 🤔 

Anyway - a good one for getting visibility.

src/scorecard.js
```diff
@@ -0,0 +1,90 @@
+class Scorecard {
+  constructor() {
+    this._pinsKnocked = []
+    this._frame = 0
+    this._roll = 0
```
alicelieutier, line 5: The fact that you need to keep track of which roll in the frame and which frame in the game point to me towards the idea of extracting Frame functionality into its own class, that could keep its own pins and scores.

Then the game is responsible for creating a new frame when needed. The frame could tell it if it's finished, if it still needs some bonuses, etc.

alicelieutier (commented):
Overall your code is really good - well organised and readable.
I hinted at a possible refactor below.

PR #1294, opened 2021-01-17
Bowling Challenge - Javascript

kikidawson:
Couldn't take the tenth frame logic out into separate function because then return stopped working and all my tests failed.
Haven't added front end.


spec/ScorecardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+    scorecard.roll(4)
+    scorecard.roll(10)
+    scorecard.roll(2)
+    scorecard.roll(4)
+
+    expect(scorecard.getTotal()).toEqual(119)
+  })
+
+  it('gives 0 for gutterball', function() {
+    scorecard.roll(0)
```
xavierloos, line 94: Perhaps you could use a loop here!

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+    scorecard.roll(10)
+    scorecard.roll(10)
+    scorecard.roll(2)
+    scorecard.roll(6)
+
+    expect(scorecard.getTotal()).toEqual(48)
+  })
+
+  it('works with ten frames', function() {
+    scorecard.roll(1)
```
xavierloos, line 71: You can check which ones are repetitive and put it in a (for) loop, I can see a lot of `4`s.

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+    scorecard.roll(0)
+    scorecard.roll(0)
+    scorecard.roll(0)
+    scorecard.roll(0)
+
+    expect(scorecard.getTotal()).toEqual(0)
+  })
+
+  it('tenth frame bonus - test with spare', function() {
+    scorecard.roll(1)
```
xavierloos, line 119: Same thing as above!

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,158 @@
...
+    scorecard.roll(10)
+    scorecard.roll(2)
+    scorecard.roll(8)
+    scorecard.roll(6)
+
+    expect(scorecard.getTotal()).toEqual(133)
+  })
+
+  it('tests a perfect game', function() {
+    scorecard.roll(10)
```
xavierloos, line 143: Try to make you test shorter.

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,158 @@
+'use strict';
+
+describe('Scorecard', function() {
+  var scorecard;
+
+  beforeEach(function() {
+    scorecard = new Scorecard();
+  });
+
+  it('starts with empty scorecard', function() {
```
xavierloos, line 10: You could scoop your code in another `describe`, I can see you are testing for `getFrames()`, you could have all those tests for this method in an another describe block `describe("getFrames()",()=>{all the it here})`, and do the same for the repetitive methods.

src/Scorecard.js
```diff
@@ -0,0 +1,193 @@
...
+        return;
+      }
+    }
+    if (this._isFirstRollAndNonStrike()) {
+      return;
+    }
+    this._firstFrameLogic()
+  }
+
+  _firstFrameLogic() {
```
xavierloos, line 43: Extra points for using "Private" methods

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,158 @@
+'use strict';
+
+describe('Scorecard', function() {
```
xavierloos, line 3: I think you are not testing for the private methods!

src/Scorecard.js
```diff
@@ -0,0 +1,193 @@
+'use strict';
+
+class Scorecard {
```
xavierloos, line 3: Make your methods/variables shorter _(and easy to read/remember)_ but always meaning of what it does.

xavierloos (commented):
Overall I think your code is immaculate, but you could make shorter, I think there are some things you could test for _(false statements & private)_. Make usage of scoops and loops, your code will cleaner.  Make more `commit`s when you do any change, one test and its solution, `commit` it.
Excelent**E**!

spec/ScorecardSpec.js
xavierloos, line 3: instead of _function(){}_ use **`()=>{}`**

src/Scorecard.js
```diff
@@ -0,0 +1,193 @@
...
+      } else if (this._isSecondRoll()) {
+        this._addFramesAndBonusTotal()
+        return;
+      } else {
+        this._addFramesAndBonusTotal()
+        this._addThirdRollToTotal()
+        return;
+      }
+    }
+    if (this._isFirstRollAndNonStrike()) {
```
xavierloos, line 37: if you are returning one line of code in an IF there is no need of using `{ }`

src/Scorecard.js
```diff
@@ -0,0 +1,193 @@
+'use strict';
+
+class Scorecard {
+  constructor() {
+    this.frames = [];
```
xavierloos, line 5: No needs of `;` in JS

PR #1295, opened 2021-01-18
Bowling Challenge 

Ben-glitch-cloud:
I was hopping to get a little further in the challenge than I did. I got stuck on the 10th bowl trying to great an if statement that would if the player an extra bowl if they got a stick or spare.  

Over all I probably should have started again with the if statement I had created.


src/bowling.js
```diff
@@ -0,0 +1,110 @@
...
+    this.number_rounds = 0
+
+    } 
+
+    gameRound(num_one, num_two) {    
+        while (this.number_rounds <= 10) {   
+            this.number_rounds ++
+            if (this.number_rounds <= 10) { 
+                if(num_one + num_two > 10) { 
+                    return "Thats not a realy score, it must be below 10"
```
ollienorman, line 19: Good to have a guard clause for this, I didn't think to do that - may be worth having it as an error?

src/bowling.js
```diff
@@ -0,0 +1,110 @@
...
+                    } 
+    
+            } 
+            
+        }
+     
+    
+    game_score() {   
+
+        this.x = []
```
ollienorman, line 50: Are this.x, this.s and this.a used in the code still? I couldn't find them

src/bowling.js
```diff
@@ -0,0 +1,110 @@
...
+
+        this.round.forEach((num, index) => num === "X" ? score.push(index) : null) 
+        
+        score.forEach((entry) => Number.isInteger(entry) ? this.round.push(this.round[entry + 1], this.round[entry + 2]) : null)
+
+        this.round.forEach((num, index) => num === "S" ? this.number.push(index) : null)
+        
+        this.number.forEach((entry) => Number.isInteger(entry) ? this.round.push(this.round[this.s + 1]) : null)
+
+        this.score_two = (this.round.filter(a => a !== 'X'))
```
ollienorman, line 66: Cool use of the .filter function - I need to look into this!

ollienorman: Also cool use of the ES6 arrow function syntax :)

PR #1297, opened 2021-01-18
Bowling Challenge - JS (Ara)

aracho1:
Check out on HEROKU:
https://bowling-scorecard-ara.herokuapp.com/

Approach & challenges in README

Files for review: game.js, index.js


src/Game.js
```diff
@@ -0,0 +1,136 @@
...
+
+  finalRound(frame){
+    return frame === this.MAX_FRAME - 1;
+  }
+
+  totalScore() {
+    return this.frameScores[this.MAX_FRAME - 1];
+  }
+
+  play() {
```
lisabardelli, line 112: All your methods have not more than 5 lines. it might be worth seeing if you can reduce the number of line of the play  method

stylesheet.css
```diff
@@ -0,0 +1,56 @@
+body {
+  text-align: center;
+  margin-top: 50px;
```
lisabardelli, line 3: your interface looks really nice :)

spec/GameSpec.js
```diff
@@ -0,0 +1,98 @@
...
+      game.frameScores = [9];
+      game.updateScores(1)
+      game.updatePreviousFrame(1);
+      expect(game.frameScores).toEqual([9,19]);
+    });
+
+  describe("calculate strike and spare scores", () => {
+    it("should calculate the final score for a strike", () => {
+      game.frameScores = [10, 19, 20, 34, 49, 55, 65, 74, 94, 101]
+      expect(game.totalScore()).toEqual(101);
```
lisabardelli, line 94: that's good that you have tested all the possible scenarios

index.js
```diff
@@ -0,0 +1,148 @@
...
+      round++; 
+      roll = 0; 
+      
+      return;
+    } else if (game.frames[round] === undefined) { 
+      game.frames[round] = [score]; 
+      roll++; 
+    } else {
+      if (!game.finalRound(round) && (game.frames[round][0] + score > 10)){
+        alert("Invalid pin. Please try again");
```
lisabardelli, line 96: you can replace  **alert("Invalid pin. Please try again");** with **alert(err.message)**;  so you don't have to copy the same message twice
try { 
  // code...
} catch (err) {
alert(err.message)
} 
https://javascript.info/try-catch#the-try-catch-syntax

lisabardelli (commented):
Your code is clear and readable and your interface is amazing. I can only think that you can split the code in more than 1 class. Well done!

index.js
```diff
@@ -0,0 +1,148 @@
...
+  document.getElementById('score-ten').onclick = () => {
+    updateFrame(10);
+    updateScoreBox();
+    updatePinBox();
+  };
+  document.getElementById('reset').onclick = () => {
+    window.location.reload();
+  }
+
+
```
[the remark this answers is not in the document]
alicelieutier: If you take a step back, maybe a good refactor could be to have two main methods. One that updates the whole interface based on the game, and one that is attached to the number buttons and enters a roll.
The rest of the logic, in-between, should entirely be in the game.
So the interface does not need to know if something is a spare or a strike, etc, the game can figure that out.

With this plan, what you need to know is:
- how can I tell the game what I've rolled. Maybe an `inputRoll` method that takes a number
- what does the game need to return so that I can fill in the interface? Probably the pins + scores for each frame. So maybe you can have the game have a method named `dataToDisplay` that could return something like `[{ "roll1" : 9,  "roll2": 1, "score":  12}, { "roll1" : 1,  "roll2": 1, "score":  14}]` ... That would be calculated on the game class based on the inputs.

Then the interface only job is to link the button click to the right inputRoll, and then use the data to display to add the correct text to the HTML.

Let me know if that makes sense.


Then you can develop the game class, using tests that s

alicelieutier (commented):
Overall, this is a good attempt (and I love the styling). You can refactor further, I've added a number of comments to guide you.

Also, I found a bug:
If I roll a 0, and then roll a 10 on one frame, it should be a spare, but it shows like a strike.

PR #1299, opened 2021-01-18
translated, need to write tests

ZeenLamDev:
In complete

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Bowling Scorecard</title>
+</head>
+<body>
+  <div id='container'>
+    <h1 id='main-header'>Bowling Scorecard</h1>
+  </div>
+  
+</body>
+</html>
--- a/src/bowling.js
+++ b/src/bowling.js
@@ -1,0 +1,107 @@
+class Bowling {
+  constructor() {
+    this.bonus = 0;
+    this.total = 0;
+    this.frame = 1;
+    this.currentFrame = [];
+  }
+
+  addRoll(pins) {
+    var score = 0;
+    if(!this.isGameContinues()) throw new Error('The game is over');
+    this.isBonus() ? score = this.multiplier(pins) : score = pins;
+    this.applyScore(pins, score);
+    if(this.isEndFrame()) this.endFrame();
+  }
+
+  endFrame() {
+    if(this.isSpare()) this.bonus++;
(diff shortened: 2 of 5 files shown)


public/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Bowling Scorecard</title>
```
lukewickens1989, line 6: Love the title - explains exactly what it is going to be on the page.

public/index.html
```diff
@@ -0,0 +1,14 @@
...
+</head>
+<body>
+  <div id='container'>
```
lukewickens1989, line 9: Like the use of a container div. Its good to encase all your code in a div so that it can be manipulated easier.

src/bowling.js
```diff
@@ -0,0 +1,107 @@
+class Bowling {
+  constructor() {
+    this.bonus = 0;
+    this.total = 0;
+    this.frame = 1;
```
lukewickens1989, line 5: Nice setting this variable to 1. Makes thinking about the process easier.

src/bowling.js
```diff
@@ -0,0 +1,107 @@
...
+
+  isDoubleBonus() {
+    if (this.bonus >= 3) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+   isGameContinues() {
```
lukewickens1989, line 88: I like the use of this conditional. Something that I dont think I implemented in my code :) nice work

src/bowling.js
```diff
@@ -0,0 +1,107 @@
...
+    if(this.frame < 10 || (this.frame === 10 && this.isTenthFrameException())) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  isTenthFrameException() {
```
lukewickens1989, line 96: I like how you've encapsulated this in its only function.

src/bowling.js
```diff
@@ -0,0 +1,107 @@
...
+    if(this.currentFrame.length < 2 || (this.currentFrameSum() >= 10 && this.currentFrame.length < 3)) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  currentFrameSum() {
```
lukewickens1989, line 104: Nice use of the reduce method.

lukewickens1989 (commented):
Nice easily readable and encapsulated code. Real evidence of using SRP in your function design. I like the use of all of the predicate functions rather than having too much complex logic to do a simple thing. Obviously there are no tests, so TDD was not used, which would be something that could be improved on however all in all some really nicely written and well structured class design.

PR #1300, opened 2021-01-18
Bowling challenge js in progress

sujee09:
This is my attempt at the bowling challenge in Javascript.


spec/BowlingSpec.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+describe ('The Bowling Game', function() {
+
+  var game;
+
+  beforeEach(function () {
+    game - new BowlingGame();
```
calavell, line 8: should this be "=" instead of "-"

spec/BowlingSpec.js
```diff
@@ -0,0 +1,29 @@
...
+
+  it('can roll a gutter game', function () {
+      rollMany(1, 20);
+      expect(game.score()).toBe(0);
+  });
+
+  it('can roll all ones', function () {
+    rollMany(1, 20);
+    expect(game.score()).toBe(20);
+  });
```
calavell, line 19: These 2 tests have the same conditions (rollMany(1, 20)) but the expected outcomes are different. Should line 12 be "rollMany(0, 20);"?

src/Bowling.js
```diff
@@ -0,0 +1,17 @@
...
+ this.rolls.push(pins);
+};
+
+BowlingGame.prototype.score = function () {
+  var result = 0;
+  for (var i = 0; i < 20; i++) {
+    result += this.rolls[i];
+  }
+  return result;
+};
```
calavell, line 17: Will this method work if less than 20 rolls are played?

PR #1305, opened 2021-02-14
Bowling Challenge

mscwilson:


spec/PlayerSpec.js
```diff
@@ -0,0 +1,58 @@
+describe("Player", function() {
```
KarstenFinlay, line 1: I am going to presume that this is the left over jasmine scripts from the standalone files hahaha.

index.html
```diff
@@ -0,0 +1,81 @@
...
+        <th colspan="2">4</th>
+        <th colspan="2">5</th>
+        <th colspan="2">6</th>
+        <th colspan="2">7</th>
+        <th colspan="2">8</th>
+        <th colspan="2">9</th>
+        <th colspan="3">10</th>
+      </tr>
+<!--       <tr>
+        <td><span id="pins1-1"></span></td>
```
KarstenFinlay, line 36: What was the problem that occurred here?

spec/FrameSpec.js
```diff
@@ -0,0 +1,73 @@
+describe("Frame", function() {
```
KarstenFinlay, line 1: Really solid testing overall! Though maybe the use of Jasmine spies could of helped, however, that might be something we can discuss, as I am not fully sure on how to use them myself!

src/BowlingJQuery.js
```diff
@@ -0,0 +1,31 @@
...
+    frameScores()
+  })
+
+  function frameScores() {
+    for (let i = 1; i < 11; i++) {
+      $(`#score${i}`).text(game.cumulScores[`frame_${i}`]);
+    }
+  }
+
+  function isValidInput(input) {
```
KarstenFinlay, line 24: That's a nice guard right there, really cool.

KarstenFinlay (commented):
Really great code, very well thought out!

Shame about the interface, but other than that, really nice job. Feels like a very smooth transition from Ruby to JS

PR #1309, opened 2021-02-15
Partial solution

OscarB89:
Needs refactoring


spec/gameSpec.js
```diff
@@ -0,0 +1,43 @@
...
+        game.roll(0);
+      }
+      expect(game.score()).toBe(18);
+    });
+
+    it("can return a strike", function () {
+      game.roll(10);
+      game.roll(5);
+      game.roll(4);
+      for (var i = 0; i < 16; i++) {
```
twigz826, line 37: I think you could make this DRYer by having a helper function called rollMany which contains the for loop, something like this:
```
rollMany = (pins,rolls) => {
    for (let i = 0; i < rolls; i++) {
      game.roll(pins);
    }
  }
```

src/game.js
```diff
@@ -0,0 +1,30 @@
...
+  score() {
+    var result = 0;
+    var index = 0;
+
+    for (var i = 0; i < 10; i++) {
+      if (this._rolls[index] === 10) {
+        result +=
+          this._rolls[index] + this._rolls[index + 1] + this._rolls[index + 2];
+        index += 1;
+      } else if (this._rolls[index] + this._rolls[index + 1] === 10) {
```
twigz826, line 19: I think the if else statement could be clearer if you had separate functions for isSpare, isStrike, etc.

index.html
```diff
@@ -0,0 +1,33 @@
...
+    </div>
+    <div id="scores">
+      <button id="gutter">Gutter</button>
+      Pins: <input id="pins" type="number" min="1" max="9" />
+      <button id="spare">Gutter</button>
+      <button id="spare">Spare</button>
+      <button id="strike">Strike</button>
+    </div>
+  </body>
+</html>
```
twigz826, line 33: This looks good, does it link up with game.js?

twigz826 (commented):
It looks good so far, just a few comments on some possible refactoring.

PR #1310, opened 2021-02-15
bowling js challenge

jshields123:


HamishArro (commented):
nice quality code, short and simple :)

PR #1312, opened 2021-02-15
Sean's Bowling challenge

bear99a9:
backend works fine didnt even really get started on UI


spec/gameSpec.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    it('can work out a game with different roll amounts', () => {
+      bowling.roll(1);
+      bowling.roll(6);
+      bowling.roll(2);
+      bowling.roll(3);
+      bowling.roll(4);
+      bowling.roll(5);
+      rollMultiple(0, 14);
```
bateso88, line 21: Why are you rolling 0 14 times?

spec/gameSpec.js
```diff
@@ -0,0 +1,94 @@
...
+      bowling.roll(1);
+      expect(bowling.currentScore()).toEqual(29);
+    });
+
+  });
+
+  function rollMultiple(pins, rolls) {
+    for (let i = 0; i < rolls; i++) {
+      bowling.roll(pins);
+    }
```
bateso88, line 92: Nice!

src/game.js
```diff
@@ -0,0 +1,67 @@
...
+        index += 1;
+      }
+    }
+
+    return result;
+  };
+
+  isSpare = frameScore => {
+    return frameScore === this.MAX_SCORE;
+  };
```
bateso88, line 50: would this return true for a strike as well? Could be an issue when calculating scores

bateso88 (commented):
Nice and concise. Amazed you managed to keep your number of lines of code so small! Looks like maybe there was an issue with calculating score() / currentScore() - maybe these could be combined? Very good code, clear and easy to read

PR #1318, opened 2021-02-15
Bowling - Challenge - JS- Glykeria

GlykeriaStr:
I still need to work on the index and interface.


src/Game.js
```diff
@@ -15,16 +15,11 @@ class Game{
   }
 
   run(roll){
```
ddottyler, line 17: could have 'run' have been named anything different?

ddottyler (commented):
Great effort and lots of detail and visibility on the code which is great for putting together an interface (I'm guessing?) 

Testing seems really thorough and helps to break down what the code in src is actually doing as there is otherwise quite a lot going on.

Good idea on breaking the whole thing down into multiple classes and fair play for mocking/isolating the tests!

spec/GameSpec.js
```diff
@@ -0,0 +1,72 @@
...
+
+  describe('#AddStrikePoints', function() {
+    it('adds strike points to the previous frame', function() {
+      game.run(10);
+      game.run(3);
+      game.run(3);
+      expect(game.FrameBeforeLast().frame_score).toEqual(16);
+    })
+
+    it('does not add strike points to the previous frame', function() {
```
siellsiell, line 38: why not? The two tests sound contradictory for someone who doesn't know in which cases this should happen. Describing the scenario in more detail could help here.

same for similar cases below

spec/LastFrameSpec.js
```diff
@@ -0,0 +1,56 @@
...
+
+    it('tells us the last frame is NOT finished', function() {
+      frame10.AddRolls(5);
+      frame10.AddRolls(5);
+      expect(frame10.isFinished()).toBe(false);
+    })
+    it('tells us the last frame is NOT finished', function() {
+      frame10.AddRolls(10);
+      frame10.AddRolls(5);
+      expect(frame10.isFinished()).toBe(false);
```
siellsiell, line 53: Same comment here regarding test case descriptions (*why* is the last frame not finished?) -- if these tests fail it's not easy to understand what went wrong.

spec/GameSpec.js
```diff
@@ -0,0 +1,72 @@
+"use strict";
+
+describe('Game', function() {
+  let game, frame, frame_1;
+
+  beforeEach(function(){
+    game = new Game();
+    frame = jasmine.createSpyObj('Frame',['frame_Score', 'AddRolls', 'isStrike'])
+    frame_1 = jasmine.createSpyObj('Frame',['frame_Score', 'AddRolls', 'isStrike'])
```
siellsiell, line 9: It was a good idea to use spies here but I don't see you using these anywhere?

siellsiell (commented):
This is a nice solution! I like that you've created separate classes to encapsulate behaviour.

You could do some work on making your code easier to read and more self-documenting, especially by simplifying complicated `ifs` and improving the description of your test cases.

You also don't have a README explaining how to install and run your code so I haven't tested it in the browser.

Some more minor comments:
- Using a linter or code formatter would be good to make your style more consistent (e.g. space around brackets `Game{` should be `Game {` for example)
- I'm not sure why you have all those jQuery files in your PR? Using jQuery should only require including a single file https://www.w3schools.com/jquery/jquery_get_started.asp

PR #1319, opened 2021-02-15
DT javabowling challenge

ddottyler:


src/javabowling.js
```diff
@@ -0,0 +1,34 @@
...
+score() {
+  let score = 0;
+  let frameCount = 0;
+  
+  for ( let frame = 0; frame < 10; frame++) {
+    if (this._rolls[frameCount] == 10) {
+      score += 10 + this._rolls[frameCount + 1] + this._rolls[frameCount + 2];
+      frameCount++;
+    }else if (this._rolls[frameCount] + this._rolls [frameCount + 1] == 10) {
+      score += 10 + this._rolls[frameCount + 2];
```
GlykeriaStr, line 24: Why this._rolls[frameCount +2] ? I thought we had to add only the first roll after the frame with the spare.

src/javabowling.js
```diff
@@ -0,0 +1,34 @@
...
+      frameCount += 2;
```
GlykeriaStr, line 25: Why is it +=2?

spec/javabowlingspec.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
```
GlykeriaStr, line 1: Very nice tests! Precise and to the point!

GlykeriaStr (commented):
Very good job with the code! All the tests pass and it works fine!!
The only thing is that it lets you knock down as many pins as you want each round and it didn't seem to stop after the tenth frame. I could keep playing. Other than that, really nice code! :)

PR #1320, opened 2021-03-14
JS bowling challenge

charlierdm:


README.md
```diff
@@ -1,38 +1,46 @@
+# Bowling Challenge
 
-Bowling Challenge
-=================
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday week
+![bowling](https://user-images.githubusercontent.com/75075773/111071204-f1cef880-84cc-11eb-8278-1501835dd605.gif)
```
AJ8GH, line 4: I love the style - very 80s 🙌 ⚡️

README.md
```diff
@@ -1,38 +1,46 @@
...
+To see the interface in the browser:
 
-* Forking this repo
+```
+$ open index.html
+```
+To run the tests:
+```
+$ open SpecRunner.html
+```
```
AJ8GH, line 25: As a user I really appreciated this section. Made it very easy to get going

README.md
```diff
@@ -1,38 +1,46 @@
...
 
-In any order you like:
+Repeat these commands, in the 10th frame, strike and spare bonus functions are disabled so only base points are added(your 10th frame bonus points). 
 
-* Create a nice interactive animated interface with jQuery.
-* Set up [Travis CI](https://travis-ci.org) to run your tests.
-* Add [ESLint](http://eslint.org/) to your codebase and make your code conform.
+In the future, the User Interface could be updated so that each function is interactive and the scorecard could be used solely via it's interface.
+
+<img width="726" alt="Screenshot 2021-03-14 at 14 19 49" src="https://user-images.githubusercontent.com/75075773/111071940-5e97c200-84d0-11eb-9f5d-dc390299fa55.png">
```
AJ8GH, line 43: really like this paragraph, highlighting how the app can be extended

index.html
```diff
@@ -0,0 +1,35 @@
...
+        <button id='begin-frame'>Begin Frame</button>
+        <p>
+        <form id='enter-points' action= 'index.html'>
+          <input type='text' id='throw-input' placeholder="Enter Points"></input>
+          <input type='submit '></input>
+        </form>
+        </p>
+      </p>
+      <img src='https://bloximages.chicago2.vip.townnews.com/cumberlink.com/content/tncms/assets/v3/editorial/3/26/32657b72-9d18-5a01-a161-3fa67be1a4ba/59231a809f884.image.jpg?resize=1200%2C821' width = 800px height = 400px>
+    </section>
```
AJ8GH, line 27: why did you choose to link the image from the web instead of including in your project as static asset?

spec/frameSpec.js
```diff
@@ -0,0 +1,36 @@
...
+  it('can calculate if a strike has been thrown', function() {
+    frame.addRoll(10);
+    expect(frame.isStrike()).toBe(true);
+  })
+
+  it('can calculate if a spare has been thrown', function() {
+    frame.addRoll(8);
+    frame.addRoll(2);
+    expect(frame.isSpare()).toBe(true);
+  })
```
AJ8GH, line 34: These are great tests. easy to read and focusing on behaviour

spec/gameSpec.js
```diff
@@ -0,0 +1,103 @@
...
+  })
+
+  it('begins a new frame',function() {
+    expect(game.turn).toEqual(1)
+  })
+
+  it('has an empty scorecard at first', function() {
+    expect(game.scorecard).toEqual([])
+  })
+
```
AJ8GH, line 18: These tests strike me as testing state / implementation. pun intended. 

I say this because it's the same feedback I got from Alice last weekend.

Is it essential that the scores are stored in array? rather than `expect(game.scorecard).toEqual([])` you could do something like: 
`expect(game.scorecard.length).toBe(0)`

instead of `expect(game.scorecard[0]).toEqual([ 7 ])` you could do:
`expect(game.totalPoints).toBe(7 )`

it's a subtle difference, but I think this is what coaches mean when they tell us to test behaviour over state. I'm still learning this concept myself though really.

AJ8GH (commented):
overall a very strong piece of work
- really enjoyed your css styling with great use of striking imagery
- Seems like you finished all of the key specifications 🙌  you highlighted in your README that there wasn't spare time to hookup the UI to the logic but that was an optional extra anyway
- left a comment about testing behaviour / state - happy to discuss!
- code is clean and readable - very nice use of JS in general.

spec/gameSpec.js
```diff
@@ -0,0 +1,103 @@
...
+  it('adds the users points to each respective frame', function() {
+    game.throwBall(7);
+    game.addToScorecard();
+    expect(game.scorecard[0]).toEqual([ 7 ])
```
siellsiell, line 22: As Adam pointed out in his comment above, some of your tests (like this one) are testing state rather than implementation. It might be useful to revisit [this practical](https://github.com/makersacademy/skills-workshops/blob/master/practicals/testing/behaviour_not_state.md) on the topic. The "How do I test behaviour?" section in particular shows an example of how to think about the difference between testing state and behaviour and derive an appropriate test case. 

In this case you are directly accessing the scorecard array which means that later if you change your mind about how to store your scorecard, you're going to have to update your tests. Perhaps your `Game` and/or `Frame` classes are missing some method(s) to allow a user of the class to easily find out what's on the scorecard for a particular frame?

spec/gameSpec.js
```diff
@@ -0,0 +1,103 @@
...
+      game.addToScorecard();
+      game.beginFrame();
+      game.throwBall(7)
+      game.throwBall(2);
+      game.addToScorecard();
+      game.strikeBonus();
+    })
+
+    it('adds a strike bonus', function() {
+      expect(game.scorecard[0]).toEqual([10, 7, 2])
```
siellsiell, line 49: Similarly, what is the user-observable effect of adding a strike bonus? Can this expectation be expressed in terms of `game.totalPoints()`?

src/game.js
```diff
@@ -0,0 +1,67 @@
...
+  addToScorecard() {
+    this.scorecard.push(this.frame.score);
+    return this.scorecard;
+  }
+
+  strikeBonus() {
+    if(this.isLastFrame()) {
+      return null;
+    }
+    if(this.scorecard[this.scorecard.length -2] === [10] || [0, 10]) {
```
siellsiell, line 33: Are you sure this line is doing what you want it to do? The `===` operator takes precedence over `||`, so what you wrote there is the equivalent to doing:

`(this.scorecard[this.scorecard.length -2] === [10]) || [0, 10]`

What will be the result of this expression if `(this.scorecard[this.scorecard.length -2] === [10])` is false?

src/game.js
```diff
@@ -0,0 +1,67 @@
...
+      let strikeBonus = this.scorecard[this.scorecard.length - 1];
+      this.scorecard[this.scorecard.length - 2] = this.scorecard[this.scorecard.length - 2].concat(strikeBonus);
+    }
+  }
+
+  spareBonus() {
+    if(this.isLastFrame()) {
+      return null;
+    }
+    if(this.scorecard[this.scorecard.length -2] === [10] || [0, 10]) {
```
siellsiell, line 43: same here. It's also the same condition as the one you used in `strikeBonus()`, is that intentional? It doesn't look like this would describe a spare but you've spent more time with the bowling scoring logic than I! Do your tests currently pass?

If the code in this function is indeed meant to be the same, then you could reduce duplication by extracting the overlapping code into a helper function.

src/frame.js
```diff
@@ -0,0 +1,32 @@
...
+    this.round = number;
+    this.score = [];
+  }
+
+  addRoll(pins) {
+    this.score.push(pins);
+    return this.score;
+  } 
+
+  spareOrStrike() {
```
siellsiell, line 14: You're not using this function anywhere. It also doesn't seem very useful given that you already have `isStrike` and `isSpare`.

spec/gameSpec.js
```diff
@@ -0,0 +1,103 @@
...
+  })
+
+  it('calculates the total score', function() {
+    game.throwBall(7);
+    game.throwBall(2);
+    game.addToScorecard();
+    game.throwBall(5);
+    game.throwBall(4);
+    game.addToScorecard();
+    expect(game.totalPoints()).toEqual(18)
```
siellsiell, line 32: Here you've nicely tested the user-visible behaviour :)

src/game.js
```diff
@@ -0,0 +1,67 @@
...
+  throwBall(points) {
+    return this.frame.addRoll(points);
+  }
+
+  addToScorecard() {
+    this.scorecard.push(this.frame.score);
+    return this.scorecard;
+  }
+
+  strikeBonus() {
```
siellsiell, line 29: This function and the one below could do with some comments explaining the logic as it isn't obvious to someone unfamiliar with bowling.

siellsiell (commented):
Most of your code and tests are clear and very readable! Your next focus in terms of testing could be to improve your approach to testing behaviour over state.

I couldn't quite make sense of your logic for spares and strike bonuses (see my comments), so this could be one place to improve how self-documenting your code is.

src/game.js
charlierdm: I didn't know that === takes precedence yet, tests are passing though so it's working as expected.

src/game.js
charlierdm: Ok noted, thank you for the advice.

spec/gameSpec.js
charlierdm: I'll check that practical out. That all makes sense so I will give this some thought and practice, thank you.

PR #1321, opened 2021-03-14
Final solution

a-nawrocka:


src/ScoreCalculator.js
```diff
@@ -0,0 +1,33 @@
...
+        if (nextFrame.length === 1) {
+          bonusScore = nextFrame[0] + thirdFrame[0];
+        } else {
+          bonusScore = nextFrame[0] + nextFrame[1];
+        }
+      } else if (currentFrame.length === 2){
+        bonusScore = nextFrame[0];
+      }
+    }
+    return baseScore + bonusScore;
```
kerisic, line 31: Interesting use of for loops! Can you explain this to me?

PR #1327, opened 2021-03-15
Week 6 weekend challenge 

kerisic:
Translated all logic from ruby into Javascript successfully - unsure how to do dependency injection and spy mocking in jasmine. Added basic table html and css for the scoresheet's interfact, and jQuery for click events to trigger score adding. Score inputs and calculations are synced with the Game model.


a-nawrocka:
Well done with this challenge. Definitely I could use it to learn :)

PR #1336, opened 2021-04-11
Bowling Scorecard with UI

cspoppuppy:
- Use jQuery from CDN
- Use jest for test
- Use ESLint for code checks
- Use Travis for integration
- Functions and unit test are fully implemented, test coverage: 98.41%
- Some style and animation

- Not done feature test yet


.eslintrc.json
```diff
@@ -0,0 +1,23 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "jquery": true,
+        "jest": true
```
fg24davies, line 6: I am interested to know why you switched from Jasmine to Jest

README.md
```diff
@@ -1,75 +1,32 @@
...
-
-## Code Review
-
-In code review we'll be hoping to see:
-
-* All tests passing
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Note that referring to this rubric in advance may make the challenge somewhat easier.  You should be the judge of how much challenge you want.
+# Bowling Scorecard
```
fg24davies, line 1: Really nice clear README with good instructions on how to use the repo

index.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
```
fg24davies, line 1: Such a clean HTML file! It's very easy to read and match up to your webpage

src/bowling.js
```diff
@@ -0,0 +1,89 @@
+import Frame from './frame.js';
```
fg24davies, line 1: Nice conversion of your ruby code to javascript. Really readable and easy to understand whats going on.

src/bowling.js
```diff
@@ -0,0 +1,89 @@
...
+
+export default class Bowling {
```
fg24davies, line 3: Curious to know how this line works

src/bowling.test.js
```diff
@@ -0,0 +1,69 @@
+import Bowling from './bowling.js';
+
+let bowling;
```
fg24davies, line 3: The jest tests are so much clearer than Jasmine! Is that why you chose to change? It looks so much more like RSpec to me.

src/frame.js
```diff
@@ -0,0 +1,31 @@
+export default class Frame {
```
fg24davies, line 1: So impressed with how SRP and DRY your code is Sha, well done :)

src/interface.js
```diff
@@ -0,0 +1,110 @@
...
+      <div class="roll-2" id="roll-${i+1}-2">${frame.rolls[1] === undefined ? "" : frame.rolls[1]}</div>
+      <div class="roll-3" ${i===9 ? "class='disable'" : ""} id="roll-${i+1}-3">${frame.rolls[2] === undefined ? "" : frame.rolls[2]}</div>
+      <div class="frame-score" id="${i+1}-score">${frame.score === 0 ? "" : accumScore}</div>
+      <div class="frame-result" id="${i+1}-result">${frame.result}</div>
+    </div>
+  `);
+  }
+}
+
+function setRollArea(pinsLeft) {
```
fg24davies, line 47: This is really cool! You were adding or subtracting the number of images depending on the previous roll and then setting to 10 after a closed frame in the function nextRoll? Really cool!

style.css
```diff
@@ -0,0 +1,89 @@
...
+.anim {
+  background: linear-gradient(90deg, #f00, #fff, #f00);
+  background-repeat: no-repeat;
+  background-size: 80%;
+  animation: animate 5s linear infinite;
+  -webkit-background-clip: text;
+  -webkit-text-fill-color: rgba(255, 255, 255, 0);
+}
+
+@keyframes animate {
```
fg24davies, line 77: How does this work? I was looking for another keyframes variable but I couldn't find one

src/bowling.js
cspoppuppy: it export the class as a module

.eslintrc.json
cspoppuppy: it is easy to include the coverage, also I had problem of the Jasmine web interface to include code modules

src/bowling.test.js
cspoppuppy: Yes, it looks more like RSpec!

style.css
cspoppuppy: This styling piece is what I found from the internet, haven't dug deep on it, think this sets the start and end positions of the animation

index.html
```diff
@@ -0,0 +1,44 @@
...
+    </div>
+
+    <div class="scorecard-body">
+    </div>
+  </div>
+
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.6.0/jquery.min.js"></script>
+  <!-- <script type="module" src="src/frame.js"></script>
+  <script type="module" src="src/bowling.js"></script> -->
+  <script type="module" src="src/interface.js"></script>
```
alicelieutier, line 42: Using modules - fancy ;)
Make sure to remove commented out code when you clean up your code before a PR.

src/bowling.js
```diff
@@ -0,0 +1,89 @@
...
+  constructor() {
+    this.frames = [];
+    this.gameStatus = true;
```
alicelieutier, line 6: This name "gameStatus", coupled with true is not easy to understand. What is this property indicating? Is it that the game is not finished? If yes, then `this.isGameOver` or `this.hasGameEnded`, and changing your true to false here would be easier to understand.

[Here's an article I wrote about naming](https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098) if you want to go further there.

src/bowling.js
```diff
@@ -0,0 +1,89 @@
...
+    this.totalScore = 0;
+    this._newFrame();
+  }
+
+  roll(pins) {
+    if (!this.gameStatus) return;
+    this.currentFrame.roll(pins, this._isBonus);
```
alicelieutier, line 13: Is the current frame always the last one in `this.frames`? If yes, then storing it separately is duplication of information. You could have a method `this.currentFrame()` that returns the last element of `this.frames`.

src/bowling.test.js
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(() => {
+    bowling = new Bowling();
+  });
+
+  test("New frame after 2 rolls (no strike or spare)", () => {
+    expect(bowling.frames.length).toEqual(1);
+    bowling.roll(1);
+    expect(bowling.frames.length).toEqual(1);
+    bowling.roll(4);
+    expect(bowling.frames.length).toEqual(2);
```
alicelieutier, line 14: This is more of a stateful test than a behaviour test IMO.

src/bowling.test.js
```diff
@@ -0,0 +1,69 @@
...
+    expect(bowling.frames.length).toEqual(1);
+    bowling.roll(5);
+    expect(bowling.frames.length).toEqual(2);
+  });
+
+  test("New frame after Strike", () => {
+    expect(bowling.frames.length).toEqual(1);
+    bowling.roll(10);
+    expect(bowling.frames.length).toEqual(2);
+  });
```
alicelieutier, line 29: Same

src/bowling.test.js
```diff
@@ -0,0 +1,69 @@
...
+    bowling.roll(7);
+    bowling.roll(3);
+    bowling.roll(6);
+    bowling.roll(4);
+    bowling.roll(10);
+    bowling.roll(2);
+    bowling.roll(8);
+    bowling.roll(6);
+    expect(bowling.totalScore).toEqual(133);
+  });
```
alicelieutier, line 68: Great behaviour test.

src/bowling.js
```diff
@@ -0,0 +1,89 @@
...
+    this._checkGameStatus();
+    this._prepareNextRoll();
+  }
+
+  _prepareNextRoll() {
+    if (this._isNextFrame) {
+      this._adjustScoresForPastFrames();
+      this._newFrame();
+    }
+    if (this._isBonus) {
```
alicelieutier, line 24: I think the bonus functionality could be moved to the Frame class. The Frame can keep track of whether it still needs to be bonus rolls or not, and decide to count them in or not.
This way this class here can just distribute new rolls to all frames, or just to the last three frames for example, and the frame can do their own calculation.

src/bowling.js
```diff
@@ -0,0 +1,89 @@
...
+    if (this.frames.length < 2) return;
+    let prevFrame = this.frames[this.frames.length - 2];
+    let extraScore = 0;
+    if (prevFrame.result === "Strike") {
+      extraScore = (this.currentFrame.rolls.length < 2 ? this.currentFrame.score : (this.currentFrame.rolls[0] + this.currentFrame.rolls[1]));
+    } else if (prevFrame.result === "Spare") {
+      extraScore = this.currentFrame.rolls[0];
+    }
+    prevFrame.addScore(extraScore);
+    this.totalScore += extraScore;
```
alicelieutier, line 44: With the design explained above, a lot of this would be simplified.

src/bowling.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  _newFrame() {
+    this.currentFrame = new Frame();
+    this.frames.push(this.currentFrame);
+  }
+
+  get _isNextFrame() {
+    if (this.frames.length === 0 || ((this.frames[this.frames.length-1].result !== "" || 
+      this.frames[this.frames.length-1].rolls.length === 2) && this.frames.length <= 9)) {
```
alicelieutier, line 67: This condition is too long to process for the reader. you could extract part of it in method that have meaningful names.

src/frame.js
```diff
@@ -0,0 +1,31 @@
...
+    this.rolls.push(pins);
+    this.score += pins;
+    if (bonus===false) {
+      this.pinsLeft -= pins;
+    } 
+    if (this.pinsLeft===0 && this.rolls.length===1) {
+      this.result = "Strike";
+    } else if (this.pinsLeft===0 && this.rolls.length===2) {
+      this.result = "Spare";
+    }
```
alicelieutier, line 20: Instead of calculating the result and setting it on the class, a better approach would be to make `result()` and `score()`  (and maybe `pinsleft`) methods that calculates on the fly based on the rolls. This avoids duplication of information (here basically the info is stored 4 times, in different guises), and means this roll method is only responsible for updating the rolls.

eg:
```
score(){
  this.rolls.reduce(function(a, b){ return a + b })
}

result() {
  if (this.rolls.length > 0 && this.rolls[0] === 10) {
    return "Strike"
  } 
  // .... etc
}
```

alicelieutier: You can also probably get rid of pinsLeft that way. Less duplication of information is better, because it avoids mistakes and inconsistency between sources of truth.

src/frame.test.js
```diff
@@ -0,0 +1,45 @@
...
+    frame = new Frame();
+  });
+
+  test ("Strike if roll 10 in first go", () => {
+    frame.roll(10);
+    expect(frame.pinsLeft).toEqual(0);
+    expect(frame.score).toEqual(10);
+    expect(frame.result).toEqual("Strike");
+    expect(frame.rolls[0]).toEqual(10);
+  });
```
alicelieutier, line 16: In order to know if a test is testing state/implementation or behaviour, think of what should be accessible/visible outside the class. It's likely users of the Frame class only need to know the result and the score, and only need to input the rolls. That defines the interface of the class with the outside world, so that is hat your tests should focus on:
```
INPUT | OUTPUT
some rolls | expect score and result to be what I want them to be.
```

src/frame.test.js
```diff
@@ -0,0 +1,45 @@
...
+    expect(frame.pinsLeft).toEqual(5);
+  });
+
+  test ("can add score (for adjust scores for past strike/spare frames)", () => {
+    expect(frame.score).toEqual(0);
+    frame.addScore(6);
+    expect(frame.score).toEqual(6);
+  })
+
+});
```
alicelieutier, line 45: Detail, but having a newline at the end of your file is typical. I'm surprised your linter or IDE does not enforce it.

src/interface.js
```diff
@@ -0,0 +1,110 @@
...
+function initiateGame() {
+  bowling = new Bowling();
+  refresh();
+}
+
+function refresh() {
+  setRollArea(bowling.currentFrame.pinsLeft);
+  $(".scorecard-body").empty();
+  let accumScore = 0;
+  for (let i = 0; i < bowling.frames.length; i++) {
```
alicelieutier, line 31: Maybe iterate from 1 to length + 1, and get rid of all the `i+1` below?

alicelieutier: or have a variable like `const frameNumber = i+1`

alicelieutier (commented):
I've added a number of comments, mostly about code design and testing, but overall this is a really good challenge.

PR #1337, opened 2021-04-11
Not great, but something

Emmapr123:


spec/framesSpec.js
```diff
@@ -0,0 +1,65 @@
+describe('roll', function() {
+  let frame = new Frame;
```
StuBehan, line 2: could maybe have used a before each for your Frame objects, but the semantics of the object names is useful to the readability of the tests so I dunno which I would prefer. 🤷

spec/scoreCardSpec.js
```diff
@@ -0,0 +1,152 @@
...
+  let scoreCard = new ScoreCard;
+  it('should return 300 on a prefect game', function() {
+    for(let i = 0; i < 12; i++) {
+      scoreCard.play(10)
+    }
+
+    expect(scoreCard.play(10)).toEqual('Game ended! Your score was 300')
+  })
+})
+
```
StuBehan, line 152: Your tests are nice, the only comment would be the same as the previous spec. It's good that you've got complete game tests here and non just testing the tenth frame in isolation.

README.md
```diff
@@ -1,75 +1,56 @@
...
+    bonus = number of pins in the next roll
+
+- 10th Frame: Player gets spare or strike in the last frame
+    bonus = the last roll
+    never more than 3 rolls
+
+- Gutter game: No pins hit at all
+
+- Perfect game: 12 strikes
+    300 points
```
StuBehan, line 56: Would be nice to have a quick list of console commands to run the logic to see it in action, but you have covering tests so it's not too big of a deal!

src/frames.js
```diff
@@ -0,0 +1,18 @@
...
+    this.rolls.push(pins);
+  }
+  isStrike() {
+    return (this.rolls[0] === 10 ? true : false)
+  }
+  isSpare() {
+    return (this.rolls.length == 2 && this.rolls[0] + this.rolls[1] === 10 ? true : false)
+  }
+  count() {
+    return (this.rolls.length > 1 ? this.rolls[0] + this.rolls[1] : this.rolls[0])
```
StuBehan, line 15: The ternary if is well applied here and makes this class look super sleek 👍

src/scoreCard.js
```diff
@@ -0,0 +1,83 @@
+class ScoreCard {
+  constructor(frames, frame, score) {
+    this.frames = []
+    this.frame = null
+    this.lastFrame = null
```
StuBehan, line 5: Assigning null instead of leaving a var undefined is good, it means you can test for it later too with ===. 

worth thinking about because var == null and var == undefined are the same whereas var === null and var === undefined aren't!

src/scoreCard.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+  previousFrameIsSpare(index) {
+    return (this.frames[index - 1] && this.frames[index - 1].isSpare())
+  }
+  previousTwoFramesAreStrikes(index) {
+    return (this.frames[index - 2] && this.frames[index - 2].isStrike() && this.previousFrameIsStrike(index))
+  }
+  finalFrame() {
+    return this.frames.length === 9
+  }
```
StuBehan, line 50: nice use of helper methods for if logic, they seem to get unwieldy really fast with js

StuBehan (commented):
Good execution, was really interesting to see an alternative take on the bowling logic!

PR #1338, opened 2021-04-11
Bowling Challenge - Javascript

louballantyne:


README.md
```diff
@@ -1,75 +1,11 @@
...
-
-## Code Review
-
-In code review we'll be hoping to see:
-
-* All tests passing
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Note that referring to this rubric in advance may make the challenge somewhat easier.  You should be the judge of how much challenge you want.
+If the user inputs incorrect rolls, they will have to reset the app. Ideally, I'd rectify this.
```
jackmcc08, line 11: Simple set up instructions, which work! Great. 

The UI worked and I could enter scores and get a final score which was great, it also seems to handle strikes, spares, gutter balls as well. 

I noticed an error where if you enter a second roll before a first roll it returns NaN and does not allow you to rectify.

An idea for the README could be to have a bit of explanation about how the logic of the code works.

SpecRunner.html
```diff
@@ -0,0 +1,30 @@
...
+  <script src="spec/gameSpec.js"></script>
+  <script src="spec/frameSpec.js"></script>
+  <script src="spec/scoreSpec.js"></script>
+
+
+</head>
+
+<body>
+</body>
+</html>
```
jackmcc08, line 30: Well set up spec runner, worked on first try and all tests passed!

index.html
```diff
@@ -0,0 +1,385 @@
...
+          </select>
+          <div id="f10score"> </div>
+          <button type="button" name="button" id="go">Go!</button>
+        </div>
+      </div>
+    </div>
+    <button type="button" name="button" id="reset">Reset</button>
+
+  </body>
+</html>
```
jackmcc08, line 385: HTML interface works and looks neat and tidy. 

Only idea I had is, there must be a way to remove duplication of the code...not sure how this would be done in HTML. With html.erb you can use partials - not sure if there is a similar way in Javascript.

spec/frameSpec.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+    it('returns a bonus of 1, given a spare', function(){
+      expect(frame.calcBonus(3,7)).toEqual(1);
+    });
+    it('returns a bonus of 0, if there is no strike or spare', function(){
+      expect(frame.calcBonus(2,3)).toEqual(0);
+    });
+  });
+
+});
```
jackmcc08, line 36: I like the tests here, and they seem to test all the functions of the Frame class.

spec/gameSpec.js
```diff
@@ -0,0 +1,121 @@
...
+      game.roll(10);
+      game.roll(10);
+      game.roll(10);
+      game.roll(10);
+      game.roll(10);
+      game.roll(10);
+      game.roll(10);
+      game.roll(10);
+      game.roll(10,10,10);
+      var score = game.gameScore.reduce((a,b) => a + b, 0);
```
jackmcc08, line 73: Idea: could use a for loop to avoid duplication.

src/frame.js
```diff
@@ -0,0 +1,27 @@
...
+  calcBonus (roll1, roll2) {
+    if (this.strike(roll1) === true) {
+      return 2;
+    } else if (this.spare(roll1, roll2)) {
+      return 1;
+    } else {
+      return 0;
+    }
+  }
+}
```
jackmcc08, line 27: Frame class is very clean, with no duplication.

One thing you could look at is reducing the strike and spare classes to one line. 

I think I understand how your bonus mechanism works, and I like how you are assigning a number so it knows how much bonus needs to be applied'

src/game.js
```diff
@@ -0,0 +1,38 @@
...
+      return console.log('Rolls must not exceed a total of ten');
+    }
+    let bonus = this.frameClass.calcBonus(roll1, roll2);
+    if (this.frame === 9) {
+      var score = this.scoreClass.calculateFrame10(roll1, roll2, roll3);
+    } else {
+      var score = this.scoreClass.calculate(roll1, roll2, bonus);
+    }
+    this.storeScore(score[0], score[1], score[2]);
+    this.frame += 1;
```
jackmcc08, line 23: JS shorthand you can use ++ instead of += 1

I like the control flow here. 

The only thing that confused me was the use of 9 as the 10th frame, but guess this revolves around indexing.

src/game.js
```diff
@@ -0,0 +1,38 @@
...
+      this.gameScore[this.frame - 1] = this.gameScore[this.frame - 1] + bonusScore;
+    } else if (this.frame > 1) {
+      this.gameScore[this.frame - 1] = this.gameScore[this.frame - 1] + bonusScore;
+      this.gameScore[this.frame - 2] = this.gameScore[this.frame - 2] + previousBonusScore;
+    }
+    if (this.frame === 9) {
+      return this.gameScore.reduce((a, b) => a + b, 0);
+    }
+  }
+}
```
jackmcc08, line 38: How does this work for frame 0?

src/game.js
```diff
@@ -0,0 +1,38 @@
+class Game {
```
jackmcc08, line 1: Game class has simple conditionals which are nice to follow. 

I like the use of guard clauses. 

There is the potential to pull out some functionality into private methods

src/interface.js
```diff
@@ -0,0 +1,394 @@
...
+    $('#f3score').text(game.gameScore[2]);
+    $('#f4score').text(game.gameScore[3]);
+    $('#f5score').text(game.gameScore[4]);
+    $('#f6score').text(game.gameScore[5]);
+    $('#f7score').text(game.gameScore[6]);
+    $('#f8score').text(game.gameScore[7]);
+    $('#f9score').text(game.gameScore[8]);
+    $('#f10score').text(game.gameScore[9]);
+  }
+});
```
jackmcc08, line 394: I like the use of JQuery and the hideing and text addition, it makes the interface look good. 

A couple of thoughts: 
- there is a lot of duplication here, looks like there is the potential to DRY out. 
- why did you store some of the game logic here? Like the strike and total functions?

src/score.js
```diff
@@ -0,0 +1,65 @@
...
+    return [score, this.bonusScore, this.bonusScorePrevious];
+  }
+
+  calculateFrame10 (roll1, roll2, roll3) {
+    this.calcPreviousBonus(roll1, roll2);
+    let score;
+    roll3 === null ? score = (roll1 + roll2) : score = (roll1 + roll2 + roll3);
+    return [score, this.bonusScore, this.bonusScorePrevious];
+  }
+}
```
jackmcc08, line 65: I like what you did here, your logic works and it is a good use of having lots of functions and creating good readability. 

Only thought is you can use the javascript private method convention with an underscore, e.g. _calculate to indicate which methods are used in this class only.

jackmcc08 (commented):
Overall you model logic is simple, and works well. The tests passed and I could enter scores and get an overall game score! 

There are a few ideas where I found some of the code confusing, or some ideas for adjustment. 

But overall I like how simple the model is, much better than my mess :D

PR #1339, opened 2021-04-11
Bowling Challenge - JS - Jack M.

jackmcc08:
My answer to the javascript bowling challenge. 

See README and development notes for details on solution.


src/BowlingScorecard.js
```diff
@@ -0,0 +1,143 @@
...
+    return results
+  }
+
+  _isInvalidScore(score) {
+    if (score > this._maxRoll() || score < 0) return true
+    if (`${score}`.match(/\D/) !== null) return true
+
+    return false
+  }
+
```
louballantyne, line 52: Nice use of private functions

src/BowlingScorecard.js
```diff
@@ -0,0 +1,143 @@
...
+    return 10
+  }
+
+  _checkFrameOver() {
+    this._updateTotal()
+
+    if (this.frame === 10 && this._currentFrame.firstRoll !== null && this._currentFrame.secondRoll !== null && this._currentFrame.thirdRoll !== null ) return "EndGame";
+    if (this.frame === 10 && this._currentFrame.firstRoll !== null && this._currentFrame.secondRoll !== null && this._currentFrame.strike === false && this._currentFrame.spare === false) return "EndGame";
+
```
louballantyne, line 77: to DRY your code, perhaps you could extract 'this.frame === 10 && this._currentFrame.firstRoll !== null && this._currentFrame.secondRoll !== null' into a method

src/BowlingScorecard.js
```diff
@@ -0,0 +1,143 @@
+// const Frame = require('./Frame.js');
+
```
louballantyne, line 2: Your BowlingScorecard class has a few responsibilities: -it increments the frame, it stores the frames, it checks rolls are valid, it applies bonuses etc. I think you could extract another class from this class.

src/BowlingScorecard.js
```diff
@@ -0,0 +1,143 @@
...
+    if (this._currentFrame().isComplete()) {
+      this.frame++;
+      this.frames.push(new Frame(this.frame, this.currentScore()));
+    }
+  }
+
+  _currentFrame() {
+    return this.frames[this.frames.length - 1]
+  }
+
```
louballantyne, line 87: Love the notation here for strikes and spares

src/BowlingScorecard.js
```diff
@@ -0,0 +1,143 @@
...
+    if (this.frame === 10 && this._currentFrame.firstRoll !== null && this._currentFrame.secondRoll !== null && this._currentFrame.thirdRoll !== null ) return "EndGame";
+    if (this.frame === 10 && this._currentFrame.firstRoll !== null && this._currentFrame.secondRoll !== null && this._currentFrame.strike === false && this._currentFrame.spare === false) return "EndGame";
+
+    if (this._currentFrame().isComplete()) {
+      this.frame++;
+      this.frames.push(new Frame(this.frame, this.currentScore()));
+    }
+  }
+
```
louballantyne, line 83: Nice method to make your code readable

src/BowlingScorecard.js
```diff
@@ -0,0 +1,143 @@
...
+  _updateTotal() {
+    this._currentFrame().updateCurrentTotal();
+  }
+
+  _applyBonus() {
+    if (this.frame === 1) return;
+    if (this.frame === 10 && this._currentFrame().thirdRoll !== null) return;
+
+    let previousFrame = this.frames[this.frames.length - 2];
+
```
louballantyne, line 111: To make your code more readable here, and to reduce the length of your 'if' statements, I think you could extract this into further methods. However, the well-named methods do make it as easier to understand :)

spec/BowlingScorecardSpec.js
```diff
@@ -0,0 +1,210 @@
...
+        i++
+      }
+      testScorecard.enterRoll(5)
+      testScorecard.enterRoll(5)
+      testScorecard.enterRoll(5)
+
+      expect(testScorecard.generateScorecardInfo()).toEqual(SOLUTION_ALT_THREE_THROW_SCORECARD)
+    })
+
+
```
louballantyne, line 146: This is a really nice way of managing your tests

src/index-jQuery.js
```diff
@@ -0,0 +1,48 @@
...
+  // alert( "Welcome to the scorecard!");
+  // console.log(scorecard._tidyDisplay(5))
+  displayScorecard()
+
+  $( '#enterRoll-form' ).submit(function( event ) {
+    let score = $("#rollScore").val()
+    console.log(scorecard.enterRoll(score))
+    displayScorecard()
+    event.preventDefault();
+  });
```
louballantyne, line 23: What does preventDefault do here?

louballantyne (commented):
I like that each frame is an instantiation of a class with its own attributes. This is a really nice way of managing the game and therefore enabling the extension of editing frames, if necessary. I think some of your code could be extracted into a further class and into further methods, in order to reduce the responsibility of each section of your code and to improve the readability. However, your methods were aptly named, which meant that it was trivial to determine what each part was doing. I would be interested to know how you might plan to correct the logic of the tenth frame, such that it doesn't allow the user to roll three balls if they don't get a strike or a spare.

src/index-jQuery.js
jackmcc08: Prevent Default stops the normal function of clicking submit on the form (which seems to refresh the page if not given a direction) so you need preventDefault to stop the standard function of the submit button and allow the javascript methods to resolve and stay on the page.

louballantyne: I would have thought you didn't need it as your form doesn't have an action or a method. It would be worth trying without the prevent default in, to see if it still works as intended.

PR #1340, opened 2021-04-12
Bowling scorecard js

taran314:
Most of the logic was completed, a few things remain such as a reset game function and a game over message. Wanted to hide the buttons in the interface so user could not score invalid combinations (9, 9) but did not have time to implement this. Further notes, a TODO list and instructions on how to run the app are in the README.


index.html
```diff
@@ -0,0 +1,80 @@
...
+<div>
+  <span id="8roll1">Roll 1:</span> <span id="8roll2">Roll 2:</span>
+</div>
+
+<h2 id="frame9">Frame 10:</h2>
+<div>
+  <span id="9roll1">Roll 1:</span> <span id="9roll2">Roll 2:</span> <span id="9roll3">Roll 3:</span>
+</div>
+  
+</html>
```
mstrivens, line 80: I like the web interface layout. Having buttons instead of user input is a nice touch as well and i like the idea of graying out buttons to limit user input and remove edge cases.

spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
+  jasmine.addMatchers({
+    toBePlaying: function () {
+      return {
+        compare: function (actual, expected) {
+          var player = actual;
+
+          return {
+            pass: player.currentlyPlayingSong === expected && player.isPlaying
+          };
```
mstrivens, line 10: I think this might be from the original download?

src/frame.js
```diff
@@ -0,0 +1,54 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  isSumTen() {
+    var sum = this.scores.reduce((a, b) => a + b, 0);
+    return sum === 10;
+  }
+}
```
mstrivens, line 54: I like how you've managed dot extract this into a different class.I don't completely understand how the logic works here but that could be just as much to do with my unfamiliarity with javascript as the way it's written.

mstrivens: I've commented a few places above where it's a bit confusing

src/frame.js
```diff
@@ -0,0 +1,54 @@
+class Frame {
+  constructor() {
+    this.scores = new Array;
+    this.updated = false;
+    this.final = 0;
+    this.added = false;
+  }
+
+  isSpare() {
+    if (this.scores.length == 2) {
```
mstrivens, line 10: how does this work?

src/frame.js
```diff
@@ -0,0 +1,54 @@
...
+      return this.isSumTen();
+    } else {
+      return false;
+    }
+  }
+
+  isStrike() {
+    if (this.scores.length < 3) {
```
mstrivens, line 18: and this? I don't fully understand how the length of scores array ties into a spare or strike being bowled.

src/frame.js
mstrivens: I would be good to talk through this together at some point

src/bowling.js
```diff
@@ -0,0 +1,58 @@
...
+    // consecutive strikes
+    if (this.currentFrame > 1) {
+      var twoFramesAgo = this.frames[this.currentFrame - 2];
+      if (twoFramesAgo.isStrike()) {
+        twoFramesAgo.scores.push(pins);
+      }
+    }
+  }
+}
+
```
mstrivens, line 57: The class looks goo and seems like methods have good SRP. I'm really interested to find out more about how this works as I don't immediately understand it

mstrivens (commented):
I liked the visual interface and the code seemed very well written and concise. It would be good to chat a bit further about it as my code seems to have been overcomplicated. Well done

PR #1341, opened 2021-04-12
JS Bowling game logic

dmgreaves:


src/bowling.js
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor() {
+    this.gameScore = 0;
+    this.frame = 1;
+    this.currentRoll = 1;
+    this.currentFrameScore = 0;
+    this.frameScores = [];
+  };
+
+roll(pins) {
```
TataSher, line 11: This function seems to be very long. Could be divided into several functions: function that throws an error, function that adds to this.frame, function that adds to this.currentFrameScore.

TataSher (commented):
I think I understand the logic of the code. But because the function is quite long and with a lot of conditionals it is a bit difficult to read. It seems that there is a lot of logic and features that are not covered by tests, so probably would be a good idea to cover these features to make sure they work in an intended way.

PR #1343, opened 2021-04-12
bowling-js

crotchetycrow:


README.md
```diff
@@ -1,3 +1,11 @@
+Found this neat little node called seqNum which basically acts as a counter - I haven't used it yet. Why haven't I used it yet...
+
+        https://github.com/srijs/node-seqnum
```
loushark, line 3: ooo interesting! 🤔

spec/frameSpec.js
```diff
@@ -0,0 +1,48 @@
+describe('Frame', function() {
+    beforeEach(function() {
+        frame = new Frame;
+    });
+
+    var completedFrame = function() {
```
loushark, line 6: Love the addition of this function, makes the tests really nice and readable.
Maybe it could be separated into a webHelper file? (i've not tried that myself, i wonder if you would just require the webhelper through the spec runner)

spec/frameSpec.js
```diff
@@ -0,0 +1,48 @@
...
+        for(let i = 1; i <=2; i ++) {
+            frame.storeRoll();
+        }
+        frame.totalRolls(2)
```
loushark, line 10: some of your lines have semi colons and some don't - i'm still unsure of the best practice for this, as i think they don't necessarily need them in your use case, but wonder if it's worth doing one or the other for readability/consistency of code? i'd be intrigued by this best practice for my own code

src/frame.js
```diff
@@ -0,0 +1,29 @@
+function Frame (seqNum) {
```
loushark, line 1: I've yet to use function and prototype in this way (i used class, which i understand to be a newer addition to JS), so good job on understanding the protoype stuff!

src/frame.js
```diff
@@ -0,0 +1,29 @@
...
+    Frame.prototype.storeRoll = function(roll) {
+        return (this.rolls.push(roll));
+    };
+
+    Frame.prototype.newFrame = function(updateFrame) {
+        if (this.rolls.length === 2) {
+            this.rolls.splice(0, this.rolls.length)
+            this.frames.push(updateFrame)
+        }
+        return (this.rolls.length);
```
loushark, line 21: This function is returning the length of your rolls array, do you mean the length of your frames here?

src/frame.js
```diff
@@ -0,0 +1,29 @@
...
+        // }
+        return (this.rolls.length);
+    };
+    Frame.prototype.storeRoll = function(roll) {
+        return (this.rolls.push(roll));
+    };
+
+    Frame.prototype.newFrame = function(updateFrame) {
+        if (this.rolls.length === 2) {
+            this.rolls.splice(0, this.rolls.length)
```
loushark, line 18: as you have your totalRolls function currently returning the length of your rolls array, you could call this here to keep your code Dry.

loushark (commented):
I think you could add some context notes to your readme to explain at what state your code is currently in, such as the current working logic and what's left to do/your intentions of your project. This would be not only for someone reviewing but also for your own benefit, so you can keep track of your project as you come back to work on it and you won't need to figure out where you left off. 

Despite you saying you've not worked a great deal on this code, i can already see your approach to the logic really clearly in just a few lines. It's simple and clean and i think you'd have no trouble expanding on what you have to make a really nice interactive app!! Good job Jack!

src/frame.js
crotchetycrow: My plan was to set this up as a function and then add perhaps incorporate it into a separate bowlingGame class or something like that - hopefully it would of worked that way.

src/frame.js
crotchetycrow: Ahhh, yes - basically I wanted a newFrame to reset the rolls to 0 every time we jumped into a new frame. Thanks for pointing that out, it would make more sense for it to return the frame rather than the rolls array...

spec/frameSpec.js
crotchetycrow: I did look into that, but from what I could see it isn't super easy to implement a helper in jasmine. You need to add a specific node but I read that it isn't as useful as rspec helpers. Or so the 10 mins of research told me haha

README.md
crotchetycrow: Shame I forgot about it and didn't use it!

PR #1352, opened 2021-05-10
done logic

TiagoManuelPC:


spec/GameSpec.js
```diff
@@ -0,0 +1,41 @@
...
+    expect(game.score()).toEqual(18);
+  })
+
+  it('can roll a strike', function() {
+    game.roll(10)
+    game.roll(2)
+    game.roll(4)
+    loopFunction(0, 17)
+    expect(game.score()).toEqual(22);
+
```
tremella, line 32: you've tested whether a strike works, and it does - but you haven't tested for consecutive strikes/ spares. This is important because they have a cascading effect which can be tricky to handle. I'd recommend making tests for these.

spec/GameSpec.js
```diff
@@ -0,0 +1,41 @@
...
+  })
+ 
+  function loopFunction(pins, rolls) {
```
tremella, line 35: this is a tidy way of handling the rest of the game.

src/Game.js
```diff
@@ -0,0 +1,53 @@
...
+
+  for(let frameIndex = 0; frameIndex < 10; frameIndex++) {
+    if (this.isStrike(rollIndex)) {
+      score += this.strikeBonus(rollIndex);
+      rollIndex++;
+      continue;
+    }
+
+    const frameScore = this.rolls[rollIndex] + this.rolls[rollIndex + 1];
+
```
tremella, line 25: it appears that you've written the logic for strikes and spares, but you haven't really tested all these functions below that you've written (yet). They're tidy and readable though.

tremella (commented):
Tidy and readable code, divided well into classes, clever use of a loop to handle the rest of the game so you can test the bare minimum. Very readable overall, great job. This would be worth finishing.

PR #1355, opened 2021-05-10
Jess Brand bowling challenge

tremella:


src/Scorecard.js
```diff
@@ -0,0 +1,43 @@
+function strikeOrSpare(frame){
+  return (frame[0] + frame[1] == 10 ? true : false);
+}
+function strike(frame){
+  return (frame[0] == 10 ? true : false );
+}
+// takes in a nested arr.
+function Scorecard(bowls){
```
TiagoManuelPC, line 8: thats a very big if/else statment. do you think you could refactor this in smaller funcions?

src/Scorecard.js
```diff
@@ -0,0 +1,43 @@
+function strikeOrSpare(frame){
```
TiagoManuelPC, line 1: hum.. why didnt you put this two first funcions inside the class?

spec/ScorecardSpec.js
```diff
@@ -0,0 +1,53 @@
+describe("Scorecard", ()=> {
```
TiagoManuelPC, line 1: tests looked great. fells like you follow a tdd process

PR #1357, opened 2021-05-10
wip

YohannTisserand:


src/frame.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
+const TOTAL_PINS = 10
+class Frame {
+  constructor() {
+    this.rolls = [];
+    this.score = 0;
```
KaneG9, line 6: Probably better to put score into game as you want to know the score for the whole game rather than just the individual frame

src/game.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
```
KaneG9, line 2: Put bonus calculations in here as part of the a calculate score method

spec/gameSpec.js
```diff
@@ -0,0 +1,17 @@
...
+  let frame;
+  let game;
+
+  beforeEach(() => {
+    game = new Game;
+    frame = jasmine.createSpyObj('frame', ['addRoll', 'getScore', 'isStrike', 'isSpare']);
+  });
+  
+  it("allow to add rolls to the frame", () => {
+    game.addRoll(5);
```
KaneG9, line 13: could put your frame spy in here for the test

PR #1367, opened 2021-06-08
Method

cspoppuppy:
Amend some methods


cspoppuppy:
amend result/score method to attribute

PR #1370, opened 2021-07-12
Bowling Challenge

frank-mck:


public/src/bowling.js
```diff
@@ -0,0 +1,97 @@
...
+    return this._lastFrame()[0] === ALL_PINS && this._lastFrame().length < 3
+  }
+
+  _finalFrame(frame) {
+    if (this._lastFrame().length < 2 || this._calcFrame(this._lastFrame()) === ALL_PINS) {
+      this._lastFrame().push(frame[0]);
+    } else if (this._lastFrame().length === 3 || this._finalFrameStrike()) {
+      this._lastFrame().push(frame[0]);
+    } 
+  }
```
eoinmakers, line 87: Your functions are concise in their responsibilities, which is great! Specifically because this might allow a slightly improved separation of concerns, if you were to draw out the idea of a Frame into its own class. One class could be responsible for the game logic, the other in tracking the idea of a Frame.

public/src/interface.js
```diff
@@ -0,0 +1,108 @@
...
+    index += 1
+  });
+
+  document.querySelector('#ninth').addEventListener('click', () => {
+    frame.addRoll(9);
+    updateFrameAndAddScore();
+    index += 1
+  });
+
+  document.querySelector('#tenth').addEventListener('click', () => {
```
eoinmakers, line 98: You could probably declare most of these event listeners with a loop, as the only thing that changes is two arguments

public/src/bowling.js
```diff
@@ -0,0 +1,97 @@
...
+      this._secondLastFrame().push(frame)
+      this._thirdLastFrame().push(this._lastFrame()[0])
+    }
+  }
+
+  _theLastTwoFramesAreAPerfectScore() {
+    return this._calcFrame(this._thirdLastFrame()) === 20 && this._lastFrame()[0] == ALL_PINS
+  }
+
+  _lastFrame() {
```
eoinmakers, line 49: I think the method names can be a tad more descriptive - to describe something it's 'doing', eg. _getLastFrame()

eoinmakers (commented):
Great work, I thought your unit tests were were written and covered each type of game, including edge cases, successfully. Separation of concerns to split apart some of the functionality around knowing if a frame is a strike/spare vs. calculating the bonus scores would improve the code quality a bit more, but it's a good submission.

PR #1385, opened 2021-09-05
JS Bowling challenge

ashtroboy79:
So Im able to score a complete game as well as a perfect game, I totally forgot about putting in some form of "Gutter Game" response. I also realised that I have not put in a check to ensure that the player doesn't put in more than 10 pins in either a single roll or across a standard frame. This is sort of present in the visual interface. I didn't look at JQuery due to time constraints and I have not been able to get Travis CI to run my tests.


Index.html
```diff
@@ -0,0 +1,25 @@
...
+				<button id="score">Scoring</button>
+				Score is <span id="current-score"></span>
+			</p>
+		</section>
+		<script src="src/Game.js"></script>
+		<script src="src/Frame.js"></script>
+		<script src="src/FinalFrame.js"></script>
+		<script src="src/Interface.js"></script>
+	</body>
+</html>
```
Battery0, line 25: This is great that you managed to get a working interface for the game! Maybe want to amend it so that you can differentiate between a first and second roll for a single frame?

src/FinalFrame.js
```diff
@@ -0,0 +1,22 @@
+class FinalFrame extends Frame {
```
Battery0, line 1: nice, is this inheritance in JS?

src/FinalFrame.js
```diff
@@ -0,0 +1,22 @@
...
+  constructor() {
+    super();
```
Battery0, line 3: Not sure what this is for?

src/FinalFrame.js
```diff
@@ -0,0 +1,22 @@
...
+      this.currentroll++;
+    } else if (this.currentroll === 2) {
+      this.rollTwo = value;
+      this.currentroll++;
+      if (!this.isStrike() || !this.isSpare()) {
+        this.endFrame = true;
+      }
+    } else if (this.isStrike() || this.isSpare()) {
+      this.rollThree = value;
+      this.endFrame = true;
```
Battery0, line 19: May want to look at refactoring large else if statements to a switch statement OR extract some of the code into a separate method?

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  score() {
+    this.currentscore = 0;
+    for (let i = 0; i < this.frameNo - 1; i++) {
+      if (i === 9) {
+        this.currentscore +=
+					this.frames[i]['rollOne'] +
+					this.frames[i]['rollTwo'] +
+					this.frames[i]['rollThree'];
```
Battery0, line 42: Think something went haywire with the indentation here

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+          if (this.isSpare(i)) {
+            this.currentscore += this.frames[i + 1]['rollOne'];
+          }
+        }
+        if (i < this.frameNo - 2) {
+          if (this.isStrike(i)) {
+            this.scoreStrike(i);
+          }
+        }
+        this.currentscore +=
```
Battery0, line 54: May want to think about refactoring large if statements as mentioned in earlier comment

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  rollBall(value) {
+    this.currentFrame.roll(value);
+    if (this.currentFrame.endFrame) {
+      this.nextFrame();
+    }
+  }
+
+  score() {
```
Battery0, line 35: This score method is quite large - may want to extract some code out to separate methods and just call them in this method instead?

src/Interface.js
```diff
@@ -0,0 +1,25 @@
...
+      document.querySelector('#game').innerText = game.frameNo;
+    }
+    document.querySelector('#ballno').innerText = game.currentFrame.currentroll;
+  });
+
+  document.querySelector('#score').addEventListener('click', () => {
+    game.score();
+    document.querySelector('#current-score').innerText = game.currentscore;
+  });
+});
```
Battery0, line 25: This is great!

src/FinalFrame.js
alicelieutier: @Battery0
 Goes with the above ^ When you inherit from a class, you can call super in the constructor to first execute the constructor of the parent. Most languages have a way of doing this - take a look at it!

spec/GameSpec.js
```diff
@@ -0,0 +1,119 @@
...
+    expect(game.nextFrame).toHaveBeenCalledTimes(3);
+  });
+
+  it('can score a partial non-strike/non-spare game', () => {
+    game.rollBall(5);
+    game.rollBall(4);
+    game.rollBall(3);
+    game.rollBall(2);
+    game.score();
+    expect(game.currentscore).toEqual(14);
```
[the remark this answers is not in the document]
alicelieutier: This is also a question of not duplicating information. If you already store the rolls, then the score is just a calculation based on the rolls. Storing both the rolls and the result of the calculation duplicates information in your state and is more prone to bugs if the two get out of sync for any reason.

src/FinalFrame.js
```diff
@@ -0,0 +1,22 @@
...
+
+  roll(value) {
+    if (this.currentroll === 1) {
+      this.rollOne = value;
+      this.currentroll++;
+    } else if (this.currentroll === 2) {
+      this.rollTwo = value;
+      this.currentroll++;
+      if (!this.isStrike() || !this.isSpare()) {
+        this.endFrame = true;
```
alicelieutier, line 15: Why do you need this `endFrame` variable? Surely that can be a method that calculates based on the info?
Then you could have:
```js
roll(value) {
  switch(this.currentroll) {
    case 1:
      this.rollOne = value;
      break;
    case 2:
      this.rollTwo = value;
      break;
    case 3:
      this.rollThree = value;
  }
  this.currentroll++;
}

endFrame() {
  return (
      (this.currentroll === 4 && (this.isStrike() || this.isSpare()))
      || (this.currentroll === 3)
  )
}
```

src/Frame.js
```diff
@@ -0,0 +1,39 @@
+class Frame {
+  constructor() {
+    this.currentroll = 1;
+    this.rollOne = 0;
+    this.rollTwo = 0;
+    this.endFrame = false;
+    this.frameType = 'Normal';
```
alicelieutier, line 7: I think there is too much duplicated information here again.
You only really need the rolls, and maybe the currentroll (although you could get away with that too, by having the rolls initialised at null or undefined). Everything else can be calculated based on that information.

src/FinalFrame.js
alicelieutier: You could also have a `needsThirdRoll` method that returns `isSpare() || isStrike()` to simplify further.

src/Game.js
```diff
@@ -0,0 +1,80 @@
+class Game {
+	constructor(frame = new Frame()) {
+		this.frameNo = 1;
+		this.frames = [];
+		this.currentFrame = frame;
+		this.currentscore = 0;
+		this.gameOver = false;
```
alicelieutier, line 7: Too much info again.

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	nextFrame(frame = new Frame()) {
+		if (this.frameNo > 10) {
```
alicelieutier, line 11: For example, `frameNo` could re placed by `this.frames.length`

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+			return (this.gameOver = true);
+		} else {
+			this.frames.push(this.currentFrame);
+			this.frameNo++;
+			if (this.frameNo < 10) {
+				this.currentFrame = frame;
```
alicelieutier, line 17: `this.currentFrame` could be a method that returns`this.frames[this.frames.length - 1]`

src/Game.js
```diff
@@ -0,0 +1,80 @@
+class Game {
+	constructor(frame = new Frame()) {
```
alicelieutier, line 2: You may want to use class dependency injection rather than object dependency injection. Something like:
```js
constructor(frameClass = Frame, finalFrameClass = FinalFrame) {
  this.frameClass = frameClass;
  this.finalFrameClass = finalFrameClass;
  // ...
}
```

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+		this.frameNo = 1;
+		this.frames = [];
+		this.currentFrame = frame;
+		this.currentscore = 0;
+		this.gameOver = false;
+	}
+
+	nextFrame(frame = new Frame()) {
```
alicelieutier, line 10: This whole method could be much shorter:
```js
nextFrame() {
  if (this.frames.length < 9) {
    this.frames.push(new this.frameClass())
  } else if (this.frames.length === 9) {
    this.frames.push(new this.finalFrameClass())
  }
}
```

Something like that. I've kept the current frame within the frame array here - not sure why you are keeping it separate.

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	rollBall(value) {
+		this.currentFrame.roll(value);
+		if (this.currentFrame.endFrame) {
+			this.nextFrame();
+		}
+	}
+
+	score() {
```
alicelieutier, line 35: You should probably delegate score counting to frames. They could take bonuses, too.

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+				}
+				this.currentscore +=
+					this.frames[i]["rollOne"] + this.frames[i]["rollTwo"];
+			}
+		}
+		return this.currentscore;
+	}
+
+	isSpare(index) {
+		return this.frames[index]["frameType"] === "Spare";
```
alicelieutier, line 62: Why do you look into the state of the frame rather than call `isSpare` in the frame directly? Instance variable should be private as much as possible. So `this.frames[index].isSpare()`

src/Game.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	isStrike(index) {
+		return this.frames[index]["frameType"] === "Strike";
```
alicelieutier, line 66: same

alicelieutier (commented):
General comment on your code structure -
Delegating score counting to frames and not relying as much on instance variables are the two biggest improvements you can make.
I also showed how to do class dependency injection in the comments.

PR #1386, opened 2021-09-05
Week 6 Weekend Challenge - Bowling-CHallenge-JS

ClaimingCookie5:
I'm mostly happy with what I have written.
 I'm not happy with my total() function and I need to tests for more edge cases.

I would like feedback on what I can do differntly and how to improve on what I've already done.
It seems I haven't understood encapsulation and that would possibly help me refactor a lot of my functions.


src/Frame.js
```diff
@@ -0,0 +1,68 @@
...
+    this.#noNegativeNums(number);
+
+    this._bowls.push(number);
+    this.#resetFrameCount();
+  }
+
+  #noNegativeNums(number) {
+    if (number < 0) {
+      throw Error('Cannot input negative numbers');
+    }
```
PaulRainsford, line 29: Nice guard clause

spec/CalculateScoreSpec.js
```diff
@@ -0,0 +1,93 @@
...
+      addFrames(4, 10, 0);
+      addFrames(2, 5, 5);
+      addFrames(1, 0, 10);
+      addFrames(1, 7, 0);
+      calculateScore.total();
+      expect(calculateScore._score).toEqual(170);
+    });
+  });
+
+});
```
PaulRainsford, line 93: Comprehensive testing

spec/FrameSpec.js
```diff
@@ -0,0 +1,56 @@
...
+      frame.addBowl(5);
+      expect(frame.bowls()).toEqual([5, 5, 5]);
+    });
+
+    it('cannot add negative numbers to the _bowls array', () => {
+      expect( () => { frame.addBowl(-5); }).toThrow(new Error('Cannot input negative numbers'));
+    });
+  });
+
+});
```
PaulRainsford, line 56: Good test coverage including guard clause

src/CalculateScore.js
```diff
@@ -0,0 +1,52 @@
...
+          this._score += frame.reduce((a, b) => a + b);
+        }
+      } else {
+        this._score += frame.reduce((a, b) => a + b);
+      }
+      currentFrame += 1;
+    });
+    currentFrame = 0;
+  }
+
```
PaulRainsford, line 44: Good use of javascript functions including reduce and concat

src/Frame.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+  #newFrame() {
+    if (this.#scenario(1) && this.#compareArrLength(3)) {
+      return this.#incrementAndResetFrame();
+    }
+    if (this.#scenario(2) && this.#compareArrLength(2)) {
+      return this.#incrementAndResetFrame();
+    }
+  }
+
```
PaulRainsford, line 39: Nice use of # for private function

alicelieutier (commented):
This is not a full review, but an answer about how to improve your `total()` function:
* Your class structure does not totally help here. Your classes are separated out by what they do (calculating vs storing the roll data). In fact, classes are about storing data about some entity, and they should include some methodd to do something with that data. So separating storing and calculating does not really make sense from a OOP point of view.
Here, what that means is that your two classes are heavily coupled - you cannot change the implementation of one without changing the implementation of the other. And that also means that extracting functionalities even more would feel like just additional coupling with a third class.
Instead, think of different entities. You could have a class responsible for a single frame, that holds the rolls, possible bonuses, and calculates the score for a single frame, and then you could have the class responsible for a full game, which would hold all the frames, manage them (create new ones, pass bonuses around..), and calculate the total score, by delegating individual frame score to each of the frames.
That model is much more OOP - Each class is responsible for one entity, and does not need to know how the other one works. You should also find that the code becomes clearer, because responsibilities are well defined.

PR #1393, opened 2021-09-06
Bowling-challenge js 

PaulRainsford:


spec/SpecHelper.js
```diff
@@ -0,0 +1,15 @@
+beforeEach(function () {
```
ClaimingCookie5, line 1: You don't need this file, unless you write something that helps what you're doing specifically with you jasmine spec files.

spec/scoreSpec.js
```diff
@@ -0,0 +1,72 @@
...
+    describe ('secondBowlPins', () => {
+            it ('shows number of pins knocked down by second bowl', () => {
+                score.secondBowl(9);
+                expect(score.secondBowlPins).toEqual(9);
+            });
+    });
+
+    describe ('isStrike', () => {
+        it ('checks if bowl is a strike', () => {
+        score.firstBowl(10);
```
ClaimingCookie5, line 25: Need to indent consistently. Same as lines 33 - 36

PR #1397, opened 2021-09-06
PR for review

Battery0:
PR for review


src/game.js
```diff
@@ -0,0 +1,42 @@
...
+
+  _getPlayerRolls(roll_one, roll_two) {
+    let current_frame = this.frames[this.frames.length - 1]
+    current_frame.firstRoll(roll_one);
+    current_frame.secondRoll(roll_two);
+  }
+
+  _newFrame() {
+    if (this.frames.length <= 10) {
+      this.frames.push(new Frame);
```
ashtroboy79, line 32: Did you try dependancy injection here like you did on lines 3 & 4? I know i had trouble with getting it to work

src/frame.js
```diff
@@ -0,0 +1,27 @@
...
+    this._guardClauses(roll);
+    this.roll_one = roll
+  }
+
+  secondRoll(roll) {
+    this._guardClauses(roll);
+    this.roll_two = roll
+  }
+
+  _guardClauses(roll) {
```
ashtroboy79, line 18: nicely designed function, I only thought about validating user input late yesterday and i was too tired by then,

src/score.js
```diff
@@ -0,0 +1,19 @@
...
+  constructor() {
+    this.frameScore = 0;
+    this.totalScore = 0;
+  }
+
+  currentFrame(frame_rolls) {
+    this.frameScore = (frame_rolls.roll_one + frame_rolls.roll_two);
+  }
+
+  scoreSoFar(frames_played) {
```
ashtroboy79, line 12: this seems so much cleaner than my scoring method, didn't know about reduce, could you perhaps talk me through the calls as it seems almost recursive.

README.md
```diff
@@ -1,5 +1,59 @@
...
+[Project Planning Draft Notes Page 2](https://github.com/Battery0/bowling-challenge/blob/main/public/planning2.jpg)
+
+## To Do
+
+1. Account for spares bonus points
+2. Account for strike bonus points
+3. Include stubs & doubles to isolate tests
+4. Add test for console output on 'game over'
+
+## Getting started
```
ashtroboy79, line 20: nice clear Readme instructions, totally forgot to add those in mine. Nice clear classes, each with a single responsibility, and the tests are clear in what they do.  I think if given some more time you could easily write scoring for spares and strikes, as you are already storing the rolls for each frame. As you mentioned doubling/stubbing I can see that you only need it for the score class really.

PR #1399, opened 2021-09-06
DR JS bowling challenge

danroche10:


spec/unit/frameSpec.js
```diff
@@ -0,0 +1,45 @@
...
+      expect(frame._rolls.length).toEqual(1);
+    });
+  });
+
+  describe("adds bonus score to bonus score variable", function () {
+    it("correctly updates bonus score", function () {
+      frame.addBonusScore(5);
+      expect(frame._bonus_score).toEqual(5);
+    });
+  });
```
alicelieutier, line 26: Don't test state. It's like private methods.
You should only test the public interface of your class, so that tests do not fix your code into a specific implementation.

spec/unit/frameSpec.js
```diff
@@ -0,0 +1,45 @@
...
+
+  describe("updates instance variable to frame total", function () {
+    it("updates frame total to correct amount", function () {
+      frame.addRoll(5);
+      frame.addRoll(5);
+      frame.addBonusScore(5);
+      expect(frame.calcFrameTotal()).toEqual(15);
+    });
+  });
```
alicelieutier, line 44: These are the same test twice? It would be nice to see a few different frames in the tests here. Maybe a normal one, a spare and a strike.

spec/unit/gameSpec.js
```diff
@@ -0,0 +1,86 @@
...
+      addRoll: () => {},
+      addBonusScore: () => {},
+      calcFrameTotal: () => {},
+      setFrameTotal: () => {},
+      isFinalFrameComplete: () => {},
+      isFrameComplete: () => {},
+      isStrike: () => {},
+      isSpare: () => {}
+    };
+    spyOn(frameFactory, "createFrame").and.returnValue(frame);
```
alicelieutier, line 16: This is good.

spec/unit/gameSpec.js
```diff
@@ -0,0 +1,86 @@
...
+    frame = { 
+      addRoll: () => {},
+      addBonusScore: () => {},
+      calcFrameTotal: () => {},
+      setFrameTotal: () => {},
+      isFinalFrameComplete: () => {},
+      isFrameComplete: () => {},
+      isStrike: () => {},
+      isSpare: () => {}
+    };
```
alicelieutier, line 15: Do you really need all these methods?

spec/unit/gameSpec.js
```diff
@@ -0,0 +1,86 @@
...
+
+  describe("new instancs of game class", function () {
+    it("frame array is empty for frame instance", function () {
+      expect(game.frames).toEqual([]);
+    });
+
+    it("have current frame set to frame passed in", function () {
+      expect(game.currentFrame).toEqual(frame);
+    });
+  });
```
alicelieutier, line 28: These seem like implementation/state tests. I don't think the current Frame should be part of the public interface.

spec/unit/gameSpec.js
```diff
@@ -0,0 +1,86 @@
...
+
+  describe("method for each bowl", function () {
+    it("correctly adds 1 frame to frames array", function () {
+      spyOn(game, "_isNewFrame").and.returnValues(false, true);
```
alicelieutier, line 32: Spying / stubbing a method of the class under test is not good. You only should stuck/mock external classes.

spec/unit/gameSpec.js
```diff
@@ -0,0 +1,86 @@
...
+      game.roll(5);
+      game.roll(4);
+      expect(game.frames.length).toEqual(1);
```
alicelieutier, line 35: This is a state/implementation test.

spec/unit/gameSpec.js
```diff
@@ -0,0 +1,86 @@
+describe("Game", function () {
```
alicelieutier, line 1: A number of problems in this file, with some good attempts. I'm not commenting everywhere, but try to think about whether my comments apply to different tests or not. Happy to chat to clarify if needed.

src/Game.js
```diff
@@ -0,0 +1,125 @@
+class Frame {
+  constructor () {
+    this._rolls = [];
+    this._bonus_score = 0;
+  }
+
+  isStrike() {
+    return (this._rolls[0] == 10);
```
alicelieutier, line 8: nit - use `===` instead of `==`

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  isSpare() {
+    return (this._rolls[0] + this._rolls[1]== 10);
+  }
+
+  isFrameComplete() {
+    if (this.isStrike()) { return true; }
```
alicelieutier, line 16: This may be enough to work, but it feels like also checking that there's one roll would be good? Happy to get pushback on this.

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+    else if (this._rolls.length == 2 ) {return true;}
+  }
+
+  isFinalFrameComplete() {
+    if (this._rolls.length == 3) {return true;}
```
alicelieutier, line 21: Surely the final frame sometimes only needs two rolls.

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+}
+
+class frameFactory {
+  static createFrame() {
+     return new Frame();
+  }
+}
+
+class Game {
+  constructor () {
```
alicelieutier, line 52: The frameFactory should be passed as a parameter if you want to use it for dependency injection.

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+    this.currentFrame.addRoll(score);
+    if (this._isNewFrame()) {this._completeFrame();}
+  }
+
+  scoreTotal() {
+    let score = 0;
+    this.frames.forEach(frame => {
+      score += frame.calcFrameTotal();
+  });
+  return score;
```
alicelieutier, line 67: nit - indentation

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  _isDoubleStrike() {
+    return (this.frames.length > 1 && this.frames.slice(-1)[0].isStrike() && this.frames.slice(-2)[0].isStrike());
+  }
+
+  _isLastFrameStrike() {
+    if (this.frames.slice(-1)[0].isStrike()) { return true; }
+    else {return false;}
+  }
```
alicelieutier, line 104: Simplify the code here into:
```js
return (this.frames.slice(-1)[0].isStrike())
```

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+
+  _isLastFrameSpare() {
+    if (this.frames.slice(-1)[0].isSpare()) { return true; }
```
alicelieutier, line 107: same

src/Game.js
```diff
@@ -0,0 +1,125 @@
...
+  scoreTotal() {
+    let score = 0;
+    this.frames.forEach(frame => {
+      score += frame.calcFrameTotal();
+  });
+  return score;
+  }
+
+  _updateFrameTotals() {
+    if (this._isNotFirstFrame()) { this._addAnyBonuses();}
```
alicelieutier, line 71: I think the check to know whether it's the first frame or not would be better pushed down further towards where it's important to know that.
Here, this check is only important because you later try to know if the last frame was a spare or strike. If your `_isLastFrameStrike` and `_isLastFrameSpare` methods were more robust, they would not give an error when called while we are still on the first frame. Instead, they could just say false (if first frame, then last frame was not a spare/strike). This way, you save yourself this check here, and an extra method.

alicelieutier: If you do that, then you can simplify further, as your method `_updateFrameTotals` would then only be calling `this._addAnyBonuses()`, which means you could just call that directly, and remove the `_updateFrameTotals` method.

alicelieutier (commented):
Overall this is much better design than last time  saw your code. It's well separated and easier to read.
You can refactor further still and simplify the code more (bowling is always a refactor challenge).
It's good to see feature tests. The unit tests are not there yet though - they could be more thorough and test more behaviour. See my comments below.

spec/frameSpec.js
```diff
@@ -0,0 +1,52 @@
...
+  beforeEach(function () {
+    frame = new Frame();
+  });
+
+  describe("any new instances of Frame class have empty rolls array", function () {
+    it("rolls array is empty for frame instance", function () {
+      expect(frame._rolls).toEqual([]);
+      expect(frame._bonus_score).toEqual(0);
+    });
+  });
```
alicelieutier, line 13: This test should just be removed entirely.

spec/frameSpec.js
```diff
@@ -0,0 +1,52 @@
...
+
+  describe("expects to capture when a strike has been scored", function () {
+    it("isStrike returns 'true' when a 10 is rolled", function () {
+      frame.addRoll(10);
+      expect(frame.isStrike()).toEqual(true);
+    });
+  });
+});
```
alicelieutier, line 52: Other tests are great, but you could add more.
Think of edge cases. For example, would a roll that is a strike return true for isSpare? It's good to think of unhappy cases, not just happy ones.

spec/gameSpec.js
```diff
@@ -1,70 +1,53 @@
...
-      addBonusScore: () => {},
       calcFrameTotal: () => {},
-      setFrameTotal: () => {},
-      isFinalFrameComplete: () => {},
       isFrameComplete: () => {},
+      isFinalFrameComplete: () => {},
       isStrike: () => {},
       isSpare: () => {}
     };
     spyOn(frameFactory, "createFrame").and.returnValue(frame);
```
alicelieutier: This is great.

src/Frame.js
```diff
@@ -1 +1,44 @@
...
+  }
+
+  calcFrameTotal() {
+    return this._rolls.reduce((a, b) => a + b, 0) + this._bonus_score;
+  }
+
+  isFrameComplete() {
+    if (this.isStrike() && this._rolls.length === 1) { return true; }
+    else if (this._rolls.length === 2 ) {return true;}
+  }
```
alicelieutier, line 22: You function returns `true` or `undefined`. It would be cleaner to return `true` or `false`

alicelieutier: A test that looks at unhappy paths would find this.

src/Frame.js
```diff
@@ -1 +1,44 @@
...
+
+  isFinalFrameComplete() {
+    if (this._rolls.length === 2 && this.calcFrameTotal() < 10) {return true;}
+    else if (this._rolls.length === 3) {return true;}
+  }
```
alicelieutier, line 27: Same as above.

src/Game.js
```diff
@@ -1,57 +1,8 @@
...
-
-class frameFactory {
-  static createFrame() {
-     return new Frame();
-  }
-}
-
 class Game {
-  constructor () {
+  constructor (frameFactory) {
```
alicelieutier, line 2: You could have a default value here, so that you would only pass an argument when mocking. eg:
```js
constructor (frameFactory = FrameFactory) {
```

src/frameFactory.js
```diff
@@ -1 +1,5 @@
-// frameFactory class to go here
\ No newline at end of file
+class frameFactory {
```
alicelieutier, line 1: classes should have names starting with a capital.

alicelieutier (commented):
This is really good. Such an improvement since the first version!
I left a few comments and also happy to have a chat about testing.

log722 (approved):

PR #1401, opened 2021-09-06
Bowling 

BlessingUb:


sa-mcquanzie (commented):
I think this is a really lovely, simple solution and beautiful & clear Javascript!

I only have two suggestions.

First, I think you've tested perfectly for all the edge-cases in a legal game, so you might want to now write some tests for illegal games. What happens if I try to roll a 4 and a 9 in the same frame? What happens if I try to score a game with too many rolls, etc.

Second, and this is just a tiny thing, maybe you want to move that rollMany() function in your BowlingGameSpec.js into a SpecHelper file?

PR #1404, opened 2021-10-03
bowling challenge in JS 

Maby0:


spec/frameSpec.js
```diff
@@ -0,0 +1,145 @@
+describe('Frame', () => {
+  describe('constructor', () => {
+    it('following values are set upon initialisation', () => {
+      let frame = new Frame;
+      expect(frame.roll1).toEqual(null);
+      expect(frame.roll2).toEqual(null);
+      expect(frame.frameScore).toEqual(null);
+      expect(frame.strike).toEqual(false);
+      expect(frame.spare).toEqual(false);
+      expect(frame.frameFinished).toEqual(false);
```
alicelieutier, line 10: Don't test values after initialisation - that's not behaviour :)

spec/frameSpec.js
```diff
@@ -0,0 +1,145 @@
...
+
+  describe('bowl', () => {
+    describe('first frame', () => {
+      let frame;
+
+      beforeEach(() => {
+        frame = new Frame;
+        spyOn(window, 'prompt').and.returnValue(1);
+        frame.bowl();
+        frame.bowl();
```
alicelieutier, line 22: This is interesting... you class reads the prompt directly? It's useful typically to separate the logic and the user interface. It also makes things easier to test.

spec/frameSpec.js
```diff
@@ -0,0 +1,145 @@
...
+      })
+  
+      it('requests a score for first roll if first roll is null', () => {
+        expect(frame.roll1).toEqual(1);
+      })
+  
+      it('requests a score for second roll if first roll is not null', () => {
+        expect(frame.roll2).toEqual(1);
+      })
```
alicelieutier, line 31: Do these methods (`roll1`, `roll2`) exist for the interface? Or is there another reason they are surfaced? It feels like the inside of the Frame is not super encapsulated if all of it is made to be accessed.

spec/frameSpec.js
```diff
@@ -0,0 +1,145 @@
...
+        spies.and.returnValue(10);
+        Scorecard.startGame("Mabon");
+        for (let i = 0; i < 21; i++) {
+          Scorecard.bowl();
+        }
+        expect(Scorecard.currentGame.score()).toEqual(300);
+      })
+    })
+  })
+})
```
alicelieutier, line 145: These should be in a separate file called "featureSpec".
Good feature tests, though ;)

spec/scorecardSpec.js
```diff
@@ -0,0 +1,67 @@
+describe('Scorecard', () => {
+  let scorecard;
+
+  describe('constructor', () => {
+    it('has a name and empty array upon initialisation', () => {
+      let scorecard = new Scorecard("Mabon");
+      expect(scorecard.name).toEqual("Mabon");
+      expect(scorecard.frames).toEqual([]);
+    })
+  })
```
alicelieutier, line 10: Same as the other one.

spec/scorecardSpec.js
```diff
@@ -0,0 +1,67 @@
...
+
+  describe('score', () => {
+    it('returns total running score', () => {
+      frame = new Frame;
+      frame.frameScore = 5;
```
alicelieutier, line 15: This is not isolated.

spec/scorecardSpec.js
```diff
@@ -0,0 +1,67 @@
...
+        Scorecard.currentGame.frames[Scorecard.currentGame.frames.length - 1].frameFinished = true;
+      }
+      expect(Scorecard.bowl()).toEqual(`The game has ended. Your final score was 45.`);
+    })
+
+    it('adds a frame to frames array if last frame has finished', () => {
+      Scorecard.startGame("Mabon");
+      Scorecard.currentGame.frames[Scorecard.currentGame.frames.length - 1].frameFinished = true;
+      Scorecard.bowl();
+      expect(Scorecard.currentGame.frames.length).toEqual(2);
```
alicelieutier, line 64: This doesn't feel like a behaviour test.
Overall, I have found it difficult to understand the behaviour/interface of your class. A domain model may be useful here to clarify them.

src/frame.js
```diff
@@ -0,0 +1,127 @@
+class Frame {
+  constructor() {
+    this.roll1 = null;
+    this.roll2 = null;
+    this.roll3 = null;
+    this.frameScore = null;
+    this.strike = false;
+    this.spare = false;
+    this.frameFinished = false;
```
alicelieutier, line 9: There is information duplication in your class. The score, strike, spare and finished could be calculated on the fly based on the rolls. Calculating on the fly is better, because it avoids information getting out of sync between variables, and ultimately bugs are less likely.

alicelieutier: So basically, this class should only have rolls 1 to 3 as instance variables, everything else should be methods. here's one example of such a method:
```js
strike() {
  return this.roll1 === 10;
}
```

alicelieutier: Basically, it's almost like reversing the flow. Here you are receiving a roll, and setting all variables. The other way around you only set the roll, and then everything is calculated when you need it / request it. Potentially less computing to do, tot.

src/frame.js
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+      return Scorecard.currentGame._getPreviousFrame().strike;
+    }
+  }
+
+  _check2ndPreviousFrameForStrike() {
+    if (Scorecard.currentGame._get2ndPreviousFrame() === undefined) {
+      return false;
+    } else {
+      return Scorecard.currentGame._get2ndPreviousFrame().strike;
```
alicelieutier, line 124: You have a circular dependency here, where the Scorecard depends on Frame and the Frame depends on the Scorecard. These are best avoided, and typically that means your classes are very coupled. Is there a way the scorecard can pass the correct previous frame to the frame, instead of the frame reading from the scorecard?

alicelieutier (commented):
Good attempt, but a few oddities in your code organisation that is making testing and code understanding more difficult.

PR #1406, opened 2021-10-03
Bowling Game week 6

leahcolleen:


src/BowlingGame.js
```diff
@@ -0,0 +1,3 @@
+var BowlingGame = function () {
+
+};
```
lildann, line 3: You could make BowlingGame a class, and then write a bowl function within that class to store properties of the game (such as an array of bowls) and then access those for scoring.

lildann (commented):
Good setup of the project with initial jasmine test written before code. A README.md including details for what you plan to do next would be good, and help you to think about the next stage.

PR #1407, opened 2021-10-04
Bowling Challenge

Ifrahhssn:


src/bowling.js
```diff
@@ -0,0 +1,49 @@
...
+  roll(pins) {
+    this.rolls.push(pins);
+  }
+
+  score() {
+    let result = 0;
+    let rollIndex = 0;
+    let game = this;
+
+    for (let frame = 0; frame < 10; frame++) {
```
JohnForster, line 15: It would be good to store 10 in a constant somewhere, so that it's clear what its meaning is.

src/bowling.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    function getSpareScore() {
+      return game.rolls[rollIndex] + game.rolls[rollIndex + 1] + game.rolls[rollIndex + 2];
+    }
+
+    function getNormalScore() {
+      return game.rolls[rollIndex] + game.rolls[rollIndex + 1];
+    }
+  }
```
JohnForster, line 48: You have defined these functions within the score method so that you have access to the current `rollIndex`. Could you extract them to the class, and pass the `rollIndex` in as an argument instead? eg. 
```js
    getStrikeScore(rollIndex) {
      return game.rolls[rollIndex] + game.rolls[rollIndex + 1] + game.rolls[rollIndex + 2];
    }
```

spec/bowlingSpec.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it("can roll a strike with bonus", () => {
+    game.roll(10);
+    game.roll(7);
+    game.roll(2);
+    rollMany(0, 16);
+    expect(game.score()).toBe(28);
+  });
+
```
JohnForster, line 38: It would be nice to include one test which tests a "real" game of bowling, including rolling 0s, strikes, spares together.

JohnForster (commented):
Overall this is a really nice solution, with a very clean process to calculate the score. The only thing that feels like it could potentially be more robust is the usage of the rollIndex to work out where in the rolls we are. If we wanted to add a method that printed out the game history so far, it wouldn't be clear which rolls were part of which frames.

My suggestion would be to either group rolls together by their frames, with either an array of arrays, or a Frame class. The advantage to using a Frame class is that you can shift some of the scoring logic into that class, such as working out if that frame is a strike, spare etc. and the logic for the bonuses.

Overall though this is a very nice solution, well done 👍

PR #1408, opened 2021-10-04
Makers weekend challenge, week 6

lildann:
The challenge gave me some good practice writing Javascript. I used my Ruby code from the previous week as a framework, but got stuck on the tenth frame calculations, so it is incomplete. 
Ideally I would have split my Game class into two, with a Frame class, but I got a bit stuck on how to do this.

### DIFF
--- a/jasmine/spec/gameSpec.js
+++ b/jasmine/spec/gameSpec.js
@@ -12,37 +12,38 @@ describe('Game', () => {
     expect(game.runningScore).toEqual(0);
   });
 
-  it('adds the roll together', () => {
-    game.roll(3, 4)
-    expect(game.runningScore).toEqual(7);
+  describe('Scoring', () => { 
+    it('adds the roll together', () => {
+      game.roll(3, 4)
+      expect(game.runningScore).toEqual(7);
+    });
+
+    it('adds all 10 rolls together', () => {
+      for (let i = 0; i < 10; i++) {
+        game.roll(3, 4);
+      }
+      expect(game.runningScore).toEqual(70);
+    });
+
+    it('gutter game equals zero', () => {
+      for (let i = 0; i < 10; i++) {
+        game.roll(0, 0);
+      }
+      expect(game.runningScore).toEqual(0);
+    });
   });
 
-  it('adds all 10 rolls together', () => {
-    for (let i = 0; i < 10; i++) {
-      game.roll(3, 4);
-    }
-    expect(game.runningScore).toEqual(70);
+  describe('Spares and strikes scoring', () => { 
+    it('calculates score for spare', () => {
+      game.roll(5, 5);
+      game.roll(4, 2);
+      expect(game.runningScore).toEqual(20)
+    });
+  
+    it('calculates score for strike', () => {
+      game.roll(10, 0);
+      game.roll(4, 4);
+      expect(game.runningScore).toEqual(26)
+    });
   });
-
-  it('gutter game equals zero', () => {
-    for (let i = 0; i < 10; i++) {
-      game.roll(0, 0);
-    }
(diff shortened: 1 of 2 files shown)


jasmine/spec/gameSpec.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+describe('Game', () => {
+
+  let game;
+
+  beforeEach(() => {
```
leahcolleen, line 7: Good refactoring to prevent the repetition of game = new Game

jasmine/spec/gameSpec.js
```diff
@@ -0,0 +1,49 @@
...
+    it('calculates score for spare', () => {
+      game.roll(5, 5);
+      game.roll(4, 2);
+      expect(game.runningScore).toEqual(20)
+    });
+  
+    it('calculates score for strike', () => {
+      game.roll(10, 0);
+      game.roll(4, 4);
+      expect(game.runningScore).toEqual(26)
```
leahcolleen, line 46: Clearly you have thought thoroughly about the logic of the game as it is very well constructed. The only thing I would suggest to think about is how you would construct the user interface.

jasmine/src/game.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+class Game {
+  constructor(runningScore, bowlCard, scoreCard, index) {
+    this.runningScore = 0;
+    this.bowlCard = [];
+    this.scoreCard = [];
```
JohnForster, line 7: There's a concept in programming called "single source of truth" which is where you avoid duplicating state wherever possible. This has a number of benefits, as it prevents any chance of the two variables getting out of sync, which ultimately reduces the chance of bugs. It also makes it easier to add functionality to our code. For example, if we wanted to add the functionality to edit a previous roll that had been entered incorrectly, currently we would need to update three different variables in different ways, whereas if the source of truth was just the rolls, and everything else was calculated from those each time, we would only need to change that piece of data, and we could be confident that the class would treat it like it had always been that way.

In the case of this class, you have the rolls, frame scores and runningScore stored separately, when the last two are derived from the first. Can you think of a way to simplify this so that you only need to store the rolls, and the scores can be calculated on the fly as needed?

The same goes for the index. As far as I can see, the index refers to how many frames have so far been entered. Could you use this, instead of storing it in its own instance variable?

jasmine/src/game.js
```diff
@@ -0,0 +1,52 @@
...
+    return bowl1 + bowl2;
+  }
+ 
+  calculateSparePoints() {
+    let sparePoints = this.bowlCard[this.index][0];
+    if (this.scoreCard[this.index-1] === 10 && this.bowlCard[this.index-1][0] !== 10) {
+      this.scoreCard[this.index-1] += sparePoints;
+      this.runningScore += sparePoints;
+    } return;
+  }
```
JohnForster, line 26: Methods like this can be hard to read, because it's not immediately obvious what something like `this.bowlCard[this.index-1][0]` refers to. It would be good to name these with variables to make it easier to read, eg
```js
const previousFrameRolls = this.bowlCard[this.index-1]
const previousFrameTotal = this.scoreCard[this.index-1]
if (previousFrameTotal === 10 && previousFrameRolls[0] !== 10) {...}
```

jasmine/src/game.js
```diff
@@ -0,0 +1,52 @@
...
+    this.calculateSparePoints();
+  }
+
+}
+ 
+let game = new Game();
+
+console.log(game.roll(10, 0));
+console.log(game.roll(4, 4));
+console.log(game);
```
JohnForster, line 50: Since this is a script separate from the implementation of the class, it probably should be in its own file.

PR #1422, opened 2021-11-01
Bernard's Bowling Challenge

bernardleanse:
Much simpler approach to my project in ruby, could do with some class extraction.


bowlingGame.js
```diff
@@ -0,0 +1,65 @@
...
+  constructor(){
+    this.rolls = [];
+    this.currentFrame = 1;
+  };
+  roll(number){
+    if(number > 10){
+      throw "Invalid Roll";
+    }
+    else {
+      this.rolls.push( { value: number, frame: this.currentFrame } );
```
TTurvey, line 11: This is a nice way of keeping the rolls assigned to a frame.

bowlingGame.js
```diff
@@ -0,0 +1,65 @@
...
+
+  sumTheRest(currentRoll){
+    let sum = 0;
+    this.rolls.slice(currentRoll).forEach(roll => {
+      sum += roll.value
+    });
+    return sum;
+  };
+  frameIsSpare(currentRoll){
+    return this.rolls[currentRoll].value + this.rolls[currentRoll + 1].value == 10;
```
TTurvey, line 52: Let's say that the second roll in a frame was 5 and the first roll in the next frame was 5 also. Would the frameIsSpare method incorrectly think this is a somehow a spare?
I may be wrong but the frameIsSpare method doesn't seem to reference the frame that the 'currentRoll' is in so I think it may get the mentioned scenario wrong.

bowlingGame.js
```diff
@@ -0,0 +1,65 @@
...
+        totalScore += this.sumTheRest(currentRoll);
+        break;
+      }
+      else if(this.frameIsStrike(currentRoll)){       
+        totalScore += this.bonusScoring(currentRoll);
+        currentRoll += 1;
+        this.currentFrame += 1
+      }
+      else if(this.frameIsSpare(currentRoll)){
+        totalScore += this.bonusScoring(currentRoll);
```
TTurvey, line 30: There is one bonusScoring method for both the strike and spare scenarios. It looks like the spare bonus is then incorrect because you're adding the values of the next two rolls in the bonusScoring method. The spare bonus should only be the first roll in the next frame.

bowlingGame.js
```diff
@@ -0,0 +1,65 @@
...
+  score(){
+    let totalScore = 0;
+    let currentRoll = 0;
+    while(currentRoll < this.rolls.length){
+      if(this.currentFrame == 10){
+        totalScore += this.sumTheRest(currentRoll);
+        break;
+      }
+      else if(this.frameIsStrike(currentRoll)){       
+        totalScore += this.bonusScoring(currentRoll);
```
TTurvey, line 25: I think that this bonusScoring method is correct for the strike scenario. I like how it uses 'this.rolls[currentRoll + 1].value' to access the next rolls value. I'll probably end up trying to implement something like this in my code.

bowlingGame.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+describe(BowlingGame, () => {
+  beforeEach(() => {
+    this.game = new BowlingGame();
+  });
+  it('handles a gutter game', () => {
+    for(i = 0; i < 20; i++) {
+      this.game.roll(0);
+    };
+    expect(this.game.score()).toBe(0)
```
TTurvey, line 12: Good test for the gutter game scenario

bowlingGame.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+    expect(this.game.score()).toBe(20);
+  });
+  it('handles two strikes in a row', () => {
+    this.game.roll(10);
+    this.game.roll(10);
+    this.game.roll(4);
+    this.game.roll(3);
+    for(i=0; i<14; i++){
+      this.game.roll(0);
```
TTurvey, line 48: I like how you've tested many scenarios here to make sure there aren't any unexpected errors, well done.

bowlingGame.test.js
```diff
@@ -0,0 +1,85 @@
...
+    this.game.roll(10);
+    this.game.roll(5);
+    this.game.roll(5);
+    this.game.roll(10);
+    this.game.roll(5);
+    this.game.roll(3);
+    for(i=0; i<12; i++){
+      this.game.roll(0);
+    };
+    expect(this.game.score()).toBe(66);
```
TTurvey, line 63: I would be interested in seeing another test for my comments on the bowlingGame.js file. I've written a suggested test for this below. If the game works then it shouldn't think that the back to back 5s is a spare.

`it('handles mix of strikes and spares', () => {
    this.game.roll(1);
    this.game.roll(5);
    this.game.roll(5);
    this.game.roll(1);
    this.game.roll(2);
    this.game.roll(2);
    for(i=0; i<12; i++){
      this.game.roll(0);
    };
    expect(this.game.score()).toBe(16);`

TTurvey (commented):
Well done here Bernard. I think the code was written well and obeyed the SRP (although in it's in one class which I don't think matters to be honest). My comments are mainly about the way the bonuses are calculated.
1) I wasn't sure how the spare bonus and strike bonus were calculated differently.
2) I also wasn't sure if rolls from different frames adding to make 10 would cause an incorrect spare.

In conclusion, this is a great effort - be proud.

PR #1423, opened 2021-11-01
Bowling JS challenge

AJOsmaston:
Started refactoring, functionality working but not cleaned up


lib/bowling.js
```diff
@@ -0,0 +1,79 @@
+const MAX_PINS = 10
```
arthurfincham, line 1: Nice storing 10 pins in a constant - need to do this myself!

PR #1424, opened 2021-11-01
Bowling Challenge - JS

dpkerins:


hamstercat007:
Looks like you have a good understanding of conversion in JS, with correctly written instances, methods and tests.

PR #1426, opened 2021-11-01
bowling-challenge javascript

TTurvey:
Partial solution.
Bonuses not implemented fully yet.


spec/frame.test.js
```diff
@@ -0,0 +1,27 @@
...
+  it("returns the second roll of the frame", function(){
+    var next = new Frame([0,0]);
+    expect(frame.getSecondRoll()).toEqual(2);
+  });
+
+  it("returns the total of two rolls in a frame", function(){
+    var next = new Frame([0,0]);
+    expect(frame.calcRollsTotal()).toEqual(3);
+  });
+
```
bernardleanse, line 26: tests look good to me! only thing I don't get is the creating the var next frame, when is it used?

spec/game.test.js
```diff
@@ -0,0 +1,38 @@
...
+    expect(game.getRollsTotals()).toEqual([3]);
+  });
+
+  it("adds the Strike bonus to the bonuses array", function(){
+    var game = new Game();
+    game.enterFrame(10,0);
+    game.enterFrame(1,2);
+    expect(game.getBonuses()).toEqual([3,0,0,0,0,0,0,0,0,0,0,0]);
+  });
+
```
bernardleanse, line 24: this test seems well written to me, the bonus position correlates to which frame the bonus adds to. will pass when you've got get bonuses to work correctly. Not sure if there might be some testing of state here rather than behaviour but I don't really get what defines testing state so can't be sure, something for me to look into

docs/game.js
```diff
@@ -0,0 +1,84 @@
...
+
+  calcFrameBonuses(){
+    for (var i=0; i < this.frames.length; i++){
+      if (i % 2 === 0 && this.frames[i] === 10)
+        this.addStrikeBonus(i);
+      else if (this.rollsTotals[i] === 10)
+        this.addSpareBonus(i);
+    }
+  }
+
```
bernardleanse, line 45: I don't quite understand the %2, maybe a comment or a small private method extraction could make it easier to read? making bowling logic easy to read is definitely a nightmare!!

PR #1429, opened 2021-11-28
BC in JS

ginnyamazed:
I am not proud of this project or the state it is in. 
I did not include my planning notes in my readme file.  In them I wrote out how scoring worked.
Even with this, I was unable to implement diagrams to have a clear idea of my classes and the functions they should have.
I attempted to start with a pre-done array of frames, however, when it came to pull in the array and use it, I was unable to think of how to process it correctly and quickly went back to allowing user input.  
Once I started to accept user input I was unable to get my tests to pass because I have not figured out how to use mocking in jest.   I spent the last 3-4 hours of work on the program running the files in node to check function and output and make adjustments as needed.
I can have a user input 9 out of 10 frames and receive the appropriate error messages if they try to enter more than 10 pins for a frame.
However, I failed in getting the scoring to work.  I stopped after struggling with how to return the index and value of my array.  The intention was to find the arrays of strikes (i.e. value of 10 on even indexes) and then try to have it use the next 2 indexes (not always accurate for strikes but a start in the right direction) to provide an updated score based on the strike bonus rules.  If I was able to get this working, I would have implemented something similar for spares (when the total of 2 indexes = 10).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,42 @@
+{
+  "name": "bowling-challenge",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "bowling-challenge",
+      "version": "1.0.0",
+      "license": "ISC",
+      "dependencies": {
+        "readline": "^1.3.0",
+        "readline-sync": "^1.4.10"
+      }
+    },
+    "node_modules/readline": {
+      "version": "1.3.0",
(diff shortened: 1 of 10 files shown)


mhbae-dev:
The pull request is very clear in indicating the state/problem areas of the project. The project set up shows good understanding of JS classes and functions from learning over the week. The project would have benefitted from more time, working through the relationships between each of the created classes and their use cases. 

-Further reading and study required on the following:
Jest mocking

PR #1432, opened 2021-11-29
Tatiana - Bowling Challenge JS 

tatiananantes:
I have just implemented game logic but no UI/ command line. There is also some refactoring I'd like to do separating isSpare/is Strike methods.

### DIFF
--- a/game.js
+++ b/game.js
@@ -11,25 +11,29 @@ class Game {
 
   get points() {
     let points = 0;
-    let rollIndex =0;
+    let rollIndex = 0;
 
     // calculates points per game frames(therefore 10 is the iteration value)
-    for (let frame = 0; frame < 10; frame ++) {
-      const framePoints = this.rolls[rollIndex] + this.rolls[rollIndex + 1]; 
-        
-        if (framePoints === 10) {
-          points += 10 + this.rolls[rollIndex + 2]
-        } else {
-            points += framePoints;
-        }
-        rollIndex += 2;
-    }
-      return points;   
+    for (let frame = 0; frame < 10; frame++) {
+      
+      // isStrike?
+      if(this.rolls[rollIndex] === 10) {
+        points += 10 + this.rolls[rollIndex + 1] + this.rolls[rollIndex + 2];
+        rollIndex ++;
+        continue;
+      } 
+
+      const framePoints = this.rolls[rollIndex] + this.rolls[rollIndex + 1];  
+      
+      if (framePoints === 10) {
+        points += 10 + this.rolls[rollIndex + 2];
+      } else {
+        points += framePoints;
+      }
+      rollIndex += 2;
+    } 
+      return points; 
   }
-
-
-
-
 
 }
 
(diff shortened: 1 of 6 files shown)


game.test.js
```diff
@@ -0,0 +1,60 @@
+// const { it } = require('jest-circus');
+const Game = require('./game');
+
+let game;
+
+beforeEach(() => {
```
bmargerison, line 6: nice use of before

game.js
```diff
@@ -0,0 +1,40 @@
...
+  constructor() {
+    this.rolls = [];
+  }
+  
+  
+  roll(pins){
+    this.rolls.push(pins);
+  }
+
+  get points() {
```
bmargerison, line 12: very succinct logic, nice!

game.js
```diff
@@ -0,0 +1,40 @@
+class Game {
+
+  constructor() {
+    this.rolls = [];
```
bmargerison, line 4: perhaps could refactor to have a frame class? currently no concept of a frame, only rolls. having a frame class/frame objects could allow you to build onto the logic going forward (even if its not CLI)

game.js
```diff
@@ -0,0 +1,40 @@
...
+    for (let frame = 0; frame < 10; frame++) {
+      
+      // isStrike?
+      if(this.rolls[rollIndex] === 10) {
+        points += 10 + this.rolls[rollIndex + 1] + this.rolls[rollIndex + 2];
+        rollIndex ++;
+        continue;
+      } 
+
+      const framePoints = this.rolls[rollIndex] + this.rolls[rollIndex + 1];
```
bmargerison, line 26: I'm not sure if this works for the last frame. If a strike was rolled on the first shot of frame 10, we then shouldn't be adding the 2nd and 3rd shots for example (they are only added as bonus). Essentially I'm unsure if your logic fully deals with frame 10 (I might be wrong though)

game.test.js
```diff
@@ -0,0 +1,60 @@
...
+  game = new Game;
+});
```
bmargerison, line 8: try doing a test for all spares perhaps, which scores 190 (roll of 9 followed by 1 for every frame, with the third roll in the final frame being 9)

game.test.js
```diff
@@ -0,0 +1,60 @@
...
+    for(let i = 0; i < 20; i++) {
+      game.roll(0); 
+    }
+    expect(game.points).toEqual(0);
+  });
+});
+
+describe('all ones', () => {
+  it('calculates all ones', () => {
+    for(let i = 0; i < 20; i++) {
```
bmargerison, line 21: nice use of loops

game.test.js
```diff
@@ -0,0 +1,60 @@
...
+const Game = require('./game');
+
+let game;
+
+beforeEach(() => {
+  game = new Game;
+});
+
+describe('gutter game', () => {
+  it('calculates a gutter game', () => {
```
bmargerison, line 11: perhaps all the tests could be in 1 describe block?

PR #1434, opened 2021-11-29
refactored from ruby bowling challenge

mhbae-dev:
This bowling challenge has been refactored from Ruby to JavaScript. All tests passing but did not have time to check for the 10th frame to have bonus rounds. 

No user interface is working.

### DIFF
--- a/bowling.js
+++ b/bowling.js
@@ -1,0 +1,50 @@
+class Bowling {
+  constructor() {
+    this.rolls = [];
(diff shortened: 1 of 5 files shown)


bowling.js
```diff
@@ -0,0 +1,50 @@
+class Bowling {
```
ginnyamazed, line 1: good, clean, concise code

PR #1438, opened 2022-01-09
Weekend Challenge - Bowling JS, Chris Thomas

Hereloss:
Completed bowling challenge in JS, includes rl input file (cli.js) and the bowling.js file is fully tested using Jest

### DIFF
--- a/lib/bowling.js
+++ b/lib/bowling.js
@@ -1,0 +1,95 @@
+class Bowling {
+
+  constructor() {
+    this.bowls = {}
+    this.strikes = {}
+    this.turn = 1
+    this.roll = 1
+  }
+
+  remainder() {
+    if (this.roll === 1) {
+      return 10
+    }
+    else
+    {return 10 - this.bowls[this.turn][0]}
+  }
+
+  myRoll(number) {
+    if (this.roll === 4 || this.turn === 11)
+    {return 'The game is over! No more rolls'}
+
+    if (number > 10 || (this.roll === 2 && this.turn != 10 && number + this.bowls[this.turn][0] > 10) || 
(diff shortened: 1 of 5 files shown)


test/bowling.test.js
```diff
@@ -0,0 +1,187 @@
...
+    bowling.myRoll(10)
+    bowling.myRoll(10)
+    bowling.myRoll(10)
+    bowling.myRoll(10)
+    bowling.myRoll(10)
+    bowling.myRoll(10)
+    expect(bowling.myRoll(10)).toEqual({"1": [10], "10": [10, 10, 10], "2": [10], "3": [10], "4": [10], "5": [10], "6": [10], "7": [10], "8": [10], "9": [10]})
+    expect(bowling.totalScore()).toEqual(300)
+  })
+
```
Riky5, line 138: I'm a bit confused how you get 300 if in the 10th frame you cannot add your full result ... but you only add 10. But well done in adding the "perfect game"

lib/bowling.js
```diff
@@ -0,0 +1,102 @@
...
+      {this.roll = 2}
+      else if (this.turn === 10 && this.roll === 2 && this.bowls[this.turn][0] != 10 && number + this.bowls[this.turn][0] != 10)
+      {this.roll = 1; this.turn += 1}
+    else if (this.roll === 1 && number === 10 && this.turn != 10) 
+    {this.turn += 1}
+    else if (this.roll === 3)
+    {this.roll = 4; this.turn += 1}
+    else {this.roll = 2}
+    return this.bowls
+  }
```
Riky5, line 56: a switch statement could be maybe used here?

Riky5 (commented):
I've tested it in the terminal. I like how you can see the frames each time. Code is a bit long in some functions, but It works. Well done in adding the CLI

PR #1439, opened 2022-01-09
bowling_challenge_JS

s-dousse:


Alaka-K (commented):
Your code is every easy to read. I appreciate your detailed README(need to take a leaf out of your book). I feel you have a good adherence to the SPR. If I could make any suggestion at all, it would be to have a seperate classes for your scores. Apart from that, Well Done!

cli.js
```diff
@@ -0,0 +1,32 @@
...
+
+const game = new BowlingGame();
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const gameCLI = function () {
+  rl;
+  rl.question("Enter command > ", function (line) {
```
Alaka-K, line 12: good idea to use a switch and case statements instead of if/elseif

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
+class BowlingGame {
+  constructor(rollList = []) {
+    this.rollList = rollList;
+    this.sum = 0;
```
alicelieutier, line 4: As mentioned previously, I'd work to remove this.
A good way to think about which variables you should store is that information should not be duplicated inside instance variables. Here the rollList is enough to fully describe the state of your system (including the score, potential strikes, etc), so it really should be the only instance variable of the class.

README.md
```diff
@@ -69,7 +67,172 @@ More about ten pin bowling here: http://en.wikipedia.org/wiki/Ten-pin_bowling
...
+We can call these functions as methods within the getTotalScore function, but we use the `this` keyword!
+I think this is how it works:
+so `this` will point to the object that is calling the method
+in tests we do `game.getTotalScore()` (before const game = new BowlingGame())
+so in getTotalScore() we do
+`this.isAStrike(i)` will be like `game.isAStrike(i)`?
+also `this` behaves differently with arrow functions, I haven't used them here , but that's just luck and the tests are still passing after refactoring
+
+also not sure that having `this.sum` as class property is the best? but then I can pass it around the different function without using a argument
+or this is bad practice ? or I should do the same for the `this.rollList[?]`?
```
alicelieutier, line 165: Great questioning here!
As you suspect, using `this.sum` everywhere is not best practice. This is because it makes the code easier to break. A developer who doesn't know your code (or even yourself in a few weeks weeks :) ) could expect `this.sum` to be reliable as the current sum of the rolls at all times, which is not guaranteed when you use it as an intermediate summing variables.

Passing arguments is a good thing, and makes functions more independent and easier to understand separately. I'll give some suggestions for refactors.

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  isASpare(i) {
+    return this.rollList[i] + this.rollList[i + 1] == 10;
+  }
+
+  calculateStrikeScore(i) {
+    console.log("calculating strike score");
+    return (this.sum +=
+      this.rollList[i] + this.rollList[i + 1] + this.rollList[i + 2]);
```
alicelieutier, line 22: You are returning and assigning to an instance variable at the same time. 
It's good to think of class methods at able to do two things:

* changing the state of the system (changing an instance variable)
* calculating something based on the state of the system and either returning it or printing it.

In a well organised class, as much as possible, it's good to have methods that only do one of these things.
Here because the name of the method is calculateStrikeScore, it clearly seems to belong in the first category, so I would prefer it to just return that strike score, maybe:
```
return this.rollList[i] + this.rollList[i + 1] + this.rollList[i + 2];
```

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  calculateSpareScore(i) {
+    return (this.sum += 10 + this.rollList[i + 2]);
```
alicelieutier, line 26: Same.

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  calculateBasicScore(i) {
+    return (this.sum += this.rollList[i] + this.rollList[i + 1]);
```
alicelieutier, line 30: Same.

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  moveToNextFrame(str) {
+    return str === "strike" ? 1 : 2;
+  }
+
+  getFinalScore() {
+    let rI = 0; // rollList Index
```
alicelieutier, line 38: You could have another accumulator here that would be your running sum, so:
```
let sum = 0
```

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+    for (let i = 0; i < 10; i++) {
+      if (this.isAStrike(rI)) {
+        console.log("strike");
+        this.calculateStrikeScore(rI);
```
alicelieutier, line 42: sum += this.calculateStrikeScore(rI);

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+        // console.log(this.sum);
+        rI += this.moveToNextFrame("strike");
+      } else {
+        if (this.isASpare(rI)) {
+          console.log("spare");
+          this.calculateSpareScore(rI);
```
alicelieutier, line 48: same

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+          console.log("neither strike nor spare");
+          this.calculateBasicScore(rI);
```
alicelieutier, line 51: same

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+        rI += this.moveToNextFrame("not strike");
+      }
+    }
+    return this.sum;
```
alicelieutier, line 56:
```
return sum
```

Not such a big change :) but it makes the class cleaner, and the methods make more sense individually.

alicelieutier: It's also interesting to me that you totally knew to do exactly this for `rI`.

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  getFinalScore() {
+    let rI = 0; // rollList Index
+    for (let i = 0; i < 10; i++) {
+      if (this.isAStrike(rI)) {
+        console.log("strike");
+        this.calculateStrikeScore(rI);
+        // console.log(this.sum);
+        rI += this.moveToNextFrame("strike");
```
alicelieutier, line 44: Ate this point, because you know it's a strike already, you could simplify and just write:
```
rI += 1
```

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
...
+      } else {
+        if (this.isASpare(rI)) {
+          console.log("spare");
+          this.calculateSpareScore(rI);
+        } else {
+          console.log("neither strike nor spare");
+          this.calculateBasicScore(rI);
+        }
+        rI += this.moveToNextFrame("not strike");
```
alicelieutier, line 53: And `rI += 2` here. Removing the need for the move to next frame method.

test/bowlingGame.test.js
```diff
@@ -0,0 +1,105 @@
...
+  it("can be initialised with no argument", () => {
+    expect(game.rollList).toEqual([]);
+  });
+
+  it("can store the rolls as an array", () => {
+    game.roll(8);
+    game.roll(1);
+    game.roll(3);
+    expect(game.rollList).toEqual([8, 1, 3]);
+  });
```
alicelieutier, line 15: Don't test the state of your class.
You should be testing behaviour of a class, which means you only should test a class through its methods that either print or return based on the state of the class.

The correct way here is to make some rolls, then check the score.
The reason for this is that the behaviour is stable - meaning the score would always be the same given the same rolls - however, you could change the class implementation to remove the rollList, or store the rolls differently like in a hash. By having tests that look at the implementation rather than the behaviour, you making it a lot harder to refactor you code.

alicelieutier: where refactor = change implementation without changing behaviour.

alicelieutier: It can be useful to think of behaviour as input/output

test/bowlingGame.test.js
```diff
@@ -0,0 +1,105 @@
...
+  it("can roll a perfect game", () => {
+    const perfectGame = new BowlingGame();
+
+    for (i = 0; i < 12; i++) {
+      perfectGame.roll(10);
+    }
+
+    // console.log(perfectGame.rollList.length); // 12 (10 strikes + 2 bonus rolls)
+    expect(perfectGame.getFinalScore()).toBe(300);
+  });
```
alicelieutier, line 104: Great tests.

bowlingGame.js
```diff
@@ -0,0 +1,68 @@
+class BowlingGame {
```
alicelieutier, line 1: It would be nice to have something separating public and private methods in this class. A common convention in JS is to prefix all private methods with `_`, like `_isAStrike(i)`.

Especially as you are not testing those "private" methods, which is the right move, showing that they are private is useful.

There is also a way to make them really private in ES6. It's not very used but you could use it if you wanted. check https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

alicelieutier (commented):
You've done really well with this challenge, as it is quite complex but you manage to do this while keeping your code clean and simple, with plenty of smaller methods which is great.
Most of your tests are great. 

There's a bit more refactoring you could do that would make the code better, see my comments below :)

PR #1443, opened 2022-01-10
My solution to bowling challenge

kim-morgan:
Hi reviewer!
I didn't manage to write a readme, but basically there are two classes:

- Game: This controls the flow of the game and scoring. Users input rolls using the "inputRoll" method, which records them and starts a new frame if necessary. If it's the final round and the frame is a strike or a spare, extra rolls will be added.
- Frame: Stores info about each frame

It should be able to calculate the score accurately of any game, including games with extra rolls and bonuses in the last round. The tests check several normal games as well as special cases such as perfect game and they score correctly.

Known issues are verbose tests (constantly redeclaring new instance of the class) and lack of mocking in tests. (Didn't have time..)

### DIFF
--- a/frame.js
+++ b/frame.js
@@ -1,0 +1,5 @@
+class Frame {
+
+}
+
+module.exports = Frame;
--- a/frame.js
+++ b/frame.js
@@ -5,6 +5,7 @@ class Frame {
     this.scoringRolls = [];
     this.maxRolls = 2;
     this.bonus = 0;
+    this.extra = false;
   }
 
   roll(pins) {
@@ -45,7 +46,7 @@ class Frame {
   }
 
   isComplete() {
-    return this.numberOfRolls() === this.maxRolls || this.isStrike();
+    return this.numberOfRolls() === this.maxRolls || (this.isStrike() && !this.isExtra());
   }
 
   addBonus(points) {
@@ -59,7 +60,12 @@ class Frame {
   addExtraRoll() {
     if (this.maxRolls < 3) {
       this.maxRolls++;
+      this.extra = true;
     }
+  }
+
+  isExtra() {
+    return this.extra;
   }
 
 }
--- a/frame.js
+++ b/frame.js
@@ -1,5 +1,41 @@
 class Frame {
 
+  constructor() {
+    this.rolls = [];
+    this.maxRolls = 2;
+  }
+
+  roll(pins) {
+    this.rolls.push(pins);
+  }
+
+  getRolls() {
+    return [...this.rolls];
+  }
+
+  getTotal() {
+    return this.getRolls().reduce((previousValue, currentValue) => previousValue + currentValue);
+  }
+
+  isStrike() {
+    return this.getRolls()[0] === 10;
+  }
+
+  isSpare() {
+    return this.getTotal() === 10 && this.numberOfRolls() === this.maxRolls;
+  }
+
+  numberOfRolls() {
+    return this.getRolls().length;
+  }
+
+  isComplete() {
+    return this.numberOfRolls() === this.maxRolls || this.isStrike();
+  }
+
 }
+
+frame = new Frame();
+
 
 module.exports = Frame;
--- a/frame.js
+++ b/frame.js
@@ -2,11 +2,17 @@ class Frame {
 
   constructor() {
     this.rolls = [];
+    this.scoringRolls = [];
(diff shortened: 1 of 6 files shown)


frame.js
```diff
@@ -0,0 +1,76 @@
...
+    this.rolls.push(pins);
+    this.scoringRolls.push(pins);
+  }
+
+  nonScoringRoll(pins) {
+    this.rolls.push(pins);
+  }
+
+  getRolls() {
+    return [...this.rolls];
```
msc49, line 21: amazing, i didnt know ...array existed :)

frame.js
```diff
@@ -0,0 +1,76 @@
...
+  isStrike() {
+    return this.getRolls()[0] === 10;
+  }
+
+  isSpare() {
+    return this.getTotal() === 10 && this.numberOfRolls() === this.maxRolls;
+  }
+
+  numberOfRolls() {
+    return this.getRolls().length;
```
msc49, line 45: very interesting I had used an index for this but the length function is a good shout

frame.js
```diff
@@ -0,0 +1,76 @@
...
+    this.bonus += points;
+  }
+
+  getBonus() {
+    return this.bonus;
+  }
+
+  addExtraRoll() {
+    if (this.maxRolls < 3) {
+      this.maxRolls++;
```
msc49, line 62: not sure what this.maxRolls++ does, the '++' is throwing me off

frame.test.js
```diff
@@ -0,0 +1,115 @@
...
+    })
+    it("should be able to record two rolls", () => {
+      frame.roll(2);
+      expect(frame.getRolls()[1]).toBe(2);
+    })
+  })
+
+  describe("#nonScoringRoll", () => {
+    it("should be able to roll a non-scoring roll (for bonus round at end)", () => {
+      frame = new Frame();
```
msc49, line 20: all these tests look great but you have already defined frame, so perhaps you can remove this when refactoring

game.js
```diff
@@ -0,0 +1,103 @@
...
+        let bonus = this.getFrames()[i+1].getRolls()[0]
+        this.frames[i].addBonus(bonus);
+      }
+    }
+  }
+
+  calculateStrikeBonus() {
+    for (let i = 0; i < this.frames.length; i++) {
+      if (this.getFrames()[i].isStrike() && i != 9) {
+        // If there is a strike, the next two rolls will be added as a bonus.
```
msc49, line 73: really appreciated these comments

game.js
```diff
@@ -0,0 +1,103 @@
...
+        // Since frames can be of varying length, we will search through the next frames until we have counted two rolls.
+        // Bonus for the final round is calculated separately in the extraRoll function hence i != 9
+        let bonus = 0;
+        let rollsLeftToCount = 2;
+        for (let j = i + 1; j < this.frames.length; j++) {
+          for (const roll of this.getFrames()[j].getRolls()) {
+            bonus += roll;
+            rollsLeftToCount--;
+            if (rollsLeftToCount === 0) { break; }
```
msc49, line 82: good use of breaks

game.test.js
```diff
@@ -0,0 +1,194 @@
+const Game = require('./game')
+
+describe("Game", () => {
+
```
msc49, line 4: tests look great, just maybe some refactoring is the only thing i can think of

msc49 (commented):
great job kim!

frame.js
kim-morgan: Haha, I didn't know it either but I found it out by looking up how to duplicate an array in Javascript (like array.dup in Ruby)
Found it here: https://www.freecodecamp.org/news/how-to-clone-an-array-in-javascript-1d3183468f6a/

frame.js
kim-morgan: ++ is the same as += 1, i.e. the following two pieces of code mean the same thing:
`this.maxRolls += 1;
this.maxRolls++;`

This ++ syntax doesn't exist in Ruby but it's really common in a lot of programing languages like JavaScript, Python, C, etc.!

frame.test.js
kim-morgan: Yeah I just needed to start a fresh frame for this test. I agree it's a bit verbose but wasn't sure how to do "before do" syntax in Javascript... will look into it!

PR #1447, opened 2022-01-10
Bowling-challenge-js

TomasGarciaDev:
To those who will review this challenge, it's almost finished. I still have to check and set up 2 or more scenarios. Thanks in advance!


TomasGarciaDev:
I added some files that were not needed, I'll create a new pull request

PR #1448, opened 2022-01-10
Bowling Challenge JS

Riky5:
translation of the Ruby Bowling Challenge. Added a CLI.js.


cli.js
```diff
@@ -0,0 +1,46 @@
...
+
+const Game = require('./lib/game.js')
+const Score = require('./lib/score.js')
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout
+});
+
+const question = util.promisify(rl.question).bind(rl);
```
Hereloss, line 12: Nice use of setting up the promise - I couldn't get mine to work that well!

cli.js
```diff
@@ -0,0 +1,46 @@
...
+console.log('WELCOME TO Bowling.JS')
+
+async function bowlingScoreCard() {
+  console.log('')
+  console.log(`-----> Score: ${score.total} <-----`)
+  console.log('')
+  try {
+    const answer1 = await question('Enter first roll: ')
+    const asnwer2 = await question('Enter second roll: ')
+    const answer3 = await question('Enter bonus roll: ')
```
Hereloss, line 25: I find it interesting you've done the 3 rolls as separate answers - I just did mine with inputting one after the other! I can't quite tell, but does the bonus roll input still come up even if you don't have the bonus roll?

lib/game.js
```diff
@@ -0,0 +1,40 @@
+class Game {
+  constructor () {
+    this.frames = []
+  }
+
+  addRolls = (roll1, roll2, bonusRoll = 0) => {
+    this.maxTen(roll1, roll2)
+    if (this.isLastFrame() && this.isSpare(roll1, roll2) || this.isLastFrame() && this.isStrike(roll1, 0)) {
+      this.frames.push([roll1, roll2, bonusRoll])
```
Hereloss, line 9: I think this is the bit saying to add a bonus roll if it's the last frame and a spare/strike is bowled? Not fully sure however - it seems like this method is more adding the rolls than this

lib/game.js
```diff
@@ -0,0 +1,40 @@
...
+  isStrike = (roll1, roll2) => {
+    const frame = [roll1, roll2]
+    let result = (this.frames !== null && frame[0] === 10 || this.frames !== null && frame[1] === 10) ? true : false
+    return result
+  }
+
+  maxTen = (roll1, roll2) => {
+    const result = [roll1, roll2].reduce((p, c) => p + c)
+    if (result > 10 && this.isLastFrame() === false) {
+      throw new Error('Sum of the rolls cannot exceed 10')
```
Hereloss, line 30: Good way to stop too high a roll being inputted!

lib/score.js
```diff
@@ -0,0 +1,24 @@
...
+    const pFrame = this.game.frames.at(-2)
+  
+    let result = (pFrame !== undefined && this.game.isSpare(pFrame[0], pFrame[1]) || pFrame !== undefined && this.game.isStrike(pFrame[0], pFrame[1])) ? (frameSum + this.addBonus(currFrame[0], currFrame[1])) : frameSum
+    return this.total += result
+  }
+
+  addBonus (roll1, roll2) {
+    const frame = [roll1, roll2]
+    const pFrame = this.game.frames.at(-2)
+    let result = this.game.isSpare(pFrame[0], pFrame[1]) ? frame[0] : roll1 + roll2
```
Hereloss, line 19: This bit reads slightly confusingly to me - I can see spares are taken into account, but I can't see that strikes are? Does isSpare account for strikes and spares together?

test/score.test.js
```diff
@@ -0,0 +1,43 @@
...
+    it('returns a bonus if a spare is scored', () => {
+      const gameDouble = { frames: [[6, 4], [4, 1]], isSpare: () => true }
+      const score = new Score(gameDouble)
+      expect(score.addBonus(4, 1)).toEqual(4)
+    })
+
+    it('returns a bonus if a strike is scored', () => {
+      const gameDouble = { frames: [[10, 0], [5, 3]], isSpare: () => false, isStrike: () => true }
+      const score = new Score(gameDouble)
+      expect(score.addBonus(5, 3)).toEqual(8)
```
Hereloss, line 40: A feature test (testing a full game) might be a good further test to add - I'm also not sure if the last frame will add as intended (not counting any strikes or spares on that frame for bonus points), just since I can't see anything for this

test/game.test.js
```diff
@@ -0,0 +1,69 @@
...
+  describe('addRolls', () => {
+    it('add the rolls to the frames array', () => {
+      game.addRolls(1, 4)
+      expect(game.frames).toEqual([[1, 4]])
+    })
+
+    it('does not add the rolls and throws error Sum of the rolls cannot exceed 10', () => {
+      expect(() => game.addRolls(5,6)).toThrow('Sum of the rolls cannot exceed 10')
+      expect(game.frames).toEqual([])
+    })
```
Hereloss, line 25: I'm guessing this would also work  (i.e. throw error) if I put my first bowl as 11?

lib/score.js
```diff
@@ -0,0 +1,24 @@
+class Score {
```
Hereloss, line 1: I debated making this a separate class too (I did for Ruby, but not JS!) - I think it probably works better as a separate class

Hereloss (commented):
Code looks good and clean, and everything seems to be laid out well - I can't quite see how the strikes are checked, but from your tests I can tell they do work!

PR #1453, opened 2022-01-30
Bowling Challenge

cris-ch:


bowling.js
```diff
@@ -0,0 +1,47 @@
...
+    this.rolls.push(pins);
+  }
+
+  _isSpare(frameScore) {
+    return frameScore === 10;
+  }
+
+  _isStrike(rollIndex) {
+    return this.rolls[rollIndex] === 10;
+  }
```
alicelieutier, line 16: This is a detail, but these two methods have name that make it look like they do very similar things, however they expect a very different argument. I wonder if there's a way to either:
- use the same kind of argument (eg the rollIndex of the first roll in frame, maybe?) 
- or have names that hint at their difference.

This comment is about code design / programmer experience of the code. When programming, we rely on many small hints to quickly understand what the code does / how it works. names are a big one, obviously, and here you use a number of conventions really well (the `_`, the `is` hinting at a boolean return), but consistency of method usage, or even where it is defined in the class are other hints a programmers considers - even unconsciously - when choosing which method to use.

bowling.test.js
```diff
@@ -0,0 +1,56 @@
+const Bowling = require('./bowling.js');
+
+let bowling;
+
+function multipleRolls(rolls, pins) { // function to roll several times
```
alicelieutier, line 5: Tiny detail on naming here - the name `rolls` make me think that it's an array, rather than an integer. You can check this article I wrote about naming if you'd like: https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098?sk=8176415e3a62db07be59877d61bac7de

bowling.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+describe("Bowling", () => {
+
+beforeEach(() => {          
+    bowling = new Bowling;
+  })
+
+  it("is an instance of Bowling class", () => {
+    expect(bowling).toBeInstanceOf(Bowling);
+  });
```
alicelieutier, line 19: This test is not useful - it just tests that Javascript works, not the behaviour of your class.

bowling.test.js
```diff
@@ -0,0 +1,56 @@
...
+    expect(bowling.score).toEqual(24);
+  });
+
+  it('returns the correct score for a 10 frames game', () => {
+    bowling.roll(10);
+    bowling.roll(2);
+    bowling.roll(5);
+    multipleRolls(16, 2);
+    expect(bowling.score).toEqual(56);
+  });
```
alicelieutier, line 53: Great behaviour based tests! You are missing some edge cases though, as I mentioned in the README.

alicelieutier (commented):
This is really good!
Good tests and good code design. You don't have information duplication and you are making the logic look quite straightforward, well done!
Some details in my comments below, the biggest one being that you are missing edge cases in your tests (I think you code works for them but the tests should prove that :) )

PR #1457, opened 2022-01-31
bowling-challenge js 

mikitaro-27:


README.md
```diff
@@ -1,75 +1,126 @@
...
+- given a strike in the 10th frame, 2 bonus rolls are allowed
+
+(TBC)...
+
+Notes:
+- I did not know how to go about mocking in jest, need to look at that.
+
+## Bowling rules model:
+
+![alt text](https://i.imgur.com/z4arXW4.jpg)
```
fpmrqs, line 126: Super fun command line display, bonus points for a very detailed readme

main/round.js
```diff
@@ -0,0 +1,60 @@
+const Frame = require('./frame')
+// Uncomment line below for node demo
+// const showBreakdown = require('./showbreakdown')
+
+class Round {
```
fpmrqs, line 5: Good SRP skills separating these classes, using dependency injection on this interaction could be a good way of isolating them even more.

main/frame.js
```diff
@@ -0,0 +1,63 @@
+class Frame {
+  constructor() {
+    this.score = 0;
+  }
+
+  addRoll(points) {
```
fpmrqs, line 6: I see that you have functions to add roll and add bonus on both classes, maybe there is a way to DRY this a bit?

tests/frame.test.js
```diff
@@ -0,0 +1,236 @@
+const Frame = require("../main/frame");
+
+describe("frame", () => {
+  it("has a score property of 0", () => {
```
fpmrqs, line 4: Jest has a beforeEach function that helps minimize setup repetitions, like instantiating the class you're testing

tests/frame.test.js
```diff
@@ -0,0 +1,236 @@
...
+    describe('when open frame', () => {
+      it('returns false', () => {
+        const frame = new Frame();
+        frame.addRoll(2);
+        frame.addRoll(7);
+        expect(frame.isSpare()).toEqual(false);
+      })
+    })
+  });
+});
```
fpmrqs, line 236: Very well thought tests, loads of different cases/possibilities being tested.

main/round.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!this.#hasTenFrames()) this.#fillOrCreateFrame(points)
+    
+    // uncomment the 3 lines below for node demo
+    // console.log(showBreakdown.generateTopstring(this.frames))
+    // console.log(showBreakdown.generateBottomstring(this.frames))
+    // return `The current score is: ${this.getTotalScore()}`
+  }
+
+  #fillOrCreateFrame(points) {
+    let mostRecentFrame = this.frames[this.frames.length - 1]
```
fpmrqs, line 30: Using a variable to access this kind of information is a great idea! Some lines in Javascript can feel very long and confusing.

fpmrqs (commented):
Very good job, I feel like we approached the problem in a similar way, so I learned a lot by looking at your code. You worked really hard on this and it shows, well done. There's a few comments suggesting tips and tricks to make the code a bit better, hope they're useful.

main/frame.cjs
```diff
@@ -0,0 +1,63 @@
+class Frame {
+  constructor() {
+    this.score = 0;
+  }
+
+  addRoll(points) {
+    this.#validateRoll(points);
+    this.roll1 || this.roll1 === 0 ? (this.roll2 = points) : (this.roll1 = points);
```
leoht, line 8: I'd suggest expanding this rather than using a ternary, to improve readability (ternary conditions are fine when used as evaluated expressions, but less obvious to read when the different branches of the conditions contain instructions, such as assigning a variable) - and then wrap it in a private method as well. Maybe something like:
```js
#updateRoll(points) {
  if (this.roll1 || this.roll1 === 0) {
    this.roll2 = points
  } else {
    this.roll1 = points
  }
}
```

tests/frame.test.js
```diff
@@ -0,0 +1,236 @@
...
+          expect(frame.bonusPoints).toEqual(1)
+        })
+      })
+      
+      describe('when open', () => {
+        it('assigns 0 bonus points', () => {
+          const frame = new Frame();
+          frame.addRoll(5)
+          frame.addRoll(4)
+          expect(frame.bonusPoints).toEqual(0)
```
leoht, line 76: Here, you ideally need some way to access the total number of points for this frame — it should be exposed through a method, e.g `frame.getBonusPoints()`, so again, we don't expose too much the internal implementation with attribute names

main/frame.cjs
```diff
@@ -0,0 +1,63 @@
...
+
+  #validateBonus(points) {
+    if (points > 10) throw "invalid bonus";
+    if (this.bonus1 != 10) {
+      if (this.bonus1 + points > 10) throw "invalid bonus";
+    }
+  }
+
+  #summariseFrame() {
+    if (this.isStrike()) this.roll2 = null;
```
leoht, line 49: A minor one — most JS code conventions and linter configurations (I think?) would suggest adding curly braces even for a single-line conditional body, e.g:
```js
if (this.isStrike()) {
  this.roll2 = null;
}
```

PR #1460, opened 2022-01-31
Kat JS Bowling Challenge

kat-mcbrooks:
I got a bit confused with the best way to 'inject' Frame class into Scorecard. I'm not sure this is the 'best practice', so would appreciate guidance on this and the best way of testing. Currently, I'm not using doubles in my tests, because Frame is constructed within inputFrame method, rather than being injected. Perhaps if I 'injected' an instance of Frame as an argument to Scorecard's inputFrame method, I would then have to use doubles in my scorecard.test file? Thanks in advance for your help!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
 
 Bowling Challenge
 =================
-My bowling scorecard program in JavaScript calculates the scores of a bowling game. For this challenge, I was instructed to just focus on the logic for bowling, following the scoring rules below. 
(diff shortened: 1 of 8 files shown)


KiranKukar (commented):
- You have great test coverage (that are all passing) and your READMe is really clear

 - You also have good separation of responsibility

- The logic code in your frame class for calculating the score is clear and a really neat way of solving the challenge

PR #1461, opened 2022-01-31
Bowling Challenge JS - Fran

fpmrqs:


frames.test.js
```diff
@@ -0,0 +1,49 @@
...
+
+    expect(frame.totalPoints()).toEqual(7);
+  });
+
+  it("adds bonus points if previous frame was strike", () => {
+    frame.createNewFrame([10]);
+    frame.createNewFrame([3, 4]);
+
+    expect(frame.totalPoints()).toEqual(24);
+  });
```
mikitaro-27, line 20: I am not sure if your code accommodates for a case of strike-->strike-->roll, in which case the roll will have to be added to both the previous frame, and the one before it as well.

frames.js
```diff
@@ -0,0 +1,50 @@
+class Frames {
+  constructor() {
+    this.frames = [];
```
mikitaro-27, line 3: I think, it would be better to have a Frame class that does not store all the frames. You could ahve another class that stores each round's frames.

scoreCard.test.js
```diff
@@ -0,0 +1,34 @@
+const ScoreCard = require("./scoreCard.js");
+
+describe(".addKnockedPins", () => {
+  let scoreCard;
+  beforeEach(() => {
+    scoreCard = new ScoreCard();
+  });
```
mikitaro-27, line 7: this is a good way to go about replacing ruby's lets!!!

frames.test.js
```diff
@@ -0,0 +1,49 @@
...
+  it("adds bonus points for spare on tenth frame", () => {
+    for (let i = 0; i < 9; i++) {
+      frame.createNewFrame([4, 4]);
+    }
+    frame.createNewFrame([4, 6]);
+    frame.createNewFrame([8]);
+
+    expect(frame.totalPoints()).toEqual(90);
+  });
+});
```
mikitaro-27, line 49: could you add an edge case for testing the perfect score? So that you know that you can be adding more bonuses if the last frame was strike.

mikitaro-27 (commented):
[may not be right]

I am trying to use the script in node but it is difficult to see what it actually is doing. It would be great to have the CLI finished. 

From my experience with this challenge, I was trygint to figure out how would someone use the script on a website. For example, like we did in the Chitter challenge. I woudl think that have a Frame class that is responsible for only creating a frame instance would be great. Still would allow you to access all particular information about the frame by using its methods. For example, frame.roll1, frame.roll2, frame.getScore().

 You could then have a Round or Scorecard class that is responsible for storing the frames and managing the whole round. Again, getting some methods that you would be able to use to show info about the round when actually applying some friendly UI.

PR #1462, opened 2022-01-31
bowling-challenge 

smm3000:
Completed most of the feature requirements except for the bonus frame on the 10th frame. There is also currently no UI.


jmodiale:
Overall, the code is readable and easy to follow. Great job!

PR #1465, opened 2022-02-02
Kat JS Bowling Challenge

kat-mcbrooks:
I got a bit confused with the best way to 'inject' Frame class into Scorecard. I'm not sure this is the 'best practice', so would appreciate guidance on this and the best way of testing. Currently, I'm not using doubles in my tests, because Frame is constructed within inputFrame method, rather than being injected. Perhaps if I 'injected' an instance of Frame as an argument to Scorecard's inputFrame method, I would then have to use doubles in my scorecard.test file? Thanks in advance for your help!
Also appreciate any refactor suggestions/advice
Also tried the CLI, which seems to be working.


cli.js
```diff
@@ -0,0 +1,37 @@
...
+  frame = rolls.map(roll => {
+    return parseInt(roll); // turn string into number
+  });
+  scorecard.inputFrame(frame);
+  if (scorecard.frames.length > 9) { 
+    confirmOutcome(); 
+    return rl.close();
+  }
+  console.log('Thank you. Your rolls have been added to your scorecard. Your final score will be calculated after all 10 frames have been entered.\n');
+  askUserIput();
```
alicelieutier, line 22: Nice use of recursion. Something else you could have done is a while loop, potentially with a break. 
Typically a recursion can always be turned into a while loop, and vice versa.

alicelieutier: One is not better than the other. recursion is more "functional style", and sometimes easier to write, but can also be difficult to understand (her it's quite straightforward) and can lead to stack overflows.
While is a more "iterative style", and needs more state...
It's good to be aware of both and a very good skill to be able to convert one into the other and back :)

frame.js
```diff
@@ -0,0 +1,50 @@
...
+      return next_frame.rolls[0];
+    }
+    else {
+      return 0;
+    }
+  }
+  _sumFrame() {
+    const sum_rolls = (accum, roll) => accum + roll;
+    return this.rolls.reduce(sum_rolls);
+  }
```
alicelieutier, line 48: Nice, very functional style again.

frame.js
```diff
@@ -0,0 +1,50 @@
...
+
+  _bonus(next_frame, next_next_frame) { //frame class does not need to know what next_frame and next_next_frame are, because these params will be passed in when scorecard methods are called
+    if (this._isStrike() && next_frame._isStrike()) {
+      if (next_next_frame == undefined) { // i.e. 9th frame
+        return 10 + next_frame.rolls[1];
+        } 
+      else {
+        return 10 + next_next_frame.rolls[0];
+      }
+    }
```
alicelieutier, line 34: That works. Something else you could do here it to flatten the next and next next and then take two bonuses - you wouldn't need to know where they come from.

scorecard.js
```diff
@@ -0,0 +1,21 @@
+const Frame = require('./frame');
+
+class Scorecard {
+  constructor() {
+    this.frames = []
+  }
+
+  inputFrame(rolls) { //rolls will be e.g. [2,3]
+    const frame = new Frame(rolls); // would it be better to inject this in? If I inject it, would I then need to use a double in the tests?
```
alicelieutier, line 9: You can inject the Frame class in your constructor, and then reuse it.
You would then need to mock that class in some way.
```js
constructor(frameClass = Frame) {
    this.frames = []
    this.frameClass = frameClass
  }
  
  // then
  const frame = new this.frameClass(rolls)
```

scorecard.js
```diff
@@ -0,0 +1,21 @@
...
+    this.frames.push(frame);
+  }
+
+  score() {
+    let total = 0; 
+    this.frames.map((frame, index, frames) => {
+      total += frame.score(frames[index+1], frames[index+2])
+    });
+    return total;
```
alicelieutier, line 18: This is so close to being perfectly functional ;)
Genberally, we avoid using map with functions that have side effects (meaning they change the state around them), in this case changing the total.
You can use a foreach instead, which typically is used with functions that have side effects.
Or, you could go full functional style and use a reduce here. (functional style hates side effects).
Doc for reduce - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

Let me know if you want help figuring out how to write a reducer here.

alicelieutier (commented):
Good attempt, clear design and good tests overall.
Currently, your scorecard tests are basically feature tests for the logic part of your applications (they take into account the Frame+Scorecard together). It's good to have these to test the logic works as a whole, but in theory, you should also have separate isolated unit tests for Scorecard.
You could also have feature tests for the whole app, including the cli, maybe stubbing the input and checking what is sent to the output by spying on it.

You used a more functional style of programming in a number of places, which I think is on point for JS, where this mix of OO + functional style is often appreciated. See my comments about that.

Let me know if you have questions about my review.

PR #1467, opened 2022-03-05
Eddie Arnold Bowling challenge javascript

eddiearnoldz:


frame.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+  isLive() {
+    return this.live;
+  }
+  addRoll() {
+    return this.rollCounter++;
+  }
+  addPinsToScore(pins) {
+    return this.score += pins;
+  }
```
JohnForster, line 23: Similarly, as above, instead of storing a `score` variable, could you instead calculate the current score on-demand from the rolls array?

frame.js
```diff
@@ -0,0 +1,74 @@
...
+    return this.rolls;
+  }
+  addPinsToRolls(pins) {
+  return this.rolls.push(pins);
+  }
+  isStrike() {
+    return this.rolls[0] === 10;
+  }
+  isSpare() {
+    return this.rolls.reduce((roll_one, roll_two) => roll_one + roll_two) === 10 && !this.isStrike();
```
JohnForster, line 37: This is a very complicated way of writing `this.rolls[0] + this.rolls[1] === 10`

frame.js
```diff
@@ -0,0 +1,74 @@
...
+  if (this.live === false) {
+    return this.rollCounter = 1
+  }
+  }
+
+  checkEndOfFrame() {
+    if (this.isStrike() || this.rollCounter === 3) {
+      this.live = false;
+     }
+  }
```
JohnForster, line 49: this.rollCounter is only being used here. If you used `this.rolls.length` instead, you could remove `this.rollcounter` (and all its related functions) from the class entirely.

JohnForster (commented):
There's a lot of good code here, but it could use some refactoring, and could use taking a step back to consider whether each part is strictly necessary. 

You are storing a lot of extra detail in state, and this complicates your code, and makes it more error prone. It's better to derive values from a small amount of state than store them separately. This is a programming concept called the "single source of truth", and it prevents the possiblity of values getting out of sync.

Additionally, the Scorecard class has a lot of logic relating to calculating the scoring of frames, such as checking whether or not they were a strike, and updating their score if they were. Could this logic be moved into a scorecard class? If so, what extra information would need to be passed as arguments to make this work?

PR #1468, opened 2022-03-06
Bowling Challenge - JS

CBuchan5:


bowlingScore.test.js
```diff
@@ -0,0 +1,42 @@
+const BowlingGame = require('./bowlingScore')
+
+describe('bowlingGame', () => {
+  it('makes a bowling class', () => {
+    const bowlingGame = new BowlingGame();
+  })
+  it('han an empty array for frames', () => {
+    const bowlingGame = new BowlingGame();
```
EdCope, line 8: Using `beforeEach()` , you can set up the `new BowlingGame()` object just once at the top of your code. A bit like using subject in Ruby.

bowlingScore.test.js
```diff
@@ -0,0 +1,42 @@
...
+    bowlingGame.knockedPins(0)
+    expect(bowlingGame.getScore()).toBe(0)
+    })
+
+  //   it('can roll a perfect game', () => {
+  //     const bowlingGame = new BowlingGame();
+  //     bowlingGame.knockedPins(10)
+  //     expect(bowlingGame.getScore()).toBe(300)
+
```
EdCope, line 26: Even though commented out, this test would require strike scoring. So a test for a single strike may be easier first.

bowlingScore.js
```diff
@@ -0,0 +1,18 @@
+class BowlingGame {
+ constructor() {
+   this.frames = []
```
EdCope, line 3: I think the way your test works this might be an array of all rolls rather than frames. Frames may require 2 (or even 3) balls, so can be a bit difficult to decide at this stage.

bowlingScore.test.js
```diff
@@ -0,0 +1,42 @@
...
+    const bowlingGame = new BowlingGame();
+  })
+  it('han an empty array for frames', () => {
+    const bowlingGame = new BowlingGame();
+    expect(bowlingGame.frames).toEqual([])
+  })
+  // needs a double as don't want to actualy add a number to the array, just test it is being added
+  // it('roll once', () => {
+  //   const bowlingGame = new BowlingGame();
+  //   expect(bowlingGame.knockedPins(1)).toEqual(1)
```
EdCope, line 14: If you commented out this, for when pushing to the frames array. You could change the expect to something like 
`expect(bowlingGame.frames).toEqual([1])` Last bit checking that it is an array in the end.

bowlingScore.test.js
```diff
@@ -0,0 +1,42 @@
...
+  //     const bowlingGame = new BowlingGame();
+  //     bowlingGame.knockedPins(10)
+  //     expect(bowlingGame.getScore()).toBe(300)
+    
+  // })
+})
+
+/* Start with creating a frame which will be 10
+
+add a frame function which will add all the frames as an empty array
```
EdCope, line 32: Great idea to pseudocode what you will do !

PR #1470, opened 2022-03-06
Bowling Challenge - E Cope

EdCope:
Included Frame class to try a different way but unsure if done correctly. Implementing felt a lot less straightforward than my previous attempt in Ruby without the class.


frame.js
```diff
@@ -10,7 +10,7 @@ class Frame {
     if(this.ball1 === 10){
       this.strike = true;
     }
-    if(this.ball1 + this.ball2 === 10){
+    else if(this.ball1 + this.ball2 === 10){
       this.spare = true;
     }
     return this.ball1 + this.ball2;
```
CBuchan5, line 16: Clear class instructions. Easy to follow what this is doing. Good idea to use a constructor.

PR #1471, opened 2022-03-06
final Bowling js

Haroon2021:


BowlingInJs.test.js
```diff
@@ -0,0 +1,45 @@
...
+it('Should return a erfect game', () => {
+    rollMany(10,12);
+    expect(game.score).toEqual(300);
+})
+
+
+function rollMany(pins, rolls) {
+    for (let i = 0; i < rolls; i++) {
+        game.roll(pins);
+    };
```
Ledders26, line 44: Nice use of a function to avoid repetitive code

BowlingInJs.test.js
```diff
@@ -0,0 +1,45 @@
+const BowlingGame = require('./BowlingInJs');
+
+let game;
+
+beforeEach(() => {
+    game = new BowlingGame();
+});
+
```
Ledders26, line 8: Good use of beforeEach

BowlingInJs.js
```diff
@@ -0,0 +1,51 @@
...
+        let score = 0;
+        let rollIndex = 0;
+
+        for (let frameIndex = 0; frameIndex < 10; frameIndex++) {
+            if (this.isStrike(rollIndex)) {
+                score += this.strikeBonus(rollIndex);
+                rollIndex ++;
+                continue;
+            };
+            const frameScore = this.rolls[rollIndex] + this.rolls[rollIndex + 1];
```
Ledders26, line 21: Does this mean you have to record a 0 after a 10 for a strike?

BowlingInJs.js
```diff
@@ -0,0 +1,51 @@
...
+    roll(pins){
+        this.rolls.push(pins);
+
+    };
+
+    get score() {
+        let score = 0;
+        let rollIndex = 0;
+
+        for (let frameIndex = 0; frameIndex < 10; frameIndex++) {
```
Ledders26, line 15: So does this mean the frame index, if at all queried, would always be one behind what you'd expect (ie the first frame is frame 0, whereas in reality people would expect it to be 1?)

BowlingInJs.js
```diff
@@ -0,0 +1,51 @@
...
+        };
+        return score;
+    };
+    isSpare(frameScore) {
+            return frameScore === 10;
+    };
+
+    spareBonus(rollIndex) {
+        return 10 + this.rolls[rollIndex + 2];
+    };
```
Ledders26, line 39: Nice easy working out of Bonus (and same below with strike)

Ledders26 (commented):
Good code.  Nicely done in places.  Couple of potential pointers:

- could there be a few more tests?  Some unit tests maybe for the Methods and proof that more edge case work?  For example how does it handle a spare in the final game?
- README needs to be amended to something of your own
- for next steps:
- what happens if a game finishes early?  Can you get the score?
- Potentially split out frames into a separate class?
- Can you find a way to record a strike without there having been a second roll of 0?

PR #1473, opened 2022-03-06
Trushna's bowling challenge

trushnarabadia:


bowling.js
```diff
@@ -0,0 +1,106 @@
...
+
+  get score() {
+    let score = 0;
+    let rollIndex = 0;
+
+    for (let frameIndex = 0; frameIndex < 10; frameIndex++) {
+      if (this.isStrike(rollIndex)) {
+        score += this.strikeBonus(rollIndex);
+        rollIndex++;
+        continue;
```
CTJMID, line 18: good use of continue key word

bowling.js
```diff
@@ -0,0 +1,106 @@
+class Bowling {
+  constructor() {
+    this.rolls = [];
+  };
+
+  roll(pins) {
+    this.rolls.push(pins);
+  };
+
+  get score() {
```
CTJMID, line 10: good use of 'get' key word

CTJMID (commented):
Nice clean code that. does exactly what you want it. to do. I would recommend adding a few tests to prove that the code catches edge cases, and doesn't allow the game to go on (or. something like that).
Very DRY and readable code! Nice use of some new ruby concepts.

PR #1475, opened 2022-03-07
Bowling game scorecard js

AliCo-Hash:


bowlingGame.test.js
```diff
@@ -42,4 +42,11 @@ describe('BowlingGame', () => {
     }
     expect(game.score()).toBe(26);
   })
+
```
nelsonclaire, line 45: Good clear tests, did you consider any mocking?

bowlingGame.js
```diff
@@ -10,6 +10,13 @@ class BowlingGame {
     let rollIndex = 0
 
     for(let j = 0; j < 10; j++) {
```
nelsonclaire, line 12: How does rollIndex and j fit together?

bowlingGame.test.js
```diff
@@ -0,0 +1,17 @@
+const { it, expect } = require('@jest/globals');
```
nelsonclaire, line 1: You could go further by using readline to allow user to input their scores

bowlingGame.js
nelsonclaire: How would this work for the bonus on the final frame if applicable?

PR #1477, opened 2022-03-07
Bowling_Challenge_JS

Rakshak81:


now-what (commented):
Nice work! Your code is clear and easy to read. Next step would be to work on some refactoring, but otherwise good job :)

PR #1478, opened 2022-03-07
Bowling challenge in javascript - Jane

jparsons-git:
Bowling challenge in javascript - Jane


bowler.js
```diff
@@ -0,0 +1,247 @@
+
+const Frame = require('./frame');
+const Roll = require('./roll');
+
+class Bowler{   
+
+  constructor(id, name, frames, bowlerTotal, strikeMsg, spareMsg, noScoreMsg){
```
winters145, line 7: I like the idea of being able to set up your own name and personalised messages!

frame.js
```diff
@@ -0,0 +1,92 @@
+
+//export default class Frame{
+class Frame{
```
winters145, line 3: Good use of multiple classes that all interact

frame.test.js
```diff
@@ -0,0 +1,47 @@
+const Frame = require('./frame');
+const Roll = require('./roll');
+
```
winters145, line 3: Using doubles in your tests would ensure your classes aren't dependent on each other to work, the way you've set up your constructors should mean you could use dependency injection to do this?

bowler.test.js
```diff
@@ -0,0 +1,160 @@
+const Bowler = require('./bowler');
+const Frame = require('./frame');
+const Roll = require('./roll');
+
```
winters145, line 4: Loads of tests! Very thorough coverage

winters145 (commented):
Really well done on this! Incredibly in-depth logic used and all bases covered, with loads of tests to go with it!

A little refactoring of the use of getters and setters could tidy things up that little bit further.

Good job!

PR #1481, opened 2022-03-07
Ben H bowling challenge

Ledders26:
Please note, currently incomplete.  Aim to work more on this during the week.

### DIFF
--- a/tests/scoreboard.test.js
+++ b/tests/scoreboard.test.js
@@ -1,4 +1,4 @@
-const Scoreboard = require('./scoreboard')
+const Scoreboard = require('../scoreboard')
 
 let scoreboard;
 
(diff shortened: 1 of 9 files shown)


tests/scoreboard.test.js
```diff
@@ -18,5 +18,9 @@ describe (Scoreboard, () => {
     it ('should return an error if a single roll is greater than 10', () => {
       expect(() => scoreboard.singleRoll(14)).toThrow('Score recorded for throw 1 is invalid');
     });
+
+    it ('should throw an error if a single roll is less than 0', () => {
+      expect(() => scoreboard.singleRoll(-1)).toThrow('Score recorded for throw 1 is invalid');
+    });
   });
 });
```
Haroon2021, line 26: This code is all well written and well structured. The variable names are clear so as to help people understand what each variable is supposed to do/keep a tally of.  I really like how you have implemented some input validation to ensure people write in a number between 1 and 10. You also tested for this which was good. The use of a reduce method was also something that helped make your code even more efficient (as opposed to using a loop).

I know that the code is incomplete at the moment, and I think going forward you want to implement some logic to consider spares and strikes.

PR #1482, opened 2022-03-07
Initial creation of program and tests.

Jaroslaw-Roman-Cudzich:


bowlingGame.js
```diff
@@ -0,0 +1,29 @@
...
+//           end
+//           rollIndex += 2
+//         end
+//         result
+//     end
+
+//     def spare? rollIndex
+
+//     end
+
```
benpartington, line 28: Nice work, you've got the strike calculation and normal frame score calculation. What you could add is the spare score calc, which is the same as the strike score code, except when you come to advance the rollIndex, you advance by 2 (rollIndex += 2 when it's a normal frame or spare, rollIndex +=1 when it's a strike).

PR #1484, opened 2022-03-07
Ben Partington Bowling JS

benpartington:


frame.test.js
```diff
@@ -0,0 +1,55 @@
+const Frame = require('./Frame');
+
```
Jaroslaw-Roman-Cudzich, line 2: const Frame = require('./Frame');       should be the following instead please to match the name of the file you are testing against (lower case f for frame): const Frame = require('./frame');         Thank you. Your code looks really good.

scorecard.js
```diff
@@ -0,0 +1,19 @@
+const Frame = require('./frame');
+
```
Jaroslaw-Roman-Cudzich, line 2: Great coding with splitting the functionalities into two separate classes and requiring one into the other. Excellent.

PR #1488, opened 2022-04-03
Gaya's Bowling Challenge in JS

gdrmvrm:
The program returns the perfect score. However, I was unsure how to tests mocks of the 'Add Bonus Points' function in the 'Calculate Bonus Points' function (in the Score Card Class) as the 'Calculate Bonus Points' function it does not return values.  I would much appreciate if I can get feedback on how this is done. Also, any feedback on my JS code setup and how I can improve will be very helpful! Thanks!

### DIFF
--- a/lib/frame.js
+++ b/lib/frame.js
@@ -18,8 +18,12 @@ class Frame {
     return this.roll3;
   }
 
-  tenthFrame() {
+  isTenthFrame() {
     return this.tenthFrame;
+  }
+
+  setTenthFrame() {
+    this.tenthFrame = true;
   }
 
   addRoll(roll) {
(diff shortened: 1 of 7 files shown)


lib/frame.js
```diff
@@ -0,0 +1,123 @@
...
+    return this.tenthFrame;
+  }
+
+  setTenthFrame() {
+    this.tenthFrame = true;
+  }
+
+  addRoll(roll) {
+    if (roll > 10) {
+      throw "Maximum score of 10 allowed";
```
adamwoodcock98, line 32: Nice job catching this edge case!

adamwoodcock98 (commented):
Really nice job Gaya!

You used some really nice features of JS that we've not covered yet like the try-catch blocks, so well done.

Lots of tests which is fantastic!

Didn't have enough time to clone the project and run it, but it seems very good and you've covered the tenth frame and the perfect score so great job!

leoht (commented):
Really good work! Re: mocking, you've correctly mocked the `Frame` class in the tests, and I left a few comments on things that I see could be improved — for example, you could simplify / extract the bonus calculation further (see my comments on the code below). Let me know if you have any more questions on it :)

PR #1491, opened 2022-04-04
Bowling Javascript

jamesamooney:
Frame class complete however not completed the scorecard class

### DIFF
--- a/frame.js
+++ b/frame.js
@@ -1,12 +1,70 @@
 class Frame {
   constructor(round = 1) {
-    this.round = round
+    this.points = 0;
+    this.round = round;
     this.rollOne = null;
     this.rollTwo = null;
-    this.score = 0
+    this.rollThree = null;
+    this.complete = false;
   };
 
+  roll(pins) {
+    if (this.complete) throw "It's over! Stop lobbing the balls!";
+  
+    if (this.rollOne === null) {
+      this.firstGo(pins);
+    } else if (this.round < 10 && this.rollTwo === null) {
+      this.secondGo(pins);
+    } else if (this.rollTwo === null) {
+      this.secondGoRound10(pins);
+    } else {
+      this.thirdRoll(pins);
+    }
+  }
 
+  strike() {
+    return this.rollOne === 10;
+  }
+  
+  spare() {
+    return this.rollOne !== 10 && this.rollOne + this.rollTwo === 10;
+  }
+
+  firstGo(pins) {
+    this.rollOne = pins;
+    if (pins === 10 && this.round < 10) {
+      this.complete = true;
+    };
+    this.frameTotal();
+  };
+
+  secondGo(pins) {
+    this.rollTwo = pins;
+    this.complete = true;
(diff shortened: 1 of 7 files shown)


frame.js
```diff
@@ -0,0 +1,70 @@
...
+    this.points = 0;
+    this.round = round;
+    this.rollOne = null;
+    this.rollTwo = null;
+    this.rollThree = null;
+    this.complete = false;
+  };
+
+  roll(pins) {
+    if (this.complete) throw "It's over! Stop lobbing the balls!";
```
PKilgarriff, line 12: Beautiful

frame.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  strike() {
+    return this.rollOne === 10;
+  }
+  
+  spare() {
+    return this.rollOne !== 10 && this.rollOne + this.rollTwo === 10;
+  }
```
PKilgarriff, line 31: These seem to be predicate methods - it might be worth naming them isSpare() and isStrike() as I believe this is convention (please don't ask me to source that)

frame.js
```diff
@@ -0,0 +1,70 @@
...
+
+  firstGo(pins) {
+    this.rollOne = pins;
+    if (pins === 10 && this.round < 10) {
```
PKilgarriff, line 35: pins === 10 could be replaced by your strike() method

frame.js
```diff
@@ -0,0 +1,70 @@
+class Frame {
+  constructor(round = 1) {
+    this.points = 0;
+    this.round = round;
+    this.rollOne = null;
+    this.rollTwo = null;
+    this.rollThree = null;
```
PKilgarriff, line 7: Could this be an array of rolls?

frame.js
```diff
@@ -0,0 +1,70 @@
...
+    if (this.rollOne === null) {
+      this.firstGo(pins);
+    } else if (this.round < 10 && this.rollTwo === null) {
+      this.secondGo(pins);
+    } else if (this.rollTwo === null) {
+      this.secondGoRound10(pins);
+    } else {
+      this.thirdGo(pins);
+    }
+  }
```
PKilgarriff, line 23: So if I'm reading this correctly - it is the job of the Frame class currently to assign the incoming roll to the correct part of the frame - using the Frame methods firstGo, secondGo, etc.?

frame.test.js
```diff
@@ -0,0 +1,110 @@
+const Frame = require("./frame");
+
+beforeEach(() => {
+  sut = new Frame();
```
PKilgarriff, line 4: Thank for telling me about 'sut' this will make my tests much neater!

PKilgarriff (commented):
Looks really good, well-named methods and variables - easy to follow the flow of the program. Looking forward to seeing the next steps as discussed on the call - I'd agree regarding testing multiple assertions - they are almost like feature tests.
- some of the logic might end up being extracted from the Frame class to the scorecard

PR #1492, opened 2022-04-04
JonnyBowl JS

jonnyabrams:


bowling.js
```diff
@@ -0,0 +1,36 @@
...
+      } else if (frameScore === 10) {
+        total += this.spareOrStrikeBonus(turn);
+        turn += 2;
+      } else {
+        total += frameScore;
+        turn += 2;
+      }
+    }
+    return total;
+  }
```
ruiined, line 27: This was really cool! Took me a few minutes to understand it, I think combining strike and spare bonuses into one was super neat 👌

bowling.test.js
```diff
@@ -0,0 +1,48 @@
+const Bowling = require('./bowling.js');
+
+const rollMultiple = (pins, rolls) => {
+  for (let i = 0; i < rolls; i++) {
+    game.roll(pins);
+  }
+}
```
ruiined, line 7: Awesome method. And, considering that your bowling scorecard is mainly for calculating the final score, I think you could move this method to your Bowling class! This way, you could utilise it, not only in your test, but when running the code itself.

bowling.test.js
```diff
@@ -0,0 +1,48 @@
...
+    game.roll(3);
+    rollMultiple(0, 17);
+    expect(game.score()).toEqual(26);
+  });
+
+  it('can roll a perfect game', () => {
+    rollMultiple(10, 12);
+    expect(game.score()).toEqual(300);
+  });
+
```
ruiined, line 46: The test has everything you need, can't fault it.

ruiined (commented):
I loved your code, it was very minimal and simple. That was satisfying to read. If there was a 10th frame game  logic, it would have been a perfect code, for me at least. I think simplifying the problem is something that a lot of us are struggling with and you totally nailed it.

PR #1497, opened 2022-04-04
creted frame class and game class, unit test for each and unit test f…

ChrisPinna:
…or perfect game

### DIFF
--- a/lib/frame.js
+++ b/lib/frame.js
@@ -1,0 +1,30 @@
+class Frame {
+  constructor () {
+    this.rolls = []
+    this.score = 0
+  }
+
+  feedRoll(number) {
+    this.rolls.push(number);
+  }
+
+  calculateScore() {
+
+    if (this.rolls.length < 2) {
+      this.score = 0;
+    } else if (this.rolls.length === 2) {
+      if (this.rolls[0] + this.rolls[1]  < 10) {
+        this.score = this.rolls[0] + this.rolls[1];
+      } else if (this.rolls[0] + this.rolls[1]  === 10) {
+        this.score = 0;
+      }
+    } else if (this.rolls[0] + this.rolls[1]  >= 10){
+      this.score = this.rolls[0] + this.rolls[1] + this.rolls[2];
+    } else {
+      this.score = this.rolls[0] + this.rolls[1];
+    }
+    return this.score;
+  }
+}
+
+module.exports = Frame;
--- a/lib/game.js
+++ b/lib/game.js
@@ -1,0 +1,64 @@
+const Frame = require('./frame');
+
+class Game {
+  constructor () {
+    this.rollHistory = [];
+    this.currentFrame = 0;
+    this.totalScore = 0;
+    this.frames = [
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame,
+      new Frame
+    ];
+  }
+
+
+  roll(number) {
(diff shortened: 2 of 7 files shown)


test/unit/frame.test.js
```diff
@@ -0,0 +1,33 @@
...
+    frame.feedRoll(5);
+    frame.feedRoll(5);
+    frame.feedRoll(5);
+    frame.calculateScore();
+    expect(frame.rolls).toEqual([5, 5, 5])
+    expect(frame.score).toEqual(15)
+  });
+  
+  test('calculateScore method should add the tree fist element of the rolls array if the sum of the first two elements of rolls the rolls array are morethan or equal to 10 ', () => {
+    const frame = new Frame;
```
arhussain1, line 22: Good use of tests, Only recommendation would be to use a before each hook for the following line
``const frame = new Frame;``

That way you don't need to write it again

test/unit/game.test.js
```diff
@@ -0,0 +1,20 @@
...
+    bowling.roll(4);
+    bowling.roll(5);
+    expect(bowling.calculateScore()).toBe(9);
+  });
+  
+  test('checkFrame should increase current frame if the frame current frame has 2 rolls', () => {
+    const bowling = new Game;
+    bowling.roll(4);
+    bowling.roll(5);
+    expect(bowling.currentFrame).toBe(0);
```
arhussain1, line 16: I may be completely wrong, but toBe sometimes fails for me when I expect something like the current frame toBe 0
So I started using toEqual instead, I haven't read much into it but I think it's more appropriate for when your expecting a value.

toBe may be more appropriate for when you expect something toBe nil or toBe true

lib/game.js
```diff
@@ -0,0 +1,64 @@
+const Frame = require('./frame');
+
+class Game {
+  constructor () {
+    this.rollHistory = [];
+    this.currentFrame = 0;
+    this.totalScore = 0;
+    this.frames = [
```
arhussain1, line 8: This seems like a much better idea than what I did, It's clean and simple there will never be more or less than 10 frames in a game

lib/frame.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  feedRoll(number) {
+    this.rolls.push(number);
+  }
+
+  calculateScore() {
+
+    if (this.rolls.length < 2) {
+      this.score = 0;
```
arhussain1, line 14: If line 14 is removed does it impact your code?

lib/frame.js
```diff
@@ -0,0 +1,30 @@
...
+    } else if (this.rolls.length === 2) {
+      if (this.rolls[0] + this.rolls[1]  < 10) {
+        this.score = this.rolls[0] + this.rolls[1];
+      } else if (this.rolls[0] + this.rolls[1]  === 10) {
```
arhussain1, line 18: In line 18 does this represent a spare? If so should line 19 be written as:
``this.score = this.rolls[0] + this.rolls[1];``

lib/game.js
```diff
@@ -0,0 +1,64 @@
...
+      console.log('nothing to add');
+    }
+  }
+
+  getCurrentFrame() {
+    return this.currentFrame;
+  }
+
+}
+
```
arhussain1, line 63: Overall, I think the testing is rock solid👍, but some of the logic could be refactored, for example the roll method seems to interact with the frame quite a bit, consider extracting some of that logic into the frame class. Also, try to include a README where you jot down the steps you plan to do for each commit. 

Other than that this looks pretty good and handles the core logic of the Bowling game.

PR #1508, opened 2022-05-29
Russell's JS Bowling Scorecard

Rmorbey:


scoreCard.js
```diff
@@ -0,0 +1,53 @@
+class ScoreCard {
```
sjtinsley, line 1: Overall this is really nice - the logic is tight and the code is easy to understand. Likewise your tests are comprehensive and easy to follow. Good job!

PR #1509, opened 2022-05-29
Bowling challenge JavaScript - Erlantz

ErlantzR:
Doesn't include an interface


frame.js
```diff
@@ -0,0 +1,50 @@
...
+
+  addBonusScore(pinsKnocked) {
+    if (this.bonusCount > 0) {
+      this.bonusScore += pinsKnocked;
+      this.bonusCount -= 1;
+    }
+  }
+}
+
+module.exports = Frame;
```
GeorgeDainton, line 50: I really like the checkPins and addBonusScore methods for calculating strikes and spares - a clear, well written and elegant solution to calculating bonus points

scoreCard.test.js
```diff
@@ -0,0 +1,76 @@
...
+    card.includingRollResult(3);
+    card.includingRollResult(3);
+    card.includingRollResult(2);
+    card.includingRollResult(7);
+    card.includingRollResult(3);
+    card.includingRollResult(1);
+
+    expect(card.score).toEqual(110);
+  });
+});
```
GeorgeDainton, line 76: Great test converge here for scoring 2 different test games

specialFrame.js
```diff
@@ -0,0 +1,39 @@
...
+        this.complete = true;
+      } else {
+        this.roll = 3;
+      }
+    } else {
+      this.complete = true;
+    }
+  }
+
+  checkPins() {
```
GeorgeDainton, line 27: I think this method is really efficient- concise and clear

specialFrame.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    expect(frame.complete).toEqual(false);
+
+    frame.countRoll(10);
+    frame.calculateScore();
+
+    expect(frame.score).toEqual(20);
+    expect(frame.complete).toEqual(true);
+  });
+});
```
GeorgeDainton, line 45: Great tests- each one is extensive but there are no more than necessary

GeorgeDainton (commented):
Excellent work Erlantz

PR #1518, opened 2022-05-30
Stevie Spiegl's JS bowling

S-Spiegl:
Incomplete

### DIFF
--- a/frame.js
+++ b/frame.js
@@ -3,9 +3,9 @@ class Frame {
     this.rolls = []
   }
 
-  input_roll(roll) {
-    if(roll <= 10 && this.sum_it(this.rolls) + roll <= 10){
-      this.rolls.push(roll)}
+  input_roll(pins) {
+    if(pins <= 10 && this.sum_it(this.rolls) + pins <= 10){
+      this.rolls.push(pins)}
     else {
       throw 'pick a lower number';
     }
@@ -25,6 +25,31 @@ class Frame {
     }
     else {
       return false}
+  }
+
+  bonus() {
+    if(this.strike === true || this.spare === true) {
+      return true
+    }
+    else {
+      return false
+    }
+  }
+
+  complete() {
+    if(this.rolls.length === 2) {
+      return true
+    }
+    else if(this.spare()) {
+      return true 
+    }
+    else if(this.strike()) {
+      return true
+    }
+
+    else {
+      return false
+    }
   }
   
   sum_it(array) {
--- a/frame.js
+++ b/frame.js
@@ -1,0 +1,41 @@
+class Frame {
+  constructor() {
+    this.rolls = []
+  }
+
+  input_roll(roll) {
+    if(roll <= 10 && this.sum_it(this.rolls) + roll <= 10){
+      this.rolls.push(roll)}
+    else {
+      throw 'pick a lower number';
+    }
+  }
+
+  strike() {
+    if (this.sum_it(this.rolls) === 10 && this.rolls.length === 1) {
+      return true}
+    else {
+      return false
+    }
+  }
+
+  spare() {
+    if (this.sum_it(this.rolls) === 10 && this.rolls.length === 2) {
+      return true
+    }
+    else {
+      return false}
+  }
+  
+  sum_it(array) {
+    let sum = 0
+    for (let i = 0; i < array.length; i++) {
+    sum += array[i]
+    }
+    return sum
+  }
+
+}
+
+
+module.exports = Frame
--- a/frame.test.js
+++ b/frame.test.js
@@ -1,0 +1,41 @@
+const Frame = require('./frame')
+
+describe('Frame', () => {
+    it('allows a player to input each of their rolls', () => {
+      let frame = new Frame();
+      frame.input_roll(9)
+      expect(frame.rolls).toEqual([9])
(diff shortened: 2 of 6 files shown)


frame.js
```diff
@@ -0,0 +1,66 @@
...
+
+  input_roll(pins) {
+    if(pins <= 10 && this.sum_it(this.rolls) + pins <= 10){
+      this.rolls.push(pins)}
+    else {
+      throw 'pick a lower number';
+    }
+  }
+
+  strike() {
```
lukestorey95, line 14: Use conditional return

frame.js
```diff
@@ -0,0 +1,66 @@
...
+    if (this.sum_it(this.rolls) === 10 && this.rolls.length === 1) {
+      return true}
+    else {
+      return false
+    }
+  }
+
+  spare() {
```
lukestorey95, line 22: Use conditional return

frame.js
```diff
@@ -0,0 +1,66 @@
...
+    if (this.sum_it(this.rolls) === 10 && this.rolls.length === 2) {
+      return true
+    }
+    else {
+      return false}
+  }
+
+  bonus() {
```
lukestorey95, line 30: Use conditional return

frame.js
```diff
@@ -0,0 +1,66 @@
...
+    if(this.strike === true || this.spare === true) {
+      return true
+    }
+    else {
+      return false
+    }
+  }
+
+  complete() {
```
lukestorey95, line 39: Another conditional return

frame.js
```diff
@@ -0,0 +1,66 @@
...
+    else if(this.strike()) {
+      return true
+    }
+
+    else {
+      return false
+    }
+  }
+  
+  sum_it(array) {
```
lukestorey95, line 55: Consider array.reduce() to sum array

frame.test.js
```diff
@@ -0,0 +1,41 @@
+const Frame = require('./frame')
+
+describe('Frame', () => {
```
lukestorey95, line 3: Consider a beforeEach to define the frame and dry up the tests

game.js
```diff
@@ -0,0 +1,54 @@
...
+
+    push_complete_frame_to_all_rolls() {
+        this.all_rolls.push(this.frame.rolls)
+    }
+
+    add_complete_frame_to_score() {
+        this.score += (this.frame.sum_it(this.frame.rolls))
+    }
+
+    game_over() {
```
lukestorey95, line 31: Another conditional return

game.test.js
```diff
@@ -0,0 +1,44 @@
+const Game = require('./game')
+
+describe('Game', () => {
```
lukestorey95, line 3: Again could be dryer with a beforeEach block

game.test.js
```diff
@@ -0,0 +1,44 @@
...
+        for (let i = 0; i < 10; i++) {
+            game.roll(10)
+            }
+            expect(game.all_rolls.length).toEqual(10)
+            expect(game.game_over()).toEqual(true)
+    })
+
+    it('returns a score if game is complete', () => {
+        let game = new Game(); 
+        for (let i = 0; i < 10; i++) {
```
lukestorey95, line 38: You use this code to make multiple rolls several times, could extract it to its own method

game.js
```diff
@@ -0,0 +1,54 @@
...
+        if(this.all_rolls.length === 10) {
+            return true
+        }
+        else {
+            return false
+        }
+    }
+
+    calulate_final_score() {
```
lukestorey95, line 40: Would love to see the implementation of this now you have a good setup

lukestorey95 (commented):
A good start, with all the structure setup nicely. Definitely some room to dry up both the code and tests, but it all works. Would love to have seen the implementation of the calculatescore function, as I think you have everything you need to get it working.

JohnForster (commented):
The overall solution here is nice, you've tested thoroughly, and made a good effort to keep your code simple and DRY. There are a few areas where it could be improved:

Try to keep your state as simple as possible. ie. A Frame stores an array of rolls, a Game stores an array of Frames. All other values can be calculated from this state, so this is all the state we should really need. By keeping the state simple in this way, our code is more flexible and easy to change.

You are also testing state over behaviour. When testing, we should be looking to test against
- What is _returned_ when we call our function
- What _external methods_ are called when we call our function
- How the _behaviour of other methods_ change after calling our function.

There's a video of me walking through the bowling challenge in JS [here](https://drive.google.com/file/d/1DkI1P5JPEmE17zDH_LiJbPWFdS9KNcrV/view) which you might find useful.

frame.js
```diff
@@ -0,0 +1,66 @@
...
+    let sum = 0
+    for (let i = 0; i < array.length; i++) {
+    sum += array[i]
+    }
+    return sum
+  }
```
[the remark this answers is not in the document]
S-Spiegl: Hi John, do you mean define this function in an another file and inject it into the class?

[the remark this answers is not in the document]
JohnForster: Yes, exactly. It doesn't need to be injected, it can be required from a util file. There is a package called `lodash` which contains a lot of utility methods, so you don't have to reinvent the wheel every time. I'd recommend using that.

PR #1519, opened 2022-05-30
Bowling Challenge (JS)

GeorgeDainton:


scorecard.js
```diff
@@ -0,0 +1,39 @@
...
+    return this.score
+  }
+
+  displayFrame() {
+    return this.frame
+  }
+
+  displayRoll() {
+    return this.roll
+  }
```
ErlantzR, line 18: I believe technically these are not needed. You can call 'instance_of_Scorecard'.score and it will return the score even without the method created. The same for other 2

scorecard.js
```diff
@@ -0,0 +1,39 @@
+class Scorecard {
```
ErlantzR, line 1: Good naming for all methods and variables. Not problem to understand what each method is doing. Of course you will have to include the handling of spares/strikes, which will make the code more complicated, but I like you tried to make it simple for the part that you did

PR #1534, opened 2022-08-30
Bowling challenge in JS - Naomi

naomischlosser:
Only calculates the total score, haven't TDD's the user input in the app.js

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,0 +1,35 @@
+const Frame = require('./frame');
+const Score = require('./score');
+const prompt = require('prompt-sync')({sigint: true});
+
+class App {
+  constructor() {
+    this.roll_1 = 0;
+    this.roll_2 = 0;
+    this.score = new Score();
+  };
+
+  run() {
+    console.log('Welcome to bowling!');
+
+    for (let i = 1; i <= 10; i++) {
+      console.log(`\nFrame ${i}`)
+      this.roll_1 = prompt('Please enter the first number of knocked down pins: ');
+      this.roll_2 = prompt('Please enter the second number of knocked down pins: ');
+
+      this.addFrames(Number(this.roll_1), Number(this.roll_2));
+    }
+
+    console.log(`\nCongratulations, your total score is ${this.score.total()}!`);
+  };
+
+  addFrames(roll_1, roll_2) {
+    const frame = new Frame(roll_1, roll_2);
+    this.score.add(frame);
+  };
+};
+
+module.exports = App;
+
+// const app = new App;
+// app.run();
--- a/frame.js
+++ b/frame.js
@@ -1,0 +1,20 @@
+class Frame {
+  constructor (roll_1, roll_2) {
+    this.roll_1 = roll_1;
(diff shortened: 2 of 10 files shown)


app.js
```diff
@@ -0,0 +1,35 @@
...
+  run() {
+    console.log('Welcome to bowling!');
+
+    for (let i = 1; i <= 10; i++) {
+      console.log(`\nFrame ${i}`)
+      this.roll_1 = prompt('Please enter the first number of knocked down pins: ');
+      this.roll_2 = prompt('Please enter the second number of knocked down pins: ');
+
+      this.addFrames(Number(this.roll_1), Number(this.roll_2));
+    }
```
Curtis-Turk, line 21: This could be broken out to another method so that your run() is simply descriptive

frame.js
```diff
@@ -0,0 +1,20 @@
...
+    this.roll_1 = roll_1;
+    this.roll_2 = roll_2;
+  };
+
+  getSum() {
+    return this.roll_1 + this.roll_2
+  };
+
+  isSpare() {
+    return this.roll_1 !== 10 && this.getSum() === 10;
```
Curtis-Turk, line 12: I like the use of the first method here. Concise

score.js
```diff
@@ -0,0 +1,38 @@
...
+    this.frames.push(frame);
+  };
+
+  forFrame(number) {
+    if (this.frames[number - 1].isStrike()) {
+      if (this.frames[number].isStrike()) {
+        return 20 + this.frames[number + 1].roll_1
+      }
+      
+      return 10 + this.frames[number].getSum();
```
Curtis-Turk, line 16: I like the bonus calculating using the previous frame, I encountered a lot of difficulty with the reverse i.e using the next frame

score.test.js
```diff
@@ -0,0 +1,95 @@
+const Score = require('./score');
+
+describe('Score', () => {
+  it('returns the result when player plays two frames', () => {
+    const score = new Score();
+    const mockFrame_1 = { getSum: () => 5, isSpare: () => false, isStrike: () => false };
+    const mockFrame_2 = { getSum: () => 8, isSpare: () => false, isStrike: () => false };
```
Curtis-Turk, line 7: Good use of mocks to ensure that the classes are independent

score.js
```diff
@@ -0,0 +1,38 @@
...
+    };
+  };
+
+  total() {
+    const calculateFrame = (i) => {
+      i += 1;
+      return this.forFrame(i);
+    }
+
+    const totalScore = [...Array(10).keys()].map(calculateFrame);
```
Curtis-Turk, line 32: This is a concise way to calculate the scores, If a little unreadable, potentially could be better to break out the [...Array(10).keys()] into a descriptive variable eg. 10Frames

Curtis-Turk: You could also just map onto the original frames

Curtis-Turk (commented):
Well done! nicely test driven code, obviously the app needs testing but I had the same issues there

PR #1535, opened 2022-08-30
weekend projet

ruben-wilson:


stephenfletchtek (commented):
It looks neat overall with just a few suggestions to tidy up here and there.

I guess you were planning to develop it a bit further?

PR #1537, opened 2022-10-03
incomplete conversion from ruby

ddrmv:


src/frame.js
```diff
@@ -0,0 +1,57 @@
...
+    this.spare = false;
+    this.unplayed = true;
+    this.complete = false;
+  }
+
+  setFirstRoll(firstRoll) {
+    this.firstRoll = firstRoll;
+    this.unplayed = false;
+    if (firstRoll === 10) {
+      this.spare = true;
```
iniffur, line 15: Should be

`this.strike = true:`

src/frame.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  getFirstRoll() {
+    return this.firstRoll;
+  }
+
+  setSecondRoll(secondRoll) {
+    this.secondRoll = secondRoll;
+    this.complete = true;
+    if (this.firstRoll + secondRoll === 10) {
```
iniffur, line 27: Check if firstRoll == 10?

PR #1550, opened 2022-11-21
Challenge submission

lwly-jpg:
- No UI - just core game logic.
- Frame score input assumes correct input for each bowl (no validation in place)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,15 @@
+module.exports = {
+  env: {
+    browser: true,
+    es2021: true
+  },
+  extends: 'standard',
+  overrides: [
+  ],
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module'
+  },
+  rules: {
+  }
+}
--- a/README.md
+++ b/README.md
@@ -1,75 +1,35 @@
+# Bowling Scorer
 
-Bowling Challenge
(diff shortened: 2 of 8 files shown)


williamlines:
Looking very good Adon! You have managed to implement the logic for all the features requested!

The class structure is clear with helpful comments to explain what each function is doing, and they are not too long while being named in a way that it is clear which action they perform. 

You have also constructed logical describe blocks in your testing, splitting your blocks into situations, and then actions that would follow it, for example describe block "after a full 10 frame game" follows with tests for every possible type of game that could have been played, with strikes, spares, or neither, and the perfect game. The tests are extensive and thorough. Perhaps you could have written a test for the gutter game case? Although I'm sure your code would have passed it!

Your comments were helpful for understanding your functions, and in particular in explaining your class structure, for example when you have written by your BowlingScore constructor "constructs with a scorecard from bowlingGame". It would have been useful to have an example of what a 'frame' looks like (I had to work it out from the tests), though that is hypocritical as I haven't written any comments at all. As you said yourself there is no validation in place, but that is not necessarily an issue when this program is not yet taking input from a user. 

I am very impressed with your work, and sorry for the wall of text!

lwly-jpg:
Thanks @williamlines ! Appreciate the review and feedback!

PR #1567, opened 2023-01-26
MichalSmyk Dec22

MichalSmyk:
Refactored last week's challenge in ruby into JS


sirdavy (commented):
This code looks really neat. I wish I had done it this way. Well done.

PR #1572, opened 2023-01-29
Bowling - Challenge

AndreeaDinculescu:


scoreCard.js
```diff
@@ -0,0 +1,14 @@
+const score = (user) => {
+    let scores = user.map((array, i) => {
+        if ((array[0] === 10) && (i < 9)) {
+          return 10 + user[i+1].reduce((a, b) => a + b);
```
SyntheticDBX, line 4: I like the use of the reduce method. Makes the function a lot shorter with less methods.

scoreCard.test.js
```diff
@@ -0,0 +1,37 @@
...
+          [1, 4],
+          [8, 1], 
+          [3, 5],
+          [2, 6],
+          [4, 3]
+        ];
+        expect(score(user)).toEqual(82);
+      });
+
+      it('returns the score with strike bonus', () => {
```
SyntheticDBX, line 20: Strike bonus is working as intended. Well done!

scoreCard.js
```diff
@@ -0,0 +1,14 @@
...
+        } else if (((array.reduce((a, b) => a + b)) === 10) && (i < 9)) {
+          return 10 + user[i+1][0];
+        } else {
+          return array.reduce((a, b) => a + b);
```
SyntheticDBX, line 8: For readability, I would give "a, b" names so it makes it easier to understand.

SyntheticDBX (commented):
Overall great solution, my only suggestion would be to give the parts that are "a,b / a + b" names to make it easier to understand at first glance.

PR #1580, opened 2023-01-30
Alex Shabib Hoggett - Bowling challenge

alexHoggett:


src/bowling.ts
```diff
@@ -0,0 +1,157 @@
+const prompt = require('prompt-sync')();
```
junocchi, line 1: Nice use of typescript.

junocchi (commented):
Great that you managed to keep it in one class and the table showing the results on command line is amazing. Just need to fix the bug which is affecting partial and total scores.

PR #1583, opened 2023-02-24
add example of usage/make clearer UI is optional

Samir70:
Wondering if the examples are suggesting too much about how the code should look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,9 @@ An example of how your code might be used could be:
 
 ```javaScript
 let scorecard = new Scorecard()
-scorecard.addRoll(5)
-scorecard.calculateScore() // returns 5
-scorecard.addFrame([3, 5]) // returns "Please finish the current frame [5, ?]"
-scorecard.addRoll(2)
-scorecard.addFrame([3, 5])
+scorecard.calculateScore() // returns 0
+scorecard.addFrame(2, 5) 
+scorecard.addFrame(3, 5)
 scorecard.calculateScore() // returns 15
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,16 +9,33 @@ Bowling Challenge
...
+
+```javaScript
+let scorecard = new Scorecard()
+scorecard.addRoll(5)
+scorecard.calculateScore() // returns 5
+scorecard.addFrame([3, 5]) // returns "Please finish the current frame [5, ?]"
+scorecard.addRoll(2)
+scorecard.addFrame([3, 5])
+scorecard.calculateScore() // returns 15
+scorecard.showScore() // [[5, 2], [3, 5]]
```
leoht, line 30: I would tend to remove this one, as it would tend to expose the internal state of the rolls data — the main behaviour we want the class / program to have is to calculate the final score, so I would say the `calculateScore()` method is the main one "returning" something. What do you think?

README.md
```diff
@@ -9,16 +9,32 @@ Bowling Challenge
...
 
+Start by looking in detail at the rules and the example of scoring for a complete game given below.
+
+An example of how your code might be used could be:
+
+```javaScript
+let scorecard = new Scorecard()
+scorecard.addRoll(5)
+scorecard.calculateScore() // returns 5
+scorecard.addFrame([3, 5]) // returns "Please finish the current frame [5, ?]"
```
leoht, line 26: Sorry just noticed this — I reckon we should only keep the method `addFrame()` and maybe have it take two arguments (first and second roll), so we remove the complexity of having to check the array size, or verify the current frame is correctly finished, etc. If the learner wants to implement extra methods in their solution, it's up to them, but at least this way we don't derive their attention to the main problem (calculating the score from the rolls data)

```js
let scorecard = new Scorecard()
scorecard.addFrame(5, 0)
scorecard.calculateScore() // returns 5
scorecard.addFrame(3, 5) 
```

Let me know what you think of this :)

Samir70: This makes sense, I've removed addRoll. So they can focus on addFrame and calculateScore, but add methods they think are useful.

PR #1597, opened 2023-04-29
Completed JS Bowling Challenge

siqbal181:


models/frame.js
```diff
@@ -0,0 +1,25 @@
+class Frame {
+  constructor() {
+    this.points = [];
+  }
+
+  addPoints(rollOne, rollTwo = 0, rollThree = 0) {
```
SomthingInteresting, line 6: Clear and easy to understand implementation of the Frame class.  I would recommend adding input validation to make sure over 10 cannot be entered as a value and it has to be a number.

tests/frame.test.js
```diff
@@ -0,0 +1,39 @@
...
+    frameOne.addPoints(5, 5);
+    expect(frameOne.spare()).toBe(true);
+  });
+
+  it("given the sum of one frame != 10", () => {
+    frameOne = new Frame();
+    frameOne.addPoints(4, 5);
+    expect(frameOne.spare()).toBe(false);
+  });
+});
```
SomthingInteresting, line 39: Good amount of tests for the Frame class

tests/integration/integration.test.js
```diff
@@ -0,0 +1,51 @@
...
+      frame = new Frame;
+      frame.addPoints(10, 0);
+      scorecard.addFrame(frame);
+    }
+    frame = new Frame;
+    frame.addPoints(10, 10, 10);
+    scorecard.addFrame(frame);
+    expect(scorecard.calculateScore()).toBe(300);
+  });
+});
```
SomthingInteresting, line 51: Worth adding some more tests to check different combinations of frames (strike, spare, double strike or spares and normal frames)

SomthingInteresting (commented):
Overall the code is easy to read and well organised.  Great work!

As we discussed a README with any features or notes would be beneficial to anyone looking to run the code on their machine.

PR #1606, opened 2023-05-02
Bowling challenge

tomcarmichael:


frame.js
```diff
@@ -0,0 +1,37 @@
...
+  constructor() {
+    this.rolls = [];
+    this.regularPoints = 0;
+    this.bonusPoints = 0;
+  };
+
+  roll(points) {
+    // check if frame has already ended
+    if (this.rolls.length == 2 || this.rolls.includes(10)) {
+      throw new Error('Tried to add points to a frame that is already over');
```
carolrs, line 10: I like this part where you are catching the error.

frame.js
```diff
@@ -0,0 +1,37 @@
...
+      this.rolls.push(points);
+      this.regularPoints += points;
+    };
+  };
+  // is this function necessary? This logic could be in Game
+  playFrame(points_array) {
+    points_array.forEach((points) => {
+      this.roll(points);
+    });
+  }
```
carolrs, line 24: Maybe to change this for the interface?

game.js
```diff
@@ -0,0 +1,62 @@
+class Game {
+  constructor(frameClass) {
+    this.grandTotal = 0;
+    this.frames = [];
+    for(let i = 0; i < 10; i++) {
+      let frame = new frameClass();
+      this.frames.push(frame);
+    };
```
carolrs: This could be a method, maybe?

game.js
```diff
@@ -0,0 +1,62 @@
...
+
+  playFinalFrameBonusRolls(bonusRollsArray) {
+    bonusRollsArray.forEach((points) => {
+      this.frames[9].bonusPoints += points;
+    });
+    // If player had a strike in frame 9 and 10, allocate the first bonus roll as bonus point to frame 9:
+    if (this.frames[9].isStrike() && this.frames[8].isStrike()) {
+      this.frames[8].bonusPoints += bonusRollsArray[0];
+    };
+  };
```
carolrs: Nice part

frame.js
carolrs, line 1: Hey Tom, I think your code is good, especially your tests. I believe you predicted a lot of things that could go wrong in the program in your tests, so you already threw exceptions for those cases.
I like that you've already created your frames at the beginning with 10, so you're sure you won't have a null number. That was pretty smart!

PR #1631, opened 2024-05-16
Pretty good solution in 1 day imo

wickr-man:
Didn't write tests, may do them later. A lot of manual testing looks good to me


wickr-man:
yes the code is ugly but it is very compact. If it works it works

